PR #602, opened 2026-02-03
fix(generation): resolve pthread module loading in bundled builds

andymai:
## Summary

- Fixes "Failed to resolve module specifier './brepjs_threaded.js'" error in bin designer when multi-threaded WASM is enabled
- Emscripten pthread workers need `mainScriptUrlOrBlob` set to load the main module correctly in production builds
- Without it, workers fall back to `import("./brepjs_threaded.js")` which fails after Vite bundles the code

## Changes

- Import threaded JS with `?url` suffix to emit as separate asset (instead of bundling inline)
- Dynamically import the module at runtime
- Pass `mainScriptUrlOrBlob` in config so pthread workers know where to load the main module from

## Test plan

- [x] `npm run build` succeeds
- [x] `npm run test:coverage` passes
- [ ] Verify bin designer 3D preview works in production build with cross-origin isolation (multi-threaded WASM enabled)

### DIFF
--- a/src/features/generation/worker/generation.worker.ts
+++ b/src/features/generation/worker/generation.worker.ts
@@ -105,19 +105,21 @@ async function initOpenCascade(): Promise<void> {
 
   let OC: Awaited<ReturnType<typeof opencascadeSingleInit>>;
   if (isThreaded) {
-    // Fetch module source and convert to data URL for pthread workers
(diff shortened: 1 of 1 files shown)


andymai:
Addressed the console.error feedback - removed the debug logging calls in 4ca62d08c.

PR #1342, opened 2026-04-09
fix(generation): export retry + PostHog telemetry for #1339 failures

andymai:
## Summary

#1340 (v4.37.4) was not a confirmed fix for #1339 — I never reproduced the bug, and a user was observed getting **multiple failure toasts on successive export clicks**, which rules out any single-regeneration theory like the one I shipped. This PR pivots to **instrumentation-first** and adds a safety-net retry.

## Why the previous fix wasn't enough

- My stale-triangulation story was plausible but unverified. PostHog has zero hits for \`STL_EXPORT_FAILED\` over 90 days — not because it doesn't happen, but because \`ExportDialog.tsx\` was catching the error into a toast with **no \`captureException\` call**. The failure never reached \`window.unhandledrejection\`.
- Reports of multiple successive failures mean the underlying cause survives a fresh \`generateBin()\` — suggesting a specific bin config produces topology \`StlAPI.Write\` rejects, or WASM state accumulates in a way the single-regen doesn't fix.
- Without production data on *which* bin configs fail, we're guessing.

## Changes

**Telemetry (critical, the actual goal of this PR)**
- \`ExportDialog.tsx\` now calls \`captureException\` **before** the toast on export failure, with rich context: dimensions, style, base config (lip / magnets / screws / solid / halfSockets), wall pattern, handles, cutout/insert counts, footprint mask, split config, and the first-attempt error \`.cause\` when retry fails.
- This gives us real production data to identify failing configurations.

**Retry safety net**
- \`binExporter.ts\` extracted the core export into \`runExportAttempt()\`. On first-attempt failure it clears \`lastSolid\` via \`setLastSolid(null)\` and retries with a fully fresh regeneration.
- On retry failure, the second error is thrown with its \`.cause\` set to the first error for diagnostics.
- Handles transient failure modes beyond preview-quality (WASM handle corruption, accumulated state) that #1340 didn't cover.

**Defensive disposal**
- \`setLastSolid()\` now swallows errors from the old solid's \`.delete()\` call. Leaking one WASM handle on an error path is strictly better than a double-dispose throwing and blocking the retry from ever running.

## Test plan

- [x] New regression test: \`retries once after first attempt fails on a corrupted cached solid\` simulates the exact failure class the retry exists for (disposes underlying Shape3D while leaving the export-quality flag true), exercises full retry path against real brepjs/WASM.
- [x] \`pnpm vitest run binExporter.test.ts shapeCache.test.ts\` → 22/22 passing
- [x] \`pnpm vitest run ExportDialog\` → 23/23 passing (catch path not exercised by existing tests, but happy paths unaffected)
- [ ] Post-merge: watch PostHog error tracking for \`\$exception\` events with \`source=bin_export\` — when one comes in, we'll finally know *which* bin config fails and can reproduce locally.

## Honest framing

This PR should not be read as "fixes #1339". It's "stops shipping #1339 blind" — the retry helps some class of failures (transient corruption), but the real win is finally getting telemetry so the next iteration can target the actual root cause.

### DIFF
--- a/src/features/bin-designer/components/ExportDialog/ExportDialog.tsx
+++ b/src/features/bin-designer/components/ExportDialog/ExportDialog.tsx
@@ -148,7 +148,6 @@ export function ExportDialog() {
         has_dividers: hasDividers,
         cutout_count: params.cutouts.length,
         insert_count: params.inserts.length,
-        has_footprint_mask: params.footprint !== undefined,
         // Include original error chain (from binExporter's retry path)
         // — the cause is set to the first-attempt error when retry fails.
         first_attempt_message: error.cause instanceof Error ? error.cause.message : undefined,
--- a/src/features/bin-designer/components/ExportDialog/ExportDialog.tsx
+++ b/src/features/bin-designer/components/ExportDialog/ExportDialog.tsx
@@ -17,6 +17,7 @@ import { formatPrintTime, formatFilament } from '@/features/bin-designer/utils/p
 import { generateFileName } from '@/features/bin-designer/utils/fileNaming';
 import { getSTLFileSize, estimate3MFFileSize } from '@/shared/generation/export';
 import { useToastStore } from '@/core/store/toast';
+import { captureException } from '@/shared/analytics/posthog';
 import { useTranslation } from '@/i18n';
 import { ExportDialog as SharedExportDialog } from '@/shared/components/ExportDialog';
 import type { ExportFileFormat } from '@/features/bin-designer/types';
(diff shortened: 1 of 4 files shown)


andymai:
Fixed in 84654ff — mirrored the defensive `.delete()` guard in `clearAllCaches()` for consistency with `setLastSolid()`. Same rationale: always null the pointer even if the old handle throws on disposal, so `lastSolidIsExportQuality` can't be left stale after a corrupted-state teardown.

PR #1384, opened 2026-04-17
fix(bin-designer): stop pinch-zoom from snapping camera back to isometric on mobile

andymai:
## Summary

- Mobile users could orbit and pinch the 3D preview, but the view would **snap back to the isometric preset the moment they released the gesture**, making zoom feel "clamped" and view angles un-changeable.
- Root cause: `handleCanvasDoubleTouch` in `PreviewCanvas.tsx` used a single timestamp to detect double-taps. Pinch release emits two `pointerup` events within ~10ms — they fell inside the 300ms window and fired `resetView()`.
- Fix: extract into `useDoubleTapReset`, which tracks active `pointerId`s and disqualifies the entire gesture if any frame saw multi-touch. Also rejects drags (>10px movement) and long-presses (>500ms), so quick sequential orbits don't misfire either.
- Bonus touch tuning on `OrbitControls`: `rotateSpeed` 1.0 (was 0.8), `zoomSpeed` 1.2 (was 1.0), `minDistance` unified to 20 (was 5 on mobile as a workaround for this bug).

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` — zero new warnings
- [x] `pnpm exec vitest run src/features/bin-designer` — 1708/1708 pass (includes 10 new hook tests)
- [ ] Manual: on a mobile device, pinch-zoom should stay zoomed after release
- [ ] Manual: on a mobile device, orbit to a new angle and the view should stay there
- [ ] Manual: legitimate single-finger double-tap still resets to isometric
- [ ] Manual: desktop behavior unchanged

### DIFF
--- a/src/features/bin-designer/components/PreviewCanvas/PreviewCanvas.tsx
+++ b/src/features/bin-designer/components/PreviewCanvas/PreviewCanvas.tsx
@@ -459,8 +459,11 @@ export function PreviewCanvas() {
 
   // Double-tap to reset view (touch only). The hook ignores multi-touch
   // gestures so pinch-to-zoom never misfires as a double-tap.
-  const { onPointerDown: onDoubleTapPointerDown, onPointerUp: onDoubleTapPointerUp } =
-    useDoubleTapReset({ onDoubleTap: resetView, disabled: isDesktop });
+  const {
+    onPointerDown: onDoubleTapPointerDown,
+    onPointerUp: onDoubleTapPointerUp,
+    onPointerCancel: onDoubleTapPointerCancel,
+  } = useDoubleTapReset({ onDoubleTap: resetView, disabled: isDesktop });
 
   // Scene dimensions
   const width = params.width;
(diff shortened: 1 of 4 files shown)


andymai:
Addressed review feedback in 33bfc56:

- **pointercancel leak (P1, Greptile + Copilot)**: added `onPointerCancel` to `useDoubleTapReset` and wired it in `PreviewCanvas`. The handler removes the cancelled pointer from the active set and, if it was the last active one, resets `wasMultiTouchRef`, `tapStartRef`, and `lastTapTimeRef` so subsequent clean taps detect normally. Two new tests cover cancel recovery and mixed pointerup/pointercancel during a pinch.
- **Missing preventDefault (Copilot)**: restored `e.preventDefault()` on the second tap when a double-tap is recognized, matching the previous inline handler's behavior. New test verifies it only fires on the triggering tap.

13 hook tests pass, full bin-designer suite 1711/1711. Typecheck + lint clean.

PR #1390, opened 2026-04-18
fix(bin-designer): shrink-to-fit long bin names in 3D preview

andymai:
## Summary

Long design names previously wrapped awkwardly to multiple cramped lines in the 3D preview because `maxWidth` was hard-coded at `outerW * 1.5` — only ~63mm on a 1×1 bin, which fits ~9 chars of 7mm text.

The label now:

- **Shrinks font size** from 7mm down to a 5mm floor to fit on one line
- **Falls back to 2-line wrap** at the default 7mm when text would go below 5mm
- Uses a **100mm absolute minimum** label width so tiny bins still have reasonable room

On a 1×1 bin (42mm, so 100mm floor kicks in):
- ≤21 chars → default 7mm, no wrap
- 22–29 chars → smooth shrink (7mm → 5mm)
- ≥30 chars → wrap to 2 lines at 7mm

Scope is intentionally limited to the bin designer's `BinNameLabel` — the near-duplicate in `src/shared/components/preview/BinNameLabel.tsx` (used by the layout planner) is left for a separate change.

## Test plan

- [x] `pnpm run test:run` — 9769 tests pass, incl. 4 new tests for shrink-to-fit, wrap fallback, and the 100mm floor
- [x] `pnpm run lint` — 0 errors
- [ ] Visual check in dev: open the bin designer, type a 25-char name on a 1×1 bin → label stays single-line
- [ ] Visual check: type a 60-char name on a 1×1 bin → label wraps to 2 lines at full size

### DIFF
--- a/src/features/bin-designer/components/preview/BinNameLabel/BinNameLabel.test.tsx
+++ b/src/features/bin-designer/components/preview/BinNameLabel/BinNameLabel.test.tsx
@@ -28,11 +28,7 @@ interface MockTextProps {
 
 vi.mock('@react-three/drei', () => ({
   Text: ({ children, fontSize, maxWidth }: MockTextProps) => (
-    <div
-      data-testid="r3f-text"
-      data-font-size={fontSize}
-      data-max-width={maxWidth === Infinity ? 'Infinity' : maxWidth}
-    >
+    <div data-testid="r3f-text" data-font-size={fontSize} data-max-width={maxWidth ?? 'none'}>
       {children}
     </div>
   ),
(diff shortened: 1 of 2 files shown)


andymai:
Addressed all four review comments in e49dfa78e:

1. **`maxWidth={Infinity}` → `undefined`** (Greptile #1) — widened the local `maxWidth` to `number | undefined`; React omits the prop when undefined, sidestepping any Infinity-based edge case in the renderer.
2. **Duplicate `idealFontSize` formula** (Greptile #2, Copilot #2) — now `availableWidth / estimateTextWidth(n, 1)`, so ratio constants live in one place.
3. **Off-by-one letter spacing** (Copilot #1) — `estimateTextWidth` now applies `LETTER_SPACING` across N−1 gaps instead of N. Regime boundaries (1×1 bin): ≤21 chars fits at 7mm, 22–29 chars shrinks, ≥30 chars wraps — effectively unchanged, just more accurate.

Test mock sentinel updated from `'Infinity'` to `'none'` for the no-wrap case; all 8 tests pass, typecheck clean.

PR #1453, opened 2026-04-25
chore: fix ESLint OOM and remove stale eslint-disable directives

andymai:
## Summary

- `pnpm run lint` was crashing locally with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed` — V8's 4GB heap default isn't enough for `strictTypeChecked` × 5 tsconfig.*.json projects. Bumped `NODE_OPTIONS=--max-old-space-size=8192` in the `lint` and `lint:fix` scripts.
- CI runners have more headroom, so `pnpm run lint` already succeeds in CI today; this is purely a local-dev fix that unblocks running lint, knip, and the wider quality suite without OOM.
- Bonus: ran `pnpm run lint:fix` once OOM was resolved. Autofix removed 5 stale `eslint-disable` directives whose underlying rule no longer flags the adjacent code (ESLint warned `no problems were reported from X`). Whitespace-only lines left by `--fix` were tidied up by hand.

## Why CI hasn't been complaining

`.github/workflows/ci.yml` runs lint with `continue-on-error: true`, so the broader 3846 pre-existing errors (mostly `@ts-nocheck` directives in worker/generated files) don't fail the build. They aren't introduced by this PR — they're long-standing and out of scope for this change. They'll be addressed in follow-up cleanup PRs.

## Files

- `package.json` — `NODE_OPTIONS` heap bump on `lint` + `lint:fix`
- `src/core/logger.ts` — drop 2 stale `no-console` disables
- `src/features/bin-designer/components/panel/ColorsSection/ColorZoneRow.tsx` — drop stale `i18next/no-literal-string` disable
- `src/features/generation/worker/generators/pipeline/collectOrigins.ts` — drop stale `no-console` disable
- `src/features/grid-editor/hooks/useInteraction.ts` — drop stale `no-unnecessary-type-arguments` block-disable

## Test plan

- [x] `pnpm run lint` runs to completion without OOM (was: `FATAL ERROR: Allocation failed`)
- [x] Vitest passes for the four files touched (180 / 180)
- [x] All pre-commit hooks pass (boundaries, i18n × 4, exhaustiveness, component structure, missing tests, readme reminders)
- [ ] Manual: `pnpm run quality` runs locally without OOM in someone else's worktree (CI-equivalent)


andymai:
Addressed Greptile P2 in c1b3a95cf — `lint-staged` now also gets the `NODE_OPTIONS=--max-old-space-size=8192` flag.

Re: Copilot Windows-shell concern — leaving POSIX inline `VAR=value` syntax. The repo doesn't target Windows dev (no `cross-env` in deps, no other shell-portable accommodations elsewhere; `direnv`/`.envrc` and pre-commit hooks assume bash/zsh). Adding `cross-env` just for these two scripts is scope creep for a non-supported environment. Happy to revisit if Windows dev support becomes a goal.

Issue #1466, opened 2026-04-26
Site renders a blank page on load — uncaught TypeError: Cannot read properties of undefined (reading 'map') in layout-BYehWYRE.js


gavinmcfall:
Visiting https://gridfinitylayouttool.com/ produces a completely blank (dark) page. No UI ever mounts. The browser console shows an uncaught exception thrown from the layout-* chunk during initial render, which appears to abort the app before anything is drawn to the DOM.

### Steps to reproduce

1. Open https://gridfinitylayouttool.com/ in a fresh browser session (no extensions, no prior data).
2. Wait for the page to finish loading.
3. Open DevTools → Console.

### Expected behavior
The Gridfinity Layout Tool UI loads and is interactive.

### Actual behavior

- The page stays blank. The body contains markup but nothing renders visibly. The console logs:
- Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'map')
-     at https://gridfinitylayouttool.com/assets/layout-BYehWYRE.js:2:7599
- The error fires during the initial mount and repeats on every reload.

### Troubleshooting already attempted

- Hard reload — no change.
- Cleared localStorage and sessionStorage for the origin and reloaded — same error, same blank page. So this is not caused by stale/corrupt saved layouts; it reproduces with completely empty storage.
- Confirmed only one script is loaded from the page: assets/main-ChwzS3zc.js, which then loads the failing assets/layout-BYehWYRE.js chunk.

<img width="1473" height="1219" alt="Image" src="https://github.com/user-attachments/assets/bc61b17c-f7ea-4986-beb5-8e87f14e91f8" />

andymai:
Yikes. Thx for the heads up. Putting up a fix.

PR #1470, opened 2026-04-26
ci(smoke): gate PRs on Vercel preview boot smoke (#1466)

andymai:
## Why

Issue #1466 (blank-page production outage) shipped through every existing check — typecheck, build, unit tests, bundle size, e2e all passed. Nothing in CI ever loaded the actual built bundle in a real browser to confirm it boots. This PR adds that missing layer.

## What this catches

The exact failure shape from #1466: a chunk that **builds successfully** but throws an uncaught `TypeError` during initial evaluation, leaving the page blank. Also catches MIME-type regressions, missing assets, top-level `await` cycle failures, etc. Anything that prevents React from mounting will fail this check.

## How it works

1. Vercel emits a `deployment_status` event with `state=success` and `target_url=<preview>.vercel.app` when a PR preview is ready.
2. The new workflow listens for that event, filters to non-prod previews, and runs `playwright.smoke.config.ts` with `PREVIEW_URL` set to the target URL.
3. The single spec (`e2e/smoke-preview.spec.ts`) loads `/`, asserts:
   - **No `pageerror` events fired during boot** — catches any thrown exception.
   - **`#root` has at least one child after mount** — catches blank-page even if errors were swallowed.

## Why a separate Playwright config

The main `playwright.config.ts` ships a `webServer` block that boots a local dev/preview server. The smoke runs against an external URL, so it gets `playwright.smoke.config.ts` — single chromium project, no `webServer`, reads `PREVIEW_URL` from env.

## Verified locally

```
$ PREVIEW_URL=https://gridfinitylayouttool.com pnpm exec playwright test --config=playwright.smoke.config.ts
  ✓  1 [chromium] › e2e/smoke-preview.spec.ts:13:1 › production preview boots without errors (2.5s)
  1 passed (3.0s)
```

## Test plan

- [x] `pnpm typecheck` clean
- [x] Spec passes locally against fixed prod
- [ ] After this lands: confirm the workflow fires on a fresh PR (any small follow-up PR will exercise it)
- [ ] Once first green run lands, set `smoke-preview` as a required status check on `main` via `gh api`

### DIFF
--- a/.github/workflows/smoke-preview.yml
+++ b/.github/workflows/smoke-preview.yml
@@ -5,20 +5,24 @@ name: Smoke Preview
 # but throw an uncaught exception during initial evaluation, leaving a blank
 # page in production.
 #
-# Triggered by deployment_status events that Vercel posts when a PR preview
-# is ready. We filter to state=success + non-production targets so we only
-# run on actual preview readiness.
+# Uses a `pull_request` trigger so the workflow runs from the PR branch
+# itself (deployment_status events only fire workflows on the default
+# branch, which would prevent this PR from self-testing). The wait-for-
+# preview step polls GitHub deployment_status events for the Vercel
+# preview URL, then runs the smoke spec against it.
 
 on:
-  deployment_status:
+  pull_request:
+    branches: [main]
 
 permissions:
   contents: read
   deployments: read
   statuses: read
+  pull-requests: read
 
 concurrency:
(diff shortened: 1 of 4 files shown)


andymai:
Killing this idea — the Vercel preview protection wall makes a single-PR rollout impractical without a manual bypass-token setup. Will revisit if/when we want a different gate.

Issue #1494, opened 2026-04-27
baseplate: trace which BREP op produces inconsistent face orientations on corner-3/edge piece configs


andymai:
**Tracking issue for upstream root-cause investigation of #1490.**

PR #1493 ships a downstream BFS winding-repair pass that fixes the user-visible symptom (Bambu Studio "non-manifold edges" on baseplate STL/3MF exports). This issue tracks the deferred upstream investigation.

## Symptom recap

Some `buildBaseplateSolid` outputs have face orientations inconsistent with each other — specifically, the entire bottom face (and ~30% of pocket/dovetail-cut walls) are wound backwards relative to the rest of the mesh. PR #1473 removed a flawed per-triangle heuristic and trusts brepjs face orientation verbatim — that works for corner-1, edge-y-1, and magnet baseplate cases, but fails for corner-3, corner-4, and edge-x-1 (the user's NERDZ piece configs).

## Suspicious patterns

1. **Asymmetry vs. `boxBuilder`** (`src/features/generation/worker/generators/boxBuilder.ts:65`): bins use `extrude(+totalHeight)`. The baseplate uses `extrude(-totalHeight)` followed by a final `translate(+totalHeight)` to compensate (`baseplateGenerator.ts:1122` and `:1252`). Bin geometry is winding-clean across all configs; the baseplate's negative-extrude pattern correlates with the bug. This may or may not be the cause — only some piece configs flip, so there's a second factor.

2. **Corner-rounding intersect** (`baseplateGenerator.ts:1183`): `intersect(baseplate, roundedTranslated)` runs only when corners have rounding. The rounded slab is also extruded with `-totalHeight`. Intersect can introduce orientation surprises in OCCT when both inputs share a sign convention quirk.

3. **Connector cut + fuse pipeline** (`baseplateGenerator.ts:1219-1246`): the dovetail tongues (fuse) and grooves (cut) operate on the slab. The boolean pipeline executes them as a sequence; failure modes can leave seam faces with inconsistent orientation. The piece configs that fail (corner-3, corner-4, edge-x-1) all involve specific tongue/groove placements driven by the `edges` config.

## Reproducer

[...]

Issue #1490, opened 2026-04-26
Exported build plates are not manifold


gavinmcfall:
### Describe the bug
I exported a bunch of build plates for my tool chest (files attached) and none of them are manifold. I have to fix them in bambu studio

### To Reproduce
Steps to reproduce:

1. Create a build plate
2. Export
3. Import into Bambu Studio
<img width="390" height="121" alt="Image" src="https://github.com/user-attachments/assets/7182d50f-1347-4384-b6c8-aada137a78eb" />

### Expected behavior
Object is manifold

### Files exported from https://gridfinitylayouttool.com/baseplate
[gridfinity-baseplate-13x9-magnets-padded-connectors.zip](https://github.com/user-attachments/files/27106647/gridfinity-baseplate-13x9-magnets-padded-connectors.zip)

### Notes:
The screenshot shows a different name because I renamed my local files and have saved the repaired files over the top so I re-exported with the default files names to attach to this ticket

andymai:
Thanks for the detailed report and the attached files — I downloaded the ZIP and ran the meshes through three independent topology checkers. All four corner pieces come back as **mathematically manifold** by every standard test:

| Piece | V | F | Watertight | Winding consistent | Volume | Edges with adjacency ≠ 2 |
|---|---|---|---|---|---|---|
| corner-1 | 32934 | 66004 | ✅ | ✅ | 157.93 cm³ | 0 |
| corner-2 | 28220 | 56556 | ✅ | ✅ | 138.45 cm³ | 0 |
| corner-3 | 26334 | 52776 | ✅ | ✅ | 132.63 cm³ | 0 |
| corner-4 | 22568 | 45228 | ✅ | ✅ | 116.16 cm³ | 0 |

Every undirected edge is shared by exactly 2 triangles, with consistent winding (each directed edge appears exactly once), positive total volume, no degenerate / zero-area triangles, no T-junctions (at 0.1 µm tolerance), and no near-duplicate vertices (at 1 µm tolerance, and even down to exact float comparison). Both my custom Python checker and `trimesh` confirm `is_watertight=True` and `is_winding_consistent=True`.

That doesn't make Bambu's warning fake — it just means we need to figure out *what specifically* Bambu's 18,114-non-manifold-edge count is detecting, because it isn't the standard adjacency-count definition.

A couple of asks that would help narrow it down:

[...]

PR #1496, opened 2026-04-27
chore(baseplate): document winding-repair as no-op on brepjs 15.6.1 (#1494)

andymai:
## Summary

Closes the upstream investigation tracked in #1494 (deferred from the downstream-fix PR #1493).

- Adds a **`BaseplateProbe`** hook to `buildBaseplateSolid` so diagnostic tests can inspect the intermediate solid at every construction milestone (`slabExtruded`, `pocketsCut`, `cornerIntersected`, `magnetHolesCut`, `lightweightFloorCut`, `connectorPipeline`, `final`). Production cost: one null-check per milestone.
- Adds **`__dual-kernel__/diagnoseBaseplateWinding.test.ts`** — a step-walker that meshes each intermediate solid with the same tessellation parameters as the production STL export path and reports `directedEdgeCollisions` and `signedVolume` per step. A second test diffs raw `mesh()` output against `repairMeshWinding(...)` output to quantify how many triangles the repair actually flips.
- Updates docstrings in `repairMeshWinding.ts` and the `buildBaseplateSTL` block in `baseplateGenerator.ts` to record the investigation status.

## Findings

For every config — `corner-1`, `corner-3`, `corner-4`, `edge-x-1`, magnet baseplate, **and the user-reported 13×9 magnet+lightweight+connector reproducer from #1490** — `mesh()` already produces 0 directed-edge collisions and positive signed volume:

| config                       | tris    | edge-dupes | signed vol | triangles flipped by repair |
| ---------------------------- | ------- | ---------- | ---------- | --------------------------- |
| corner-1                     | 15,540  | 0          | 37,859     | **0** / 15,540              |
| corner-3                     | 15,584  | 0          | 44,847     | **0** / 15,584              |
| corner-4                     | 15,504  | 0          | 45,005     | **0** / 15,504              |
| edge-x-1                     | 15,584  | 0          | 40,615     | **0** / 15,584              |
| user 13×9 padded + magnets   | 219,584 | 0          | 500,980    | **0** / 219,584             |

`repairMeshWinding` is a complete no-op on the current brepjs 15.6.1 + OCCT kernel.

Independent verification: with `repairMeshWinding` disabled (via a temporary env-var hack, reverted in this commit), all 6 tests in `baseplateGenerator.scenario.winding.test.ts` still pass — no `directedEdgeCollisions > 0` and no `signedVolume < 0` failures.

## Conclusion

Per the acceptance criteria of #1494: **document why the downstream repair is the correct long-term fix**. The brepjs/OCCT tessellation bug that #1490 surfaced is no longer reproducible on brepjs 15.6.1 + OCCT kernel — either an upstream fix landed since the user's original export, or the bug was gated on environmental state (kernel init order, mesh cache state) that the current codebase no longer hits.

The repair stays in place as a defensive net for any future regression: its overhead is one BFS per export, which is acceptable insurance against the user-visible Bambu Studio failure.

## Test plan
- [x] `pnpm exec vitest run --config vitest.profile.config.ts src/features/generation/worker/generators/__dual-kernel__/diagnoseBaseplateWinding` — diagnostic test passes and prints the per-step + raw-vs-repaired tables.
- [x] `pnpm run test:run src/features/generation/worker/generators/baseplateGenerator.scenario.winding.test.ts` — all 6 existing winding scenario tests pass with the repair re-enabled.
- [x] `pnpm run test:run src/shared/generation/repairMeshWinding.test.ts` — all 9 unit tests pass.
- [x] `pnpm run typecheck` — clean.
- [x] `pnpm run lint` — clean (8 pre-existing unrelated warnings in `baseplateDirectMesh.ts`).

### DIFF
--- a/src/features/generation/worker/generators/__dual-kernel__/diagnoseBaseplateWinding.test.ts
+++ b/src/features/generation/worker/generators/__dual-kernel__/diagnoseBaseplateWinding.test.ts
@@ -52,8 +52,9 @@ interface StepMetrics {
 }
 
 const QUANTIZE = 1e4;
-const vKey = (x: number, y: number, z: number): string =>
-  `${Math.round(x * QUANTIZE)},${Math.round(y * QUANTIZE)},${Math.round(z * QUANTIZE)}`;
+function vKey(x: number, y: number, z: number): string {
+  return `${Math.round(x * QUANTIZE)},${Math.round(y * QUANTIZE)},${Math.round(z * QUANTIZE)}`;
+}
 
 /**
  * Count directed-edge collisions and divergence-theorem volume on a flat
@@ -85,10 +86,15 @@ function metricsFromMesh(
     const a = vKey(ax, ay, az);
     const b = vKey(bx, by, bz);
     const c = vKey(cx, cy, cz);
-    for (const e of [`${a}->${b}`, `${b}->${c}`, `${c}->${a}`]) {
-      if (seen.has(e)) collisions++;
-      else seen.add(e);
-    }
+    const ab = `${a}->${b}`;
+    const bc = `${b}->${c}`;
+    const ca = `${c}->${a}`;
+    if (seen.has(ab)) collisions++;
+    else seen.add(ab);
+    if (seen.has(bc)) collisions++;
+    else seen.add(bc);
+    if (seen.has(ca)) collisions++;
+    else seen.add(ca);
     volume += ax * (by * cz - bz * cy) + bx * (cy * az - cz * ay) + cx * (ay * bz - az * by);
   }
   return { collisions, volume: volume / 6 };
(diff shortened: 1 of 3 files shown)


andymai:
Addressed both findings in d895c439:

- **Probe lifetime** (`baseplateGenerator.ts:1069`): doc now explicitly states the handle is valid only synchronously during the callback, must not be stored or used after return, and would otherwise be a use-after-free in WASM.
- **Cache interference** (`diagnoseBaseplateWinding.test.ts:217`): added `beforeEach(clearBaseplateCaches)`. Verified: the step-walker output now includes `slabExtruded` for all 5 configs (it was missing before because the first test had warmed `slabWithPocketsCache`). Conclusion is unchanged — edge-dupes=0 and vol>0 at every step for every config; repair flips 0 triangles.

Thanks for both — the cache one was a real defect in my diagnostic.

PR #1499, opened 2026-04-27
fix(baseplate): split planner reserves bed budget for dovetail tongues (#1498)

### LINKED ISSUE #1498: Fit to print bed does not account for padding
The print bed splitting only accounts for the grids, not the padding, making the files too large for print bed.

**To Reproduce**
Steps to reproduce:
select a size of 512x324 incl padding
print bed set to 256
Generate the files
The sizes in the print-guide.txt says 256 in width, but the file corner-2 is actually 257.8 in width
Seems like the length is correct perhaps

**Expected behavior**
That the files fit a 256 print bed

andymai:
## Summary
- Fixes #1498 — split baseplate STLs exceeded the configured print bed when `connectorNubs` was enabled, because the planner only checked `gridUnits·gridUnitMm + padding ≤ printBedMm` and ignored the 1.5mm male dovetail tongue extending past the slab on every join edge.
- Threads `connectorNubs` and `invertDovetails` through `splitPlanner` so per-piece bed budget reserves space for the male side of each join. Female sides cut into the slab and consume nothing; middle pieces collapse to a single `TONGUE_PROTRUSION` because exactly one of left/right is male regardless of invert orientation.
- `TONGUE_PROTRUSION_MM` is duplicated locally — module boundaries forbid `features/baseplate` importing from `features/generation`. A comment notes the canonical source so the constants stay in sync.

## Repro from issue #1498
Baseplate 12×7 units with 4mm L/R padding (totals 512×324mm incl. padding) on a 256mm bed with dovetail connectors enabled.

| | Before | After |
| --- | --- | --- |
| Tiling | 2 cols × 2 rows of [6, 6]·[4, 3] | 3 cols × 2 rows of [4, 4, 4]·[4, 3] |
| Right-column STL width | 257.5 mm (overflow) | 256 mm (fits) |

## Test plan
- [x] New `splitPlanner.test.ts` test reproduces the 257.5mm overflow before the fix and asserts every piece (including dovetail tongue) fits the bed after.
- [x] New "ignores dovetail protrusion when connectorNubs is disabled" test pins existing `[6, 6]` behavior so the fix doesn't pessimize unsplit cases.
- [x] All 49 splitPlanner tests pass; all 295 baseplate tests pass.
- [x] `pnpm run quality` (typecheck + lint + knip) clean of new errors.

### DIFF
--- a/src/features/baseplate/utils/printGuide.test.ts
+++ b/src/features/baseplate/utils/printGuide.test.ts
@@ -120,4 +120,21 @@ describe('generatePrintGuide', () => {
     expect(guide).toContain('padding:');
     expect(guide).toContain('3mm');
   });
+
+  it('Size line includes dovetail tongue protrusion when connectorNubs is enabled (#1498)', () => {
+    // 4×4 grid, no padding, dovetails on. 4 units * 42mm = 168mm slab.
+    // Edge piece with one male join edge → bbox 168 + 1.5 = 169.5mm.
+    const params = makeParams({
+      width: 8,
+      depth: 4,
+      connectorNubs: true,
+    });
+    const guide = buildGuide(params);
+
+    // Each piece (4 wide) has one male join edge with invertDovetails=false:
+    // - left piece: right=join is female (no protrusion) → 168.0mm
+    // - right piece: left=join is male (1.5mm protrusion) → 169.5mm
+    expect(guide).toContain('168.0 × 168.0');
+    expect(guide).toContain('169.5 × 168.0');
+  });
 });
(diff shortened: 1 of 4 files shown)


andymai:
Re: Greptile P2 on `splitPlanner.test.ts:574-585` — this was already addressed in commit 5840d20f. The current `actualPieceBbox` helper uses the precise male/female convention: `startMale ? TONGUE : 0` for left/front, `!startMale ? TONGUE : 0` for right/back. No double-counting; would fail if the planner mistakenly reserved on both sides.

PR #1511, opened 2026-04-29
feat(bin-designer): harden bin export against intermittent worker failures

andymai:
## Summary

Users had to refresh the page when the WASM worker wedged during export. Root cause: `GenerationBridge.prepareExport` never invoked the timeout that generation already used, so unresponsive workers hung exports forever with no surfaced error and no retry.

This PR layers end-to-end resilience over the bin-designer single-bin export pipeline:

- **Per-export timeouts** on all 8 export methods (`ExportTimeoutError` → `TIMEOUT` code), reusing `computeGenerationTimeoutMs` / `computeBaseplateTimeoutMs` for complexity-aware budgets.
- **`BridgeManager.refresh()` + `subscribe()` / `engineReady`** so the wrapper can self-heal a wedged worker without disturbing other consumers.
- **`exportWithResilience` middleware**: 2 silent retries (200 ms / 800 ms backoff), then `bridgeManager.refresh()` and one final attempt with a fresh worker.
- **Top-level try/catch in `useExport.ts`** replaces silent unhandled rejections; fixes the bare `catch {}` that was swallowing pool-fallback errors during split export.
- **Queued click** while the engine is warming — clicks during WASM init no longer no-op silently.
- **Final-failure toast** with **Retry** + **Report issue** actions (latter prefills a GitHub issue with bin params + error class). Timeout-specific copy for `TIMEOUT` errors.
- **Worker error taxonomy**: `BREP_BOOLEAN_FAILED`, `MESH_TESSELLATION_FAILED`, `INVALID_PARAMS`, `EMPTY_GEOMETRY`, `OUT_OF_MEMORY`, `TIMEOUT`, `UNKNOWN`. Non-retryable codes (`INVALID_PARAMS`, `EMPTY_GEOMETRY`) skip retry.
- **Telemetry**: `bin_export_succeeded` / `bin_export_failure` PostHog events with format, duration, retry/restart counts, bin params, error class/message/stack.
- **i18n**: 5 new keys translated across all 7 locales (en, de, es, fr, nb, nl, pt-BR).

## Test plan

- [x] `pnpm run lint` 0 errors (10 pre-existing warnings in untouched files)
- [x] `pnpm run check:i18n` + `:values` + `:unused` pass
- [x] `pnpm run test:run` — 10,276 tests pass
- [x] `pnpm run build` succeeds
- [x] New sibling tests: `exportWithResilience.test.ts`, `BridgeManager.test.ts` (refresh + subscribe), `GenerationBridge.test.ts` (export timeouts)
- [ ] **E2E**: `e2e/bin-designer-export.spec.ts` covers default-bin STL happy path — please run `pnpm run test:e2e -- bin-designer-export` locally before merge
- [ ] **Manual smoke test**: bin designer → mutate param → Export STL/3MF/STEP, confirm download
- [ ] **Manual smoke test**: trigger a complex bin (large grid + hex pattern + cutouts) to exercise timeout/retry path


andymai:
Addressed all three Greptile findings in 5a6478ac4:

1. **Phase 2 always failed** (`exportWithResilience.ts`): added `bridgeManager.acquire()` between `refresh()` and the final `operation()` call, with `try/finally release()` so the fresh worker actually initializes before retry. Two regression tests assert the `refresh → acquire → op → release` order and that release fires even on persistent failure.
2. **Success-on-failure** (`ExportDialog.tsx` + `useExport.ts`): switched `downloadBin`/`downloadSplit` to return `boolean` (true on completed download, false on queued-for-warmup or caught failure) and gated the success toast + `closeDialog()` on the boolean. The dialog's `try/catch` was removed since the hook owns errors end-to-end.
3. **Lost diagnostic context** (`useExport.ts`): moved the rich bin-config payload (magnet/screw diameter, wall pattern, handles, cutouts/inserts, `error.cause` chain, retry/restart counts) from the dialog's now-dead catch into the hook's `captureException` call inside `handleExportError`.

Tests: 10,278 pass (10,276 + 2 new). Typecheck clean. Lint: 0 errors (pre-existing warnings in untouched files).

PR #1538, opened 2026-05-01
refactor(cqrs): move library cloudShare persistence into event subscriber

andymai:
## Summary

- Extracts inline `persistLibrary()` from `setCloudShare`/`clearCloudShare` Zustand mutations into a dedicated CQRS event subscriber (`connectLibraryPersistence`).
- Routes `useCloudShare` through `useMutations()` instead of pulling mutations directly off `useLibraryStore` — cloudShare now flows through the command bus, fires the corresponding domain events, and the subscriber persists.
- Library store mutations are now pure Zustand+Immer `set()` calls with no I/O. This is the prerequisite for a future migration to `apply(event, draft)`-style handlers where `apply` must be a pure draft writer.

## Why

Library `setCloudShare`/`clearCloudShare` mutations performed an inline `structuredClone` + async `saveLibrary()` after each `set()`. This made the store a hidden I/O boundary and would block the planned CQRS DX redesign (where `apply` is the only writer and must be pure). Other library fields are persisted indirectly via `useAutoSave`'s 1s debounce; cloudShare needed immediate persistence because it isn't part of the layout payload that triggers `useAutoSave`.

## Behavioral parity

cloudShare changes still persist immediately (no debounce). The persistence now happens in response to the `library.cloudShareUpdated` / `library.cloudShareCleared` events emitted by the existing `handleSetCloudShare` / `handleClearCloudShare` handlers, instead of as an inline side-effect of the store mutation.

## Test plan

- [x] New unit tests for `libraryPersistence` subscriber (events trigger save, unrelated events don't, structuredClone detaches from store, unsubscribe stops persisting)
- [x] `useCloudShare` test mocking updated from `useLibraryStore.setState({ clearCloudShare: spy })` to module-mock of `@/shared/contexts/MutationsContext`
- [x] Library store test "setCloudShare persist safety" reframed to assert mutation correctness (cloudShare is set on the entry); the original "no proxy revocation throw" assertion is moot now that the inline `structuredClone` is gone
- [x] Full vitest suite (10289 tests) green
- [x] `pnpm typecheck`, `pnpm lint`, `pnpm knip` all clean

### DIFF
--- a/src/core/cqrs/index.ts
+++ b/src/core/cqrs/index.ts
@@ -162,4 +162,4 @@ export { createCqrsMutations } from './integration/mutationsAdapter';
 
 export { applyEvent, replayEvents } from './projection/replay';
 
-export { connectSelectionPruning } from './subscribers';
+export { connectSelectionPruning, connectLibraryPersistence } from './subscribers';
--- a/src/core/cqrs/integration/mutationsAdapter.ts
+++ b/src/core/cqrs/integration/mutationsAdapter.ts
@@ -15,6 +15,8 @@ import type {
   Category,
   Layer,
   Drawer,
+  LayoutId,
+  CloudShareInfo,
   BaseplateParams,
 } from '@/core/types';
 import type { Result, ValidationError, LayoutError } from '@/core/result';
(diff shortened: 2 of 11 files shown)


andymai:
Addressed Copilot's review: `libraryPersistence.test.ts` now calls `resetLibraryStore()` in `beforeEach` + `afterEach` and `bus.clear()` on teardown, mirroring `selectionPruning.test.ts`. Local subscriber tests still green (5/5).

PR #1539, opened 2026-05-01
feat(cqrs): v2 defineCommand foundations + type-inference tests

andymai:
## Summary

Type-only foundations for the redesigned CQRS command surface. **No callers migrate yet** — this PR exists to validate that the proposed `defineCommand({...})` shape can drive end-to-end inference for payload, value, error union, event payload, and an auto-derived `Mutations<R>` surface, **without inference collapsing or compile-time regressing**.

## Why now

The full CQRS DX redesign sequences ~10 PRs. PR 1 (library persistence subscriber) shipped in #1538. This PR (PR 2) is foundations-only so the most consequential risk — type-inference health — is validated *before* any caller depends on the new API. Pass-2 review of the v3 plan flagged: "if type inference adds >2s to typecheck, fall back to per-command typed dispatchers and a thin manual `Mutations` object."

**Decision-gate result:** typecheck 9.16s (no regression vs ~9s baseline), inference works as designed.

## What's in the PR

New under `src/core/cqrs/v2/`:
- `types.ts` — `CommandDefShape`, `AggregateName`, `AggregateRoot`/`Draft` (Designer is `unknown` until that migration lands), `HandleCtx`, `MiddlewareFlags`, `AnyCommandDef`
- `defineCommand.ts` — typed identity factory; captures all generics from the literal
- `createRegistry.ts` — explicit array assembly, preserves per-command literal types under `byType`
- `Mutations.ts` — `Mutations<R>` mapped type + `PayloadOf`/`ValueOf`/`ErrorOf` extractors
- `sample.ts` — two sample commands driving type-inference assertions (deleted in PR 11 cleanup)
- `types.test.ts` — vitest `expectTypeOf` assertions covering literal type capture, extractor correctness, end-to-end `Mutations` derivation

## Architectural note

`AnyCommandDef` intentionally constrains only static fields (`type`, `aggregate`, `emitted`, `schemaVersion`, `descriptionKey`). The function fields (`handle`, `apply`, `aggregateId`) are excluded because **function-arg contravariance** would block concrete defs (whose `handle` takes a narrower payload) from being assignable to a registry array typed `readonly AnyCommandDef[]`. The registry preserves the function fields via the literal type `T[number]`, not via the constraint, so type info isn't lost — only the gatekeeping shape is loosened.

## What's NOT in this PR

- Runtime `createMutations(registry, bus)` dispatcher — lands when the first domain (bin/) migrates and needs a real wire-up
- `historyStore` rename — PR 3
- `undoCoordinator` + restore-via-CQRS wiring — PR 4
- Any caller migration — PRs 5–9

## Test plan

- [x] `pnpm typecheck` — clean, 9.16s (no regression)
- [x] `pnpm test:run src/core/cqrs/v2/` — 11/11 pass
- [x] `pnpm lint` — clean
- [x] `pnpm knip` — clean
- [x] `pnpm check:exhaustiveness` — clean

### DIFF
--- a/src/core/cqrs/v2/Mutations.ts
+++ b/src/core/cqrs/v2/Mutations.ts
@@ -15,20 +15,33 @@
  */
 
 import type { Result } from '@/core/result';
-import type { AnyCommandDef, CommandDefShape } from './types';
+import type { AggregateName, AnyCommandDef, CommandDefShape } from './types';
 import type { Registry } from './createRegistry';
 
-/** Extract the value type from a command def. */
+/**
+ * Extract the value type from a command def.
+ *
+ * The seventh slot uses `AggregateName` (the full union) rather than `never`:
+ * `'layout' extends never` is `false`, which would make the conditional
+ * collapse to `never` for any concrete command def. `'layout' extends AggregateName`
+ * is trivially true, so the conditional matches and `infer V` succeeds.
+ */
 export type ValueOf<C> =
-  C extends CommandDefShape<string, unknown, infer V, string, unknown, unknown, never> ? V : never;
+  C extends CommandDefShape<string, unknown, infer V, string, unknown, unknown, AggregateName>
+    ? V
+    : never;
 
 /** Extract the error union from a command def — the union of every `err()` return inside `handle`. */
 export type ErrorOf<C> =
-  C extends CommandDefShape<string, unknown, unknown, string, unknown, infer E, never> ? E : never;
+  C extends CommandDefShape<string, unknown, unknown, string, unknown, infer E, AggregateName>
+    ? E
+    : never;
 
 /** Extract the payload type from a command def. */
 export type PayloadOf<C> =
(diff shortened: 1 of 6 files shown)


andymai:
Addressed all four pieces of review feedback in 24f67ec:

1. **ValueOf/ErrorOf/PayloadOf** — switched the aggregate-slot from `never` to `AggregateName`. Both Copilot and Greptile correctly identified that `'layout' extends never` is false, which would collapse the extractors to `never` for any concrete def.
2. **AggregateDraft** — wrapped each aggregate in Immer's `Draft<T>` and dropped the `readonly` modifiers, matching its documented purpose as a writable draft (and disambiguating it from `AggregateRoot`).
3. **createRegistry** — now throws when two commands declare the same `type`, with an inline-snapshot test for the error message.
4. **types.test.ts** — removed misleading "runtime no-op" comment.

Local: typecheck 9.28s (still within the decision-gate budget), tests 12/12 pass, lint clean.

PR #1542, opened 2026-05-01
refactor(cqrs): wire undo/redo through the command bus

andymai:
## Summary

`useHistoryStore.undo`/`redo` previously called `useLayoutStore.restoreLayout()` directly and applied selection inline, bypassing the CQRS pipeline entirely. `restoreHandlers.ts` documented this with *"Currently not yet wired — history.ts still restores directly."*

This PR closes that gap. Undo/redo now dispatch `RestoreLayoutCommand` via the command bus, which routes through `handleRestoreLayout` to apply the layout + selection state. The bus path means restore is observable by analytics middleware and replayable through the event store, while the `restore` middleware profile (`{ undo: false, validation: false }`) keeps undoCapture from re-capturing the post-undo state.

## Changes

- **`RestoreLayoutCommand`** payload extended with optional `selection: SelectionSnapshot`.
- **`handleRestoreLayout`** uses the snapshot when present (restores the user's exact selection). Falls back to pruning stale IDs when absent — preserves existing behavior for any non-snapshotted call sites.
- **`useHistoryStore.undo`/`redo`** dispatch via `commandBus.dispatch(...)` instead of calling `restoreLayout` + `applySelection` inline.
- **Dead helpers removed** from `historyStore` (`applySelection`, `restoreSelectionFromSnapshot`, `pruneStaleSelections`, `fallbackActiveLayerId`). Their logic now lives in `restoreHandlers` as `applySnapshot` / `pruneSelection`.

## Why the static import cycle is benign

`historyStore` → `commandBus` → `middleware` → `historyStore` is a cycle in the import graph, but no functions execute at module-init. The `undo`/`redo` lambdas reference `commandBus` by closure, which is fully populated by the time any user action fires. Documented inline.

## Where this fits

PR 4 in the CQRS DX redesign sequence. Prior:
- #1539 (PR 2) — defineCommand foundations + type tests
- #1541 (PR 3) — history.ts rename to cqrs/undo/

Next:
- PR 5 — bin/ migration (proof of concept for `defineCommand`)

## Test plan

- [x] `pnpm typecheck` — 8.96s (no regression)
- [x] `pnpm test:run` — full suite, 10301 tests pass (the undo/redo path is exercised by all the existing historyStore + handler + middleware tests)
- [x] Two new tests in `restoreHandlers.test.ts` covering: (a) snapshot-restore preserves the user's exact selection, (b) snapshot reconciliation drops missing-bin IDs and falls back to last layer + first category
- [x] `pnpm lint`, `pnpm knip`, `pnpm check:boundaries` — clean

### DIFF
--- a/src/core/cqrs/commands/restoreCommands.ts
+++ b/src/core/cqrs/commands/restoreCommands.ts
@@ -6,10 +6,22 @@
 
 import type { BaseCommand } from '../types';
 import type { Layout } from '@/core/types';
+import type { SelectionSnapshot } from '../undo/historyStore';
 
+/**
+ * `selection` carries the user's selection state at the moment the
+ * snapshot was captured. When present, the handler restores the exact
+ * selection (active layer/category + selected/focused bins). When absent,
+ * the handler falls back to pruning stale references against the
+ * restored layout.
+ */
 export type RestoreLayoutCommand = BaseCommand<
   'layout.restore',
-  { readonly layout: Layout; readonly direction: 'undo' | 'redo' }
+  {
+    readonly layout: Layout;
+    readonly direction: 'undo' | 'redo';
+    readonly selection?: SelectionSnapshot;
+  }
 >;
 
 export type RestoreCommand = RestoreLayoutCommand;
--- a/src/core/cqrs/handlers/restoreHandlers.test.ts
+++ b/src/core/cqrs/handlers/restoreHandlers.test.ts
@@ -93,5 +93,57 @@ describe('Restore Handlers', () => {
         expect.objectContaining({ selectedBinIds: [] })
       );
     });
+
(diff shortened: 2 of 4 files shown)


andymai:
Addressed Greptile P2 in 0e96893: `applySnapshot` now compares each resolved value against current selection state and returns only the differing fields (same sparse shape as `pruneSelection`). When the diff is empty, `restoreSelection` is skipped entirely.

Tests updated to assert the new sparse semantics:
- snapshot-applies test now confirms only the differing field is restored
- snapshot-reconciliation test confirms only fields whose reconciled value differs from current are restored
- new test: when snapshot exactly matches current, `restoreSelection` is not called at all

Behavioral parity preserved when state actually differs; no needless Zustand notifications when it doesn't. Local: 5/5 restore tests pass, typecheck + lint clean.

PR #1543, opened 2026-05-01
feat(cqrs): migrate bin.add to v2 defineCommand (proof of concept)

andymai:
## Summary

First production command migrated to the v2 `defineCommand` shape. Validates the runtime adapter end-to-end and establishes the pattern for subsequent migrations.

## What's new

- **`v2/runtime.ts`** — `wrapV2Handler()` converts a `defineCommand` into a v1-shape `CommandHandler`. Validation middleware still runs first (Zod schema in `validation/schemas.ts`); the wrapper trusts payload by the time it executes. `handle()` runs against a frozen aggregate snapshot; `apply()` runs against an Immer draft of the appropriate store, mirroring the existing `setLocal` pattern (stamps `lastEditSource` so autosave still works).
- **`v2/domain/bin/addBin.ts`** — the migrated command. `handle()` generates the BinId, validates layer existence + placement against the snapshot, returns the full `Bin` in the event payload. `apply()` pushes onto the draft. **The "re-read after Immer mutation" smell from v1's `handleAddBin` is gone** — the event carries everything.
- **`v2/registry.ts`** — list of v2 commands + the `v2HandlerOverrides` map spread into the existing handler registry.
- **`handlers/index.ts`** — spreads `v2HandlerOverrides` last so migrated commands route through `defineCommand` wrappers; v1 commands keep working unchanged.

## Caller side: zero changes

`mutationsAdapter.addBin` still calls `commandBus.dispatch(createCommand('bin.add', ...))`. The bus now routes to the v2 wrapper instead of v1's `handleAddBin` via the registry override. No `MutationsContext` rewiring needed.

## Tests

Two **property tests** per the v2 design invariants:
1. **`handle()` determinism**: two invocations against the same layout produce structurally identical events (modulo BinId).
2. **`apply()` round-trip**: `produce(layout, apply(event)) === produce(layout, push(event.bin))`. Plus an empty-layout case.

Plus:
- `pipeline.test.ts` mock extended with `setState` (v2 wrapper now calls it; the v1 path used `getState().addBin` only).
- **All 10311 existing tests pass** — behavioral parity preserved end-to-end.

## What's NOT in this PR

- Other 9 bin commands stay on v1 — migrated in subsequent PRs.
- `mutationsAdapter` not yet replaced by auto-derived `createMutations(registry)` — happens once enough commands have migrated to make that worthwhile.

## Where this fits

PR 5 in the CQRS DX redesign sequence:
- #1539 (PR 2) — defineCommand foundations + type tests
- #1541 (PR 3) — history.ts rename to cqrs/undo/
- **#TBD (PR 5)** — bin.add v2 migration ← this PR

Next: PR 6 migrates the remaining bin commands (update, delete, deleteBatch, duplicate, moveToStaging, moveFromStaging, fillLayer, fillGaps, clearLayer), each with the per-command refactor checklist (e.g. duplicate's 5-placement search moves into `handle()`).

## Test plan

- [x] `pnpm typecheck` — 8.98s (no regression vs ~9s baseline)
- [x] `pnpm test:run` — full suite, 10311 tests pass
- [x] `pnpm lint` — clean
- [x] `pnpm knip` — clean
- [x] Property tests verify v2 design invariants for the migrated command

### DIFF
--- a/src/core/cqrs/handlers/index.ts
+++ b/src/core/cqrs/handlers/index.ts
@@ -13,11 +13,15 @@ import { libraryHandlers } from './libraryHandlers';
 import { designerHandlers } from './designerHandlers';
 import { restoreHandlers } from './restoreHandlers';
 import { uiHandlers } from './uiHandlers';
+import { v2HandlerOverrides } from '../v2/registry';
 
 export { resetVersionCounters } from './shared';
 
 type HandlerFn = (command: never) => CommandResult<unknown, DomainEvent>;
 
+// v2 overrides spread LAST so migrated commands route through defineCommand
+// wrappers instead of the v1 handler. Other commands keep the v1 path until
+// their domain migrates.
 const handlerRegistry: Record<string, HandlerFn> = {
   ...binHandlers,
   ...layerHandlers,
@@ -27,6 +31,7 @@ const handlerRegistry: Record<string, HandlerFn> = {
   ...designerHandlers,
   ...restoreHandlers,
   ...uiHandlers,
+  ...(v2HandlerOverrides as Record<string, HandlerFn>),
 };
 
 export function getHandler(
--- a/src/core/cqrs/integration/pipeline.test.ts
+++ b/src/core/cqrs/integration/pipeline.test.ts
@@ -70,7 +70,17 @@ const mockStore = {
 };
 
 vi.mock('@/core/store/layout', () => ({
-  useLayoutStore: { getState: () => mockStore },
+  useLayoutStore: {
+    getState: () => mockStore,
+    // v2 wrapper (cqrs/v2/runtime.ts) calls setState directly on the layout
+    // store. Mirror the real Zustand+Immer signature: producer receives a
+    // mutable draft of the whole store state, including `layout` and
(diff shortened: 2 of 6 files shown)


andymai:
Addressed both Greptile P2 findings in a607d4a:

1. **Unbranded ids via `as Bin` cast** — brand at the CQRS boundary using the codebase's existing constructors (`toLayerId`, `toCategoryId`, `toDesignId`, `gridUnits`, `heightUnits`) so the constructed `Bin` is properly branded end-to-end. The cast is gone. The branding pass is needed at this boundary because Zod payloads arrive as plain strings/numbers — v1 store actions don't need it because their callers already pass `Omit<Bin, 'id'>` with branded fields.
2. **ID generated before validation** — `generateBinId()` now runs only after layer-existence + placement checks pass, so rejected commands don't burn an id.

While fixing #1 I caught two more brand mismatches the original cast had hidden: `clearanceHeight: HeightUnits` and `linkedDesignId: DesignId`. Both branded now too.

Local: typecheck clean, 19/19 v2 tests pass, lint clean.

PR #1544, opened 2026-05-01
feat(cqrs): migrate 5 simpler bin commands to v2 defineCommand

andymai:
## Summary

Migrates `bin.update`, `bin.delete`, `bin.deleteBatch`, `bin.moveToStaging`, and `bin.clearLayer` to the v2 `defineCommand` shape (PR 5 established the pattern with `bin.add`). Each command:

- Validates against a frozen layout snapshot in `handle()`
- Returns the full event payload required for replay (bins captured before delete, previousLayerId before staging move, full `bins[]` before clearLayer, etc.)
- `apply()` mutates the Immer draft directly — same code path for live dispatch and event replay

## Per-command notes

- **`bin.update`**: explicit `Result` return type annotation because `handle` returns both `ValidationError` (placement) and `LayoutError` (not found); TS infers `TError` to a single type without the annotation. Worth flagging for the future `defineCommand` ergonomics — most commands won't need this annotation.
- **`bin.deleteBatch`**: lenient on missing ids (matches v1) — empty `bins[]` is a valid no-op result.
- **`bin.clearLayer`**: empty layer returns `0` + empty `bins[]` without emitting a wasteful event-applied mutation.

## Registry restructure

`v2HandlerOverrides` now built as a literal object instead of via `Object.fromEntries(v2Commands.map(...))`. Reason: each `wrapV2Handler` call needs a concrete `CommandDefShape` for inference; TS can't generate the wrapper from a heterogeneous union. The literal-object form keeps each call typed independently.

## Tests

- 5 sibling test files + shared `_testHelpers.ts` (`makeLayout`, `makeBin` fixtures).
- Each command covers handle correctness, error paths, and `apply()` round-trip equivalence with the v1 native mutation.
- Full vitest suite: **10328 tests pass** (was 10311 — +17 new property tests).

## Where this fits

PR 6 in the CQRS DX redesign sequence:
- #1538 (PR 1) — library cloudShare persistence subscriber
- #1539 (PR 2) — defineCommand foundations
- #1541 (PR 3) — history.ts rename to cqrs/undo/
- #1542 (PR 4) — undo via command bus
- #1543 (PR 5) — bin.add v2 migration (proof of concept)
- **#TBD (PR 6)** — 5 simpler bin commands ← this PR

Next: **PR 7** migrates the 4 remaining bin commands with their per-command refactors:
- `bin.duplicate` — 5-placement search moves from `binActions.duplicateBin:139-163` into `handle()`; event payload encodes resolved placement
[...]

PR #1546, opened 2026-05-01
feat(cqrs): migrate bin.duplicate + bin.moveFromStaging to v2

andymai:
## Summary

PR 7 in the CQRS DX redesign sequence — the two bin commands with non-trivial per-command refactors.

## bin.duplicate

- v1's 5-placement search (`binActions.duplicateBin:139-163`) moved into `handle()`. Search order matches v1 exactly: right → below → left → above → staging fallback. Staging-source bins always copy to staging.
- Event payload carries the resolved `{sourceId, newBin}` so `apply()` is a dumb push and replay does **not** re-run the search against potentially divergent layout state.
- `trackBinCreated()` called inline in `handle()` to preserve the v1 per-method analytics aggregation. Marked as a future refinement candidate for moving into a `bin.duplicated` subscriber.

## bin.moveFromStaging

- v1 event payload carried only `{id, layerId, x, y}`. Replay drift: the bin's height needed to be re-derived from the layer at apply time, and `projection/replay.ts` didn't update height at all.
- v2 widens the `BinMovedFromStagingEvent` payload with optional `height` (optional for v1-replay back-compat; the v2 handler always populates).
- `handle()` validates against `layer.height`, returns it in the event; `apply()` restores `layerId/x/y` AND `height`.

## Tests

2 new sibling test files. Property tests cover:
- `handle()` correctness (staging fallback, source-not-found, layer-not-found, collision)
- `apply()` round-trip equivalence
- The height-restoration invariant for `moveFromStaging`

## Where this fits

- #1543 (PR 5) — bin.add v2 (proof of concept)
- #1544 (PR 6) — 5 simpler bin commands
- **#TBD (PR 7)** — bin.duplicate + bin.moveFromStaging ← this PR

Next: **PR 8** migrates `bin.fillLayer` + `bin.fillGaps` and deletes `_fillMeta` (the 5-file delta with `layoutAnalytics` subscriber rewire).

## Test plan

- [x] `pnpm typecheck` — 8.96s (no regression)
- [x] `pnpm test:run` — full suite, **10338 tests pass** (was 10328 — +10 new property tests)
- [x] `pnpm lint` — clean
- [x] `pnpm knip` — clean

### DIFF
--- a/src/core/cqrs/events/binEvents.ts
+++ b/src/core/cqrs/events/binEvents.ts
@@ -36,6 +36,12 @@ export type BinMovedToStagingEvent = BaseDomainEvent<
   { readonly id: BinId; readonly previousLayerId: LayerId }
 >;
 
+/**
+ * `height` was added in the v2 migration so apply() can reproduce the
+ * full mutation without consulting current layout state. v1-era persisted
+ * events have no height — the optional shape keeps them readable; the
+ * v2 apply() falls back to leaving height untouched when missing.
+ */
 export type BinMovedFromStagingEvent = BaseDomainEvent<
   'bin.movedFromStaging',
   {
@@ -43,6 +49,7 @@ export type BinMovedFromStagingEvent = BaseDomainEvent<
     readonly layerId: LayerId;
     readonly x: number;
     readonly y: number;
+    readonly height?: number;
   }
 >;
 
--- a/src/core/cqrs/projection/replay.ts
+++ b/src/core/cqrs/projection/replay.ts
@@ -6,7 +6,7 @@
  */
 
 import type { Layout } from '@/core/types';
-import { gridUnits, mm } from '@/core/types';
+import { gridUnits, heightUnits, mm } from '@/core/types';
 import { STAGING_ID } from '@/core/constants';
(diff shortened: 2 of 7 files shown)


andymai:
Addressed all three review findings in 07590a6:

1. **Greptile P2 (duplicate staging fallback)** — added inline doc explaining the intentional v1-matching behavior: staging at (0, 0) skips `canPlaceBin` because v1's `store.addBin` also skips placement validation for `STAGING_ID`. Staging is a free-form bucket.
2. **Greptile P2 (moveFromStaging guard)** — tightened: now rejects when `bin.layerId !== STAGING_ID`. v1's lenient behavior was a footgun (silently re-stamped height). Added test for the rejection path.
3. **Copilot (replay drift)** — `projection/replay.ts` case `'bin.movedFromStaging'` now reads `event.payload.height` when present and sets `bin.height` accordingly. Closes the drift loop for newly-emitted v2 events. v1 events without `height` keep prior replay behavior.

Local: typecheck clean, full suite 10339 tests pass (was 10338, +1 for the new staging-guard test), lint clean.

PR #1548, opened 2026-05-01
feat(cqrs): migrate fill commands to v2; delete _fillMeta side-channel

andymai:
## Summary

Final two bin commands migrated to `defineCommand`. **All 10 bin commands now route through the v2 wrapper.**

### `bin.fillLayer` / `bin.fillGaps`

- Planning helpers (`fillAllWithSize`, `fillGaps` utility) called inside `handle()` against the frozen layout snapshot. The resulting bins are encoded in the event payload — `apply()` is a deterministic push.
- `LayerFilledEvent` payload widened with optional `fillType`, `width`, `depth` so the analytics subscriber can derive what `_fillMeta` used to carry. v1-era persisted events without `fillType` are silently ignored by the new subscriber (no metric distortion from re-emitting old events).

### `_fillMeta` deletion (6-file delta)

- `src/core/store/layout/types.ts` — drop `FillMeta` interface + `_fillMeta` field
- `src/core/store/layout/index.ts` — drop `_fillMeta` initializer + barrel re-export
- `src/core/store/layout/bulkActions.ts` — drop `_fillMeta` side-effect writes
- `src/core/store/layout/bulkActions.test.ts` — drop the `_fillMeta` assertion
- `src/core/store/layoutAnalytics.ts` — drop the meta-consumer block + import
- `src/core/store/layoutAnalytics.test.ts` — drop the `_fillMeta`-driven test
- `src/test/testUtils.ts` — drop `_fillMeta` from store reset helpers

### Replacement: `cqrs/subscribers/fillAnalytics.ts`

Subscribes to the `bin.layerFilled` event and forwards to the same `mlTracking` + `trackFillOperation` + `trackBinCreated` helpers the v1 subscriber called. Wired in `main.tsx` alongside `connectSelectionPruning` and `connectLibraryPersistence`. Has its own dedicated test file covering: uniform fills, gap fills, pre-v2 events without `fillType` (ignored), empty-bin events (ignored).

## Where this fits

- #1543 (PR 5) — bin.add v2 (proof of concept)
- #1544 (PR 6) — 5 simpler bin commands
- #1546 (PR 7) — duplicate + moveFromStaging
- **#TBD (PR 8)** — fillLayer + fillGaps + `_fillMeta` deletion ← this PR

**Bin domain complete.** Subsequent PRs migrate other aggregates (layer, category, drawer, library, designer) following the same pattern.

## Test plan

- [x] `pnpm typecheck` — clean (no regression)
- [x] `pnpm test:run` — full suite, **10347 tests pass** (was 10339; +8 net after deleting the obsolete `_fillMeta` tests and adding 11 new property + subscriber tests)
- [x] `pnpm lint` — clean
- [x] `pnpm knip` — clean

### DIFF
--- a/src/core/cqrs/events/binEvents.ts
+++ b/src/core/cqrs/events/binEvents.ts
@@ -53,12 +53,21 @@ export type BinMovedFromStagingEvent = BaseDomainEvent<
   }
 >;
 
+/**
+ * `fillType`, `width`, `depth` were added in the v2 migration so the fill
+ * analytics subscriber can derive what the v1 `_fillMeta` field used to
+ * carry. v1-era persisted events have no `fillType` — analytics treats
+ * those as unknown-source fills.
+ */
 export type LayerFilledEvent = BaseDomainEvent<
   'bin.layerFilled',
   {
     readonly layerId: LayerId;
     readonly binsCreated: number;
     readonly bins: ReadonlyArray<Bin>;
+    readonly fillType?: 'uniform' | 'gaps';
+    readonly width?: number;
+    readonly depth?: number;
   }
 >;
 
--- a/src/core/cqrs/events/binEvents.ts
+++ b/src/core/cqrs/events/binEvents.ts
@@ -56,8 +56,9 @@ export type BinMovedFromStagingEvent = BaseDomainEvent<
 /**
  * `fillType`, `width`, `depth` were added in the v2 migration so the fill
  * analytics subscriber can derive what the v1 `_fillMeta` field used to
- * carry. v1-era persisted events have no `fillType` — analytics treats
- * those as unknown-source fills.
+ * carry. v1-era persisted events have no `fillType` — the subscriber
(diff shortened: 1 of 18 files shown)


andymai:
Addressed all three review findings in 43e8043:

1. **Greptile P2 (`fillGaps.ts` comment)** — reworded to explain why we keep the v1 return shape (caller compatibility) without asserting a divergence between `addedCount` and `bins.length` that doesn't exist today.
2. **Greptile P2 (gap-fill test missing assertion)** — added `expect(markFeatureUsed).toHaveBeenCalledWith('fill')` to the gap-fill case so a silent regression in the subscriber's unconditional `markFeatureUsed` call would fail the test.
3. **Copilot (`binEvents.ts` stale comment)** — updated to match actual behavior: pre-v2 events without `fillType` are ignored (not treated as "unknown-source fills") to avoid metric distortion on re-emission.

Local: 45/45 v2 + subscriber tests pass.

PR #1550, opened 2026-05-01
feat(cqrs): migrate layer commands to v2 defineCommand

andymai:
## Summary

PR 9 in the CQRS DX redesign sequence. **All 4 layer commands now flow through the v2 wrapper.**

### `layer.add`
Generates the new Layer with name and clamped height inside `handle()`. Reads `defaultLayerHeight` from the settings store (config-not-state, outside the layout aggregate snapshot).

### `layer.update`
**Per-command refactor**: v1 store action did the height-clamp inside the `setLocal` mutator (hidden side-effect of the write). v2 inlines the clamp into `handle()` so the event payload's `changes.height` always reflects the value that will actually be stored — **no replay drift between native and apply paths**.

### `layer.delete`
**Per-command refactor**: v1 event payload only carried `{layer, deletedBinCount}`. The actual displaced bin IDs were lost, so replay against a divergent layout produced the wrong cascade.
- v2 widens `LayerDeletedEvent` payload with optional `displacedBinIds: ReadonlyArray<BinId>` so `apply()` can target the exact set.
- v1 events without `displacedBinIds` keep prior behavior (heuristic: "all bins on the deleted layer").
- `projection/replay.ts` updated to use `displacedBinIds` when present, fall back otherwise.

### `layer.reorder`
Validates indices and runs the reorder collision check against the frozen layout snapshot. `apply()` does the array splice. `fromIndex === toIndex` is a no-op (success result, apply does nothing).

## Tests

5 new sibling test files (4 commands + shared `_testHelpers`). Property tests cover handle correctness, error paths, and `apply()` round-trip equivalence.

## Where this fits

- #1543–#1548 (PRs 5–8) — bin domain (10/10 commands)
- **#TBD (PR 9)** — layer domain (4/4 commands) ← this PR

Next: **PR 10** migrates `category.add` / `category.update` / `category.delete`.

## Test plan

- [x] `pnpm typecheck` — clean (no regression)
- [x] `pnpm test:run` — full suite, **10364 tests pass** (was 10347; +17 new property tests)
- [x] `pnpm lint` — clean
- [x] `pnpm knip` — clean

### DIFF
--- a/src/core/cqrs/events/layerEvents.ts
+++ b/src/core/cqrs/events/layerEvents.ts
@@ -3,7 +3,7 @@
  */
 
 import type { BaseDomainEvent } from '../types';
-import type { Layer, LayerId } from '@/core/types';
+import type { BinId, Layer, LayerId } from '@/core/types';
 
 export type LayerAddedEvent = BaseDomainEvent<'layer.added', { readonly layer: Layer }>;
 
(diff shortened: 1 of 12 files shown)


andymai:
Addressed all four review findings in ae01499:

1. **Greptile P2 (deleteLayer redundant guard)** — dropped `b.layerId !== STAGING_ID` from the displaced-bins filter; `b.layerId === id` (where id comes from layout.layers) already excludes staging since STAGING_ID is never present in that array. Comment added explaining the invariant.
2. **Greptile P2 + Copilot (reorderLayers `OK` anti-pattern)** — dropped both the unused import and the `void OK` statement.
3. **Copilot (reorderLayers no-op comment)** — fixed the lying comment: the no-op branch DOES emit an event (the runtime always does); apply short-circuits via `if (fromIndex === toIndex) return`. Matches v1 (where the v1 handler also emits unconditionally). Comment now describes the actual behavior.
4. **Copilot (addLayer.test `heightUnits` anti-pattern)** — dropped the unused import and the `void heightUnits` statement.

Local: typecheck clean, lint clean, 70/70 v2 tests pass.

PR #1554, opened 2026-05-01
feat(cqrs): migrate drawer + 5 layout-metadata commands to v2

andymai:
## Summary

PR 11 in the CQRS DX redesign sequence. Six commands migrated:
- `drawer.update` (with the per-command refactor)
- `layout.setName`
- `layout.setPrintBedSize`
- `layout.setGridUnitMm`
- `layout.setHeightUnitMm`
- `layout.setBaseplateParams`

## drawer.update — per-command refactor

- v1 store action did clamping + bin-displacement cascade inside the `setLocal` mutator (hidden side-effects). v2 inlines the clamping into `handle()` so the event payload's `changes` always reflects the value that will actually be stored.
- The displaced bin set is precomputed in `handle()` and the bin IDs are encoded in the event payload as `displacedBinIds`. `apply()` applies the drawer change AND the displacement deterministically — no subscriber needed.
- `DrawerUpdatedEvent` payload widened from `binsDisplacedToStaging: number` to also carry `displacedBinIds: ReadonlyArray<BinId>`. v1 events without the field replay as before (no bin displacement on replay).
- `projection/replay.ts` updated to use `displacedBinIds` when present.

## 5 metadata setters

`setName`, `setPrintBedSize`, `setGridUnitMm`, `setHeightUnitMm`, `setBaseplateParams`. Each `handle()` captures previous value(s), applies any clamping (matches v1: 42–500 mm for `printBedSize`, 1–200 for `gridUnitMm`, 1–50 for `heightUnitMm`, padding ≥ 0, magnet diameter 0.5–20 mm, magnet depth 0.5–10 mm), and emits the event. `apply()` installs the new value via the Immer draft.

## Tests

8 new sibling test files (6 commands + 2 shared `_testHelpers`). Property tests cover handle correctness, clamping behavior, error paths, and `apply()` round-trip.

`commandBus.test.ts` + `pipeline.test.ts` mock setup updated: now that `layout.setName` routes through v2, the mock needs `setState` (the v2 wrapper applies via `setState` rather than calling `store.setName`). pipeline.test asserts on `mockStore.layout.name` instead of the legacy `mockStore.setName` mock.

## Where this fits

- #1538 (PR 1) — library cloudShare persistence subscriber
- #1539 (PR 2) — defineCommand foundations
- #1541 (PR 3) — history.ts rename
- #1542 (PR 4) — undo via command bus
- #1543–#1548 (PRs 5–8) — bin domain (10/10)
[...]

PR #1552, opened 2026-05-01
feat(cqrs): migrate category commands to v2 defineCommand

andymai:
## Summary

PR 10 in the CQRS DX redesign sequence. **All 3 category commands now flow through the v2 wrapper.**

### `category.add`
Validates the per-layout category-count limit. Generates the new `CategoryId`, returns the full Category in the event payload.

### `category.update`
Validates exists, captures previous values for the fields being updated, emits with `{id, changes, previous}`. `apply()` does `Object.assign`.

### `category.delete`
**Per-command refactor (per Pass 3 plan): this is precondition enforcement, not a cascade.** v1 rejects when any bin references the category — v2 keeps that behavior exactly. `handle()` validates:
1. category not in use by any bin
2. layout still has at least `CATEGORIES_MIN` categories afterward
3. category exists

`apply()` filters the category out of the draft. No bin reassignment happens because deletion is blocked when any bin uses it.

## Tests

4 new sibling test files (3 commands + shared `_testHelpers`). Property tests cover handle correctness, error paths (limit reached, in-use rejection, min-count, not found), and `apply()` round-trip.

## Where this fits

- #1538 (PR 1) — library cloudShare persistence subscriber
- #1539 (PR 2) — defineCommand foundations
- #1541 (PR 3) — history.ts rename
- #1542 (PR 4) — undo via command bus
- #1543–#1548 (PRs 5–8) — bin domain (10/10)
- #1550 (PR 9) — layer domain (4/4)
- **#TBD (PR 10)** — category domain (3/3) ← this PR

**Layout-aggregate progress: 17 / ~22 commands migrated.** Remaining layout-aggregate commands: `drawer.update`, `layout.setName`, `layout.setPrintBedSize`, `layout.setGridUnitMm`, `layout.setHeightUnitMm`, `layout.setBaseplateParams`, `layout.restore`. PR 11 will tackle drawer + layout metadata commands.

## Test plan

- [x] `pnpm typecheck` — clean (no regression)
- [x] `pnpm test:run` — full suite, **10374 tests pass** (was 10364; +10 new property tests)
- [x] `pnpm lint` — clean

PR #1557, opened 2026-05-01
feat(cqrs): migrate 5 simpler library commands to v2 (library aggregate)

andymai:
## Summary

PR 13 in the CQRS DX redesign sequence. **First commands to exercise the v2 runtime's library-aggregate path** — apply receives `Draft<LayoutLibrary>` via `useLibraryStore.setState`, mirroring how the layout aggregate works.

## Migrated

- **`library.createEntry`** — generates LayoutId, builds full entry, brands the preview's dimension fields (Zod payload arrives as plain numbers). Event payload widened with optional `entry: LayoutEntry` so apply() can push deterministically; v1 events without it replay as before.
- **`library.deleteEntry`** — rejects deleting the last remaining entry. apply() filters and advances activeLayoutId when the deleted entry was active (matches v1 cascade).
- **`library.switchActive`** — captures previousLayoutId for undo replay. No validation that target id exists (matches v1's lenient behavior; the layout-activation hook handles lookup downstream).
- **`library.updateEntry`** — partial updates (name, modifiedAt, preview, author, forkedFrom). Names truncated to NAME_MAX_LENGTH inside handle.
- **`library.setAuthorName`** — captures previousName, truncates inside handle.

`LibraryEntryCreatedEvent` payload widened with optional `entry` for v2 apply() determinism. v1 events keep prior behavior.

## Tests

6 new sibling test files (5 commands + shared `_testHelpers`). Property tests cover handle correctness, error paths, apply round-trip.

## Where this fits

- #1543–#1548 (PRs 5–8) — bin domain (10/10)
- #1550 (PR 9) — layer domain (4/4)
- #1552 (PR 10) — category domain (3/3)
- #1554 (PR 11) — drawer + layout metadata (6/6)
- #1556 (PR 12) — telemetry exit (10 ui.* commands → trackEvent)
- **#TBD (PR 13)** — 5 simpler library commands ← this PR

**Library domain status:** 5/10 commands now on v2. Remaining for PR 14: duplicateEntry, setCloudShare, clearCloudShare, renameEntry, importLayout.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm test:run` — full suite, **10408 tests pass** (was 10391; +17 new property tests)
- [x] `pnpm lint` — clean
- [x] `pnpm knip` — clean

### DIFF
--- a/src/core/cqrs/events/libraryEvents.ts
+++ b/src/core/cqrs/events/libraryEvents.ts
@@ -6,11 +6,17 @@
  */
 
 import type { BaseDomainEvent } from '../types';
-import type { LayoutId, CloudShareInfo } from '@/core/types';
+import type { LayoutEntry, LayoutId, CloudShareInfo } from '@/core/types';
 
+/**
+ * `entry` was added in the v2 migration so apply() can push the full
+ * LayoutEntry deterministically. v1-era persisted events have only
+ * {layoutId, name} — replay against those would need to reconstruct the
+ * entry from defaults, which is acceptable for the audit trail but lossy.
+ */
 export type LibraryEntryCreatedEvent = BaseDomainEvent<
   'library.entryCreated',
-  { readonly layoutId: LayoutId; readonly name: string }
+  { readonly layoutId: LayoutId; readonly name: string; readonly entry?: LayoutEntry }
 >;
 
 export type LibraryEntryDeletedEvent = BaseDomainEvent<
--- a/src/core/cqrs/v2/domain/library/_testHelpers.ts
+++ b/src/core/cqrs/v2/domain/library/_testHelpers.ts
@@ -1,0 +1,32 @@
+/**
+ * Shared fixtures for v2 library command property tests.
+ */
+
+import type { LayoutEntry, LayoutLibrary } from '@/core/types';
+import { layoutId } from '@/core/types';
+
+export function makeEntry(idStr: string, name = idStr): LayoutEntry {
+  return {
+    id: layoutId(idStr),
+    name,
+    createdAt: 1000,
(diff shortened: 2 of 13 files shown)


andymai:
Addressed all 7 Copilot review findings in 0dffedd. Root cause of most was schema drift between v2 `defineCommand` schemas and the central `validation/librarySchemas.ts` — the middleware strips/rejects fields not in the central schema, so v2 fields that didn't match were unreachable at runtime.

**createEntry.ts**:
1. Dropped `author` from v2 payload (central schema doesn't accept it; v1's CQRS handler doesn't pass author either — store action's author param is unreachable via the bus). Author always falls back to `settings.authorName`.
2. Added structural runtime guard (`isStructuredPreview`) for the `preview` field; central schema treats it as `unknown`, so malformed shapes would crash inside handle. Falls back to `DEFAULT_PREVIEW`.
3. Tightened `name` to `min(1).max(NAME_MAX_LENGTH)` matching central; ALSO truncate inside handle so the event records what apply will install (defends against schema drift / direct test-tool usage).
4. `apply()` now reconstructs a best-effort entry from `{layoutId, name}` + defaults when v1-era events arrive without `entry` (instead of silently no-op'ing).

**updateEntry.ts**:
- Dropped `modifiedAt` and `forkedFrom` from v2 payload — central schema doesn't accept them; they're effectively v1 store-internal fields not CQRS-reachable.
- Tightened `name` to match central.

**setAuthorName.ts**:
- Tightened `name` to `min(1).max(NAME_MAX_LENGTH)` matching central.

Local: typecheck clean, full vitest (10408 tests) green, lint clean, 17/17 v2 library tests pass.

PR #1560, opened 2026-05-01
feat(cqrs): migrate remaining 5 library commands to v2 (library complete)

andymai:
## Summary

PR 14 in the CQRS DX redesign sequence. **Library domain now 10/10 on v2.**

## Migrated

- **`library.duplicateEntry`** — validates source exists, generates new LayoutId, builds full entry with `(copy)` suffix (truncated to NAME_MAX_LENGTH). apply() pushes — no re-fetch needed.
- **`library.setCloudShare`** — emits `library.cloudShareUpdated`. The libraryPersistence subscriber (PR 1) listens for that event and persists the library snapshot, so end-to-end behavior matches v1 exactly. Cast through `unknown` for `CloudShareInfo` (central schema is loose so type doesn't guarantee full shape).
- **`library.clearCloudShare`** — same pattern; emits `library.cloudShareCleared`.
- **`library.renameEntry`** — captures previousName for undo replay, truncates inside handle.
- **`library.importLayout`** — creates entry from imported Layout via `computePreview`; emits the same `library.entryCreated` event as createEntry so downstream subscribers (analytics, persistence) handle imports identically.

## Tests

5 new sibling test files. Property tests cover handle correctness (source-not-found for duplicate, missing-id for cloud share, name truncation), apply round-trip, and the libraryPersistence subscriber interaction (no-op when layoutId unknown).

## Where this fits

| Status | Domain | PR |
|---|---|---|
| ✅ | bin (10/10) | #1543–#1548 |
| ✅ | layer (4/4) | #1550 |
| ✅ | category (3/3) | #1552 |
| ✅ | drawer + layout metadata (6/6) | #1554 |
| ✅ | library (10/10) | #1557 + this PR |
| ✅ | ui.* removed from CQRS | #1556 |
| ⚠️ | designer.save | still v1 (intentional) |
| ⚠️ | layout.restore | still v1 (replace-entire-root doesn't fit apply-on-draft) |

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm test:run` — full suite, **10422 tests pass** (was 10408; +14 new property tests)
- [x] `pnpm lint` — clean
- [x] `pnpm knip` — clean

### DIFF
--- a/src/core/cqrs/events/libraryEvents.ts
+++ b/src/core/cqrs/events/libraryEvents.ts
@@ -24,9 +24,18 @@ export type LibraryEntryDeletedEvent = BaseDomainEvent<
   { readonly layoutId: LayoutId }
 >;
 
+/**
+ * `entry` was added in the v2 migration so apply() can push the duplicated
+ * entry deterministically. v1-era persisted events have only the ids — the
+ * `entry?` shape lets v1 events still satisfy the type.
+ */
 export type LibraryEntryDuplicatedEvent = BaseDomainEvent<
   'library.entryDuplicated',
-  { readonly sourceLayoutId: LayoutId; readonly newLayoutId: LayoutId }
+  {
+    readonly sourceLayoutId: LayoutId;
+    readonly newLayoutId: LayoutId;
+    readonly entry?: LayoutEntry;
+  }
 >;
 
 export type LibraryActiveSwitchedEvent = BaseDomainEvent<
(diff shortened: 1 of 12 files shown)


andymai:
Both Copilot findings fixed in afcc1ff:

1. **`LibraryEntryDuplicatedEvent` type widened** — added optional `entry: LayoutEntry` so typed subscribers/migrations see what's actually emitted. Same pattern as `LibraryEntryCreatedEvent`.
2. **`importLayout.apply()` fallback** — reconstructs a best-effort entry from `{layoutId, name}` + defaults when `event.payload.entry` is missing (v1-era replay). Mirrors `createEntry.apply()`.

Local: typecheck + lint clean, 445/445 cqrs tests pass.

PR #1568, opened 2026-05-01
fix(baseplate): scale orbit zoom-out cap to baseplate size

andymai:
## Summary

`OrbitControls.maxDistance` in the baseplate preview was hardcoded to `800mm`, which clamped before the natural framing distance for any baseplate above ~10×10 grid units. A 16×16 baseplate with 50mm padding/side needs ~2,200mm to fit on screen — users couldn't pull back far enough to view the whole part.

- Derive `maxDistance` from the framing distance: `max(800, idealDistance × 3)`. Floor preserves zoom-out feel for small baseplates; the multiplier gives headroom past framing for large ones.
- Bump camera `far` plane from `2000` → `20000` so geometry isn't clipped at the new zoom-out range. (Static, not reactive — `<Canvas camera>` only applies on creation.)
- Reuses the existing `calculateIdealDistance` math from `cameraUtils.ts`; no new geometry code.

## Test plan

- [x] `pnpm run test:run` — full suite passes (10,391 tests)
- [x] `pnpm exec eslint` on changed files — clean
- [x] New regression test in `cameraUtils.test.ts` asserts max-size baseplate (16×16, 50mm padding/side) can be framed past the old 800mm cap
- [ ] Manual: open baseplate generator, set 16×16 grid, verify zoom-out reaches a full overview

### DIFF
--- a/src/features/baseplate/components/BaseplatePreview/BaseplatePreview.tsx
+++ b/src/features/baseplate/components/BaseplatePreview/BaseplatePreview.tsx
@@ -8,7 +8,7 @@
  * The slab extends asymmetrically when padding differs per side.
  */
 
-import { useRef, useCallback, useState } from 'react';
+import { useRef, useCallback, useMemo, useState } from 'react';
 import { Canvas } from '@react-three/fiber';
 import { OrbitControls } from '@react-three/drei';
 import { useShallow } from 'zustand/react/shallow';
@@ -25,6 +25,7 @@ import { useResponsive } from '@/shared/hooks/useResponsive';
 import { useSettingsStore } from '@/core/store';
 import { useTranslation } from '@/i18n';
 import type { CameraPreset } from './cameraUtils';
+import { calculateIdealDistance, calculateMaxOrbitDistance } from './cameraUtils';
 import { BaseplateMesh } from './BaseplateMesh';
 import { SceneLighting } from './SceneLighting';
 import { CameraController } from './CameraController';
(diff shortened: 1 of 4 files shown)


andymai:
Pushed [`e3110987e`](https://github.com/andymai/gridfinity-layout-tool/commit/e3110987e) addressing the static-`far`-plane feedback from Greptile and Copilot:

- Confirmed the page actually allows up to 50×50 grid units with 100mm padding/side (`CONSTRAINTS.GRID_MAX = 50`, `PADDING_MAX = 100`); my earlier "16×16 max" was conflating it with `MAX_BASEPLATE_DIMENSION`, which only governs split-tiling. Copilot was right that `far: 20000` clipped at the true upper bound.
- Added `calculateFarPlane(maxOrbitDistance) = max(2000, maxOrbitDistance × 1.5)`; the 1.5× leaves room for the slab's bounding radius once the camera reaches `maxOrbitDistance`.
- `CameraController` now updates `camera.far` reactively via a `useEffect` (with `updateProjectionMatrix()` + `invalidate()`); kept the `<Canvas camera>` static initial value at 20000 so the first frame is sane while the effect adjusts upward for extreme cases.
- Widened the regression test to use the true upper bound (50×50, 100mm padding/side) and added a frustum-containment test that asserts `far > maxOrbitDistance + boundingRadius` at that case.

(Side note for posterity: the React Compiler's `react-hooks/immutability` rule fires on direct property writes to hook-returned values like `camera.far = X`, even though the same shape works in `Scene.tsx`. Wrapping the mutation in a free-function helper keeps it lint-clean and is the only reason `setCameraFar` exists.)

PR #1593, opened 2026-05-07
feat(sync): add server endpoints for layouts/designs/manifest/export/account

andymai:
## Summary

PR 3 of 6 in the multi-device sync rollout. Server-only — adds the per-user sync endpoints that the client engine (PR 4) will start calling. **No UI changes**; the user-facing surface is still gated off by `SYNC_UI_ENABLED` from PR 2.

PR 1 ([#1572](https://github.com/andymai/gridfinity-layout-tool/pull/1572)) added the storage primitives. PR 2 ([#1591](https://github.com/andymai/gridfinity-layout-tool/pull/1591)) added the auth foundation (cookie sessions, `requireSession`, CSRF defense). PR 3 builds on both.

### Endpoints (`api/sync/`)

| Endpoint                    | Method | Rate Limit  | Behavior                                          |
| --------------------------- | ------ | ----------- | ------------------------------------------------- |
| `/api/sync/layouts/[id]`    | GET    | 240/min     | Fetch envelope (200 / 404 / 410)                  |
| `/api/sync/layouts/[id]`    | PUT    | 60/min      | LWW write; 409 stale-write, 410 stale-resurrect   |
| `/api/sync/layouts/[id]`    | DELETE | 60/min      | Tombstone + blob delete                           |
| `/api/sync/designs/[id]`    | GET/PUT/DELETE | 60–240/min | Same state machine for Bin Designer designs       |
| `/api/sync/manifest`        | GET    | 240/min     | Per-user index; `If-Modified-Since` → 304 (cheap poll path) |
| `/api/sync/export`          | GET    | 240/min     | ZIP of live items + manifest.json                 |
| `/api/sync/account`         | DELETE | 60/min      | Cascade-delete sessions + KV + blobs              |

All endpoints rate-limit per `userId`, not per IP — each authenticated user gets their own budget.

### Lib additions (`api/lib/`)

- **`userIndex.ts`** — `getIndex` / `upsertEntry` / `tombstone`. Backed by `HSET users:{uid}:index:{kind}`. Each write atomically bumps `users:{uid}:indexUpdatedAt` so `/api/sync/manifest` can serve `If-Modified-Since` 304s without scanning the hash.
- **`quota.ts`** — `checkQuota(redis, uid, kind, intent)`. 100 items + 10 MB per kind, tombstones excluded. Returns a discriminated union (`{ ok: true } | { ok: false, error }`); the api/ side doesn't share `Result` with src/, so this stays a local type.
- **`rateLimit.ts`** — added `sync.write` (60/min) and `sync.read` (240/min). Generalized the parameter name from `ip` → `scope` since sync callers pass `userId`; existing share/auth callers unchanged.

## LWW + tombstone state machine (PUT)

```
existing.modifiedAt >= req      → 409 Conflict (return stored envelope)
tombstone.deletedAt >= req      → 410 Gone   (stale resurrect blocked)
no existing OR newer than ts    → write succeeds (resurrection clears tombstone)
quota check happens AFTER LWW   → rejected writes don't consume a slot
```

Tested across all 8 quadrants (only-local, only-remote, both-newer / both-older, both-tombstoned, etc.).

## Cascade-delete is idempotent

`DELETE /api/sync/account` runs:

1. `SMEMBERS users:{uid}:sessions` → `DEL session:{token}` for each
2. `HKEYS users:{uid}:index:*` → `del()` each blob
3. `DEL users:{uid}:*` (indexes, profile, sessions set, indexUpdatedAt)
4. Clear session cookie on responding device

Each step uses unconditional `DEL`, so a partial-failure replay (Vercel function timeout, cold-start mid-cascade) is safe. Per-blob errors are logged but don't block — leftover blobs are storage cost only, not correctness.

## Decisions worth flagging

- **Content filtering deliberately skipped.** The existing share endpoints filter content because shares are publicly URL-accessible; sync data is user-private (your own layouts and designs). Filtering would cause false-positive blocks on a user's own bin labels with no security benefit. Documented in `api/sync/README.md`.
- **Designer payload reuses `validateDesignerShare`** — the sync envelope wraps the design as `{ type: 'designer', version: 1, params }` internally to satisfy the existing validator. Keeps one source of truth for BinParams validity. Unwrapped to `{ design, modifiedAt }` shape on the wire.
- **Quota under concurrent writes is a soft ceiling.** Two tabs racing on different items can both pass quota and both write, briefly exceeding the cap by one item. Acceptable at this scale; documented inline. A Lua-script atomic reservation isn't worth the complexity.
- **Item ids accept the same three formats as share ids** (UUID, base36 timestamp, 12-char alphanumeric). Lets layouts round-trip between local storage / share / sync without renaming.

## Test plan

- [x] `pnpm run lint` — clean
- [x] `pnpm run knip` — clean
- [x] Targeted tests pass: `pnpm exec vitest run api/sync api/lib api/auth` — 242/242
  - `api/lib/userIndex.test.ts` — 7/7 (round-trip, tombstone, malformed-JSON tolerance)
  - `api/lib/quota.test.ts` — 8/8 (count cap, byte cap, replace, tombstone exclusion, kind isolation)
  - `api/sync/layouts/[id].test.ts` — 15/15 (full LWW × tombstone grid, id validation, rate-limit, auth)
  - `api/sync/designs/[id].test.ts` — 6/6 (designer-specific validation, round-trip)
  - `api/sync/manifest.test.ts` — 6/6 (304 cache, manifest contents, malformed header tolerance)
  - `api/sync/export.test.ts` — 4/4 (empty zip, populated zip, tombstone exclusion, rate-limit)
  - `api/sync/account.test.ts` — 5/5 (full cascade, idempotent replay, per-blob failure tolerance)
- [ ] Manual verification post-deploy (with PR 2 cookie): see `api/sync/README.md` curl examples.

## Out of scope (PR 4)

- Client engine — `src/core/sync/engine.ts`, outbox, poller, claim flow
- LayoutAdapter / DesignAdapter wiring
- Toast UX for offline / forced sign-out

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -38,21 +38,26 @@ graph TB
 
 ## Endpoints
 
-| Endpoint                        | Method | Rate Limit | Purpose                              |
-| ------------------------------- | ------ | ---------- | ------------------------------------ |
-| `/api/share`                    | POST   | 100/min    | Create layout or designer share      |
-| `/api/share/[id]`               | GET    | 100/min    | Fetch share with metadata            |
-| `/api/share/[id]`               | PUT    | 100/min    | Update share (requires delete token) |
-| `/api/share/[id]`               | DELETE | 100/min    | Delete share (requires delete token) |
-| `/api/report/[id]`              | POST   | 10/hr      | Report inappropriate share           |
-| `/api/liveblocks-auth`          | POST   | 100/min    | Collaborative editing session auth   |
-| `/api/ml-telemetry`             | POST   | 100/min    | Aggregated ML training data          |
-| `/api/auth/login/[provider]`    | GET    | 30/min     | Begin OAuth flow (Google / GitHub)   |
-| `/api/auth/callback/[provider]` | GET    | 30/min     | OAuth code exchange + session mint   |
-| `/api/auth/logout`              | POST   | 100/min    | Clear session cookie + KV row        |
-| `/api/auth/me`                  | GET    | 100/min    | Return signed-in user profile        |
+| Endpoint                        | Method         | Rate Limit | Purpose                                |
+| ------------------------------- | -------------- | ---------- | -------------------------------------- |
+| `/api/share`                    | POST           | 100/min    | Create layout or designer share        |
+| `/api/share/[id]`               | GET            | 100/min    | Fetch share with metadata              |
+| `/api/share/[id]`               | PUT            | 100/min    | Update share (requires delete token)   |
+| `/api/share/[id]`               | DELETE         | 100/min    | Delete share (requires delete token)   |
+| `/api/report/[id]`              | POST           | 10/hr      | Report inappropriate share             |
+| `/api/liveblocks-auth`          | POST           | 100/min    | Collaborative editing session auth     |
+| `/api/ml-telemetry`             | POST           | 100/min    | Aggregated ML training data            |
+| `/api/auth/login/[provider]`    | GET            | 30/min     | Begin OAuth flow (Google / GitHub)     |
+| `/api/auth/callback/[provider]` | GET            | 30/min     | OAuth code exchange + session mint     |
+| `/api/auth/logout`              | POST           | 100/min    | Clear session cookie + KV row          |
+| `/api/auth/me`                  | GET            | 100/min    | Return signed-in user profile          |
+| `/api/sync/layouts/[id]`        | GET/PUT/DELETE | 60–240/min | Per-user layout sync (LWW + tombstone) |
+| `/api/sync/designs/[id]`        | GET/PUT/DELETE | 60–240/min | Per-user Bin Designer design sync      |
+| `/api/sync/manifest`            | GET            | 240/min    | Per-user index + If-Modified-Since 304 |
+| `/api/sync/export`              | GET            | 240/min    | ZIP of all live items + manifest       |
+| `/api/sync/account`             | DELETE         | 60/min     | Cascade-delete account + KV + blobs    |
 
-See [`auth/README.md`](./auth/README.md) for the OAuth setup, cookie shape, and CSRF posture.
+See [`auth/README.md`](./auth/README.md) for the OAuth setup and [`sync/README.md`](./sync/README.md) for sync semantics, quotas, and the LWW + tombstone state machine.
 
 ## Validation Library (`lib/`)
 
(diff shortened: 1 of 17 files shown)


andymai:
Pushed [a60f59a82](https://github.com/andymai/gridfinity-layout-tool/commit/a60f59a82) addressing the review feedback.

**Fixed (8 of 10 findings):**

| # | Source | Issue | Fix |
|---|--------|-------|-----|
| G3 | Greptile | `handleDelete` calls `deleteBlob` on already-tombstoned entries (layouts) | Short-circuit when `existing.deletedAt !== undefined` |
| G4 | Greptile | Same on designs | Same fix |
| G5 | Greptile | Redundant `getRedis()` in `handleGet` | Pass `redis` as parameter, both endpoints |
| C1 | Copilot | `!existing.deletedAt` truthy-check on numeric field (layouts) | `existing.deletedAt === undefined` |
| C2 | Copilot | Same on designs | Same fix |
| C3 | Copilot | Designs `sizeBytes` mismatched between validator and storage | Build `validationPayload` once; size everything against it |
| C4 | Copilot | `parseEntry` accepts NaN/Infinity for numeric fields | `Number.isFinite` (matches `getIndexUpdatedAt`) |
| C5 | Copilot | Redundant explicit `checkCsrfDefense` in `account.ts` | Dropped — `requireSession` calls it internally |

**False positives (G1, G2):** "CSRF defense missing on `layouts/[id].ts` and `designs/[id].ts` PUT/DELETE."

The check IS performed — `requireSession` (api/lib/session.ts:99 from PR 2) calls `checkCsrfDefense` as its first action:

```ts
export async function requireSession(req, res): Promise<SessionRecord | null> {
  if (!checkCsrfDefense(req, res)) return null;
  // ...
}
```

So `if (!session) return;` short-circuits both for failed auth (401) AND failed CSRF (403). The endpoints are protected, just not via an explicit call site Greptile could grep. I considered making it explicit at every endpoint for visibility, but that would create real duplication (`requireSession` runs the same check seconds later) and a foot-gun where someone could call `requireSession` and forget the explicit check. Net: leaving the architecture as-is, fixing C5 to remove the redundancy in `account.ts`.

**Regression tests added (2):**
- `idempotent on repeated DELETE: skips deleteBlob on already-tombstoned entries` (asserts `deleteBlob` mock call count is unchanged on the second DELETE)
- `parseEntry rejects NaN/Infinity values for numeric fields`

226/226 tests pass; quality clean.

PR #1595, opened 2026-05-07
feat(sync): client engine foundation — adapters, outbox, status store

andymai:
## Summary

PR 4 of 6 in the multi-device sync rollout, **split into reviewable parts**. This is **part A — foundation only**. No engine, no triggers, no poller — those land in 4b/4c.

Nothing here is reachable from the UI in production: `SYNC_UI_ENABLED` is still off (PR 2). All of this is dormant code that the next PRs will wire up.

**Verified post-build:** `grep -E '(useSyncStatusStore|outbox|layoutAdapter|gridfinity-sync)' dist/assets/main-*.js` returns empty.

### Adapter abstraction (`src/core/sync/adapters/`)

```
adapters/
  types.ts          — SyncAdapter, LayoutAdapter, DesignAdapter,
                      AdapterChange, SyncableItem
  layoutAdapter.ts  — concrete LayoutAdapter for layouts
  layoutAdapter.test.ts
```

The engine (PR 4b) depends only on these interfaces — no direct imports of `LayoutManager` or feature-side storage. This is the seam that lets the engine push design changes (which live in `src/features/bin-designer/`) without violating the project's `core/` ↛ `features/` module-boundary rule.

`layoutAdapter` notable bits:
- `applyRemote` uses `importLayout(layout, id, 'remote')` so the store's existing `lastEditSource` flag distinguishes remote-applied changes from user edits.
- Echo suppression via a single-tick `Set<string>` + `queueMicrotask` so the library-store subscriber the engine installs in PR 4b doesn't see its own remote-write as a fresh local change.

### Outbox (`src/core/sync/outbox.ts`)

IndexedDB-backed pending-push queue. Separate DB from layouts / events / designer to isolate lifecycles.

| Property | Choice |
|---|---|
| Schema | One row per `(kind, id)` — re-enqueues **upsert** |
| Backoff | Exponential: 1s, 2s, 4s, ... capped at 5min |
| Max attempts | 8 (then \"gave-up\", caller surfaces toast) |
| Survives reload | Yes |
| Race protection | `markSuccess` only deletes when stored mtime matches pushed mtime |

The race protection is the subtle bit: if the user edits while a slow push is in flight, the second `enqueue` upserts the row to the new mtime. When the slow push completes and calls `markSuccess(kind, id, oldMtime)`, the stored mtime no longer matches — so the row stays for the next drain cycle. Without this, a user editing during a push would lose their newest change.

### Status store (`src/core/sync/status.ts`)

```ts
{ state: 'idle' | 'syncing' | 'offline' | 'error',
  lastSyncedAt?, pendingCount, lastError? }
```

- `succeed()` returns to `idle` **only when `pendingCount === 0`** — single coherent transition rather than per-item flicker.
- Settings → Account panel (PR 6) reads this; trigger hooks in 4b/4c will call `beginSync` / `succeed` / `reportOffline` / `reportError`.

## What's NOT in this PR (deferred to 4b/4c/4d)

- Engine coordinator
- DesignAdapter (touches `src/features/bin-designer/`)
- 4 trigger hooks (debounced push, visibility flush, beacon flush, periodic poll)
- Poller (the 8-quadrant diff logic)
- Boot wiring (extending SyncSessionMount to start the engine)
- Toast UX for offline / quota / 410 / 409

## Test plan

- [x] `pnpm run lint` — clean
- [x] `pnpm run knip` — clean
- [x] Targeted tests: `pnpm exec vitest run src/core/sync` — 57/57
  - outbox.test.ts — 18 (enqueue/upsert, getDue ordering, markSuccess race protection, exponential backoff, MAX_ATTEMPTS give-up, persistence)
  - status.test.ts — 9 (state transitions, pendingCount-aware succeed, reset)
  - adapters/layoutAdapter.test.ts — 8 (list, get, subscribe diff for add/update/delete, no-op suppression, unsubscribe)
- [x] `pnpm run build` — verified zero sync surface in `main-*.js` (only the locale chunks contain unrelated string literals).

### DIFF
--- a/src/core/README.md
+++ b/src/core/README.md
@@ -17,14 +17,14 @@ graph TB
 
 ## Subdirectories
 
-| Directory  | Purpose                                                                                   |
-| ---------- | ----------------------------------------------------------------------------------------- |
-| `api/`     | Cloud sharing API client (`share.ts`, `suggestName.ts`)                                   |
-| `labs/`    | Feature flag definitions and types (experimental/preview/graduated)                       |
-| `result/`  | `Result<T, E>` type system — constructors, error catalog, combinators                     |
-| `storage/` | Layout persistence — LayoutManager, LayoutService, ShareService, backends                 |
-| `store/`   | Zustand + Immer stores — layout, library, history, selection, interaction, view, settings |
-| `sync/`    | Multi-device sync session: `apiFetch`, `useSession`, `sessionApi` (engine lands later)    |
+| Directory  | Purpose                                                                                                 |
+| ---------- | ------------------------------------------------------------------------------------------------------- |
+| `api/`     | Cloud sharing API client (`share.ts`, `suggestName.ts`)                                                 |
+| `labs/`    | Feature flag definitions and types (experimental/preview/graduated)                                     |
+| `result/`  | `Result<T, E>` type system — constructors, error catalog, combinators                                   |
+| `storage/` | Layout persistence — LayoutManager, LayoutService, ShareService, backends                               |
+| `store/`   | Zustand + Immer stores — layout, library, history, selection, interaction, view, settings               |
+| `sync/`    | Multi-device sync foundation: session, `outbox` (IDB queue), `status` store, `adapters/` (engine in 4b) |
 
 ## Key Files
 
--- a/src/core/sync/adapters/layoutAdapter.test.ts
+++ b/src/core/sync/adapters/layoutAdapter.test.ts
@@ -1,0 +1,143 @@
+/**
+ * Focused tests for the layoutAdapter's read paths and the change-event
+ * suppression on `applyRemote`. Engine-level integration coverage of the
+ * full apply-remote flow (UI re-render, library entry upsert) lives in
+ * PR 4b's engine tests, where the integration is meaningful.
+ */
+
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { useLibraryStore } from '@/core/store';
+import type { LayoutEntry, LayoutId, LayoutLibrary } from '@/core/types';
+
+const loadLayoutAsyncMock = vi.fn();
+const saveLayoutAsyncMock = vi.fn();
+const saveLibraryMock = vi.fn();
+const loadLayoutSyncMock = vi.fn();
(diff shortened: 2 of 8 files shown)


andymai:
Pushed [65eadd847](https://github.com/andymai/gridfinity-layout-tool/commit/65eadd847) addressing all 9 review findings.

| # | Source | Issue | Fix |
|---|--------|-------|-----|
| C1 | Copilot | \`preview: undefined\` runtime crash on remote-applied LayoutEntry | Use \`computePreview(layout)\` for both insert and update paths |
| C2 | Copilot | \`saveLibrary\` returns Result; errors silently swallowed in \`applyRemote\` | \`isErr(result) → throw\` so the engine surfaces a toast |
| C3 | Copilot | Same \`saveLibrary\` issue in \`applyRemoteDelete\` | Same fix |
| C4 | Copilot | \`LayoutAdapter\` defaulted to \`SyncAdapter<unknown>\` | Now \`SyncAdapter<Layout>\`; \`DesignAdapter\` stays generic since \`BinParams\` lives in \`features/\` |
| C5 | Copilot | Backoff off-by-one — first failure scheduled 2s not 1s | Pre-increment value (\`backoffDelayMs(current.attempts)\`); test updated |
| C6 | Copilot | Comment said \"macrotask\" but code uses \`queueMicrotask\` | Comment fixed; documents the microtask boundary's interaction with Zustand's synchronous listeners |
| G1 | Greptile P2 | Same comment mismatch | Same fix as C6 |
| G2 | Greptile P2 | \`SyncKind\` plural vs \`OutboxItemKind\` singular | Dropped \`OutboxItemKind\`; outbox imports + re-exports \`SyncKind\` from \`adapters/types\`. No translation shim anywhere now |
| G3 | Greptile P2 | Status store didn't use \`immer\` per project convention | Migrated to \`create<State>()(immer((set) => ...))\` pattern; setter bodies use draft mutation |

**Verification:**
- 57/57 tests still pass (1 test updated to match corrected backoff sequence)
- \`pnpm run quality\` (typecheck + lint + knip) clean
- Production bundle still has zero sync surface (verified post-build)

PR #1648, opened 2026-05-13
feat(baseplate): rotational-symmetric split pieces

### LINKED ISSUE #1640: Reducing number of unique pieces
**Problem**
Right now, the way the tool splits large grids and the way it adds dovetails causes it to create several unique parts. A small change to the way it does these things can make it take advantage of rotational symmetry so that fewer unique models are created, allowing for things like stacking during prints.

**Proposed Solution**
Here’s an example of what I mean. My x1c is too small for the drawer, so it ends up split into six parts:

Image

It should be possible to make four corners and two inner pieces, by making the corners the same size, and the inner pieces narrower.

Furthermore, if the dovetail was made male on one edge, and female on the other.

With these changes, each copy of the part can rotate into the appropriate corner. A1, A2, C1, and C2 would all be the same part, each rotated 90°. B1 and B2 can similarly be the same part, rotated 180°.

One complication: the margin around the perimeter to fill in the non-integral width and height of the drawer would have to be a separate piece.

Does that make sense? If I could draw, I'd draw what I mean.

andymai:
## Summary

Closes #1640.

Opt-in `preferIdenticalPieces` toggle that lets opposite-corner pieces of a split baseplate share a single canonical mesh under 180° rotation. For a 3×2 split this cuts the unique-design count from 6 down to 3.

How it works:

- **Split planner** now arranges integer-sized chunks palindromically on each axis, so outermost corners match across the tiling.
- **Dovetail connectors** drop a tongue + groove pair at every cell boundary (M at the clockwise-earlier side of each boundary point, F at the clockwise-later side). The result is a connector layout that is itself 180°-rotationally invariant — adjacent pieces still mate, and a piece rotated 180° produces the same canonical geometry.
- **`pieceFingerprint`** canonicalizes edges under L↔R / F↔B swap so 180°-rotated pairs fingerprint identically.
- **`pieceToBaseplateParams`** also swaps padding alongside edges, so the canonical mesh ends up with padding on the correct sides post-rotation.
- **3D preview** applies the per-piece rotation around the piece center via an inner group; the **print guide** annotates rotated positions with "(rotate 180°)" so the user knows which copies to flip.

Default behavior unchanged; the toggle is gated behind the existing **Dovetail connectors** switch.

Doesn't deliver 90° rotational symmetry (all 4 corners → 1 design) — that would require a genderless self-mating connector. Issue thread covers why 180° was the practical target; a follow-up could explore the 90° case.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint` (no new warnings; pre-existing `api/auth` warnings unrelated)
- [x] `pnpm run test:run -- src/features/baseplate/` — 784 files / 10931 tests pass
- [x] `pnpm run test:run src/features/generation/worker/generators/baseplateGenerator.scenario.dovetail.test.ts` — 9 tests pass including 2 new ones verifying the doubled-dovetail mesh is watertight AND has more triangles than the single-dovetail baseline
- [x] i18n checks (keys, values, interpolation, unused) all green
- [ ] Manually verify the toggle in the panel renders, persists, and rotates pieces in the 3D preview
- [ ] Manually export a split baseplate with the toggle on and confirm the print guide annotates rotated positions correctly


andymai:
@greptile-apps thanks for the careful read on this round. I dug into the suggestion and I think this one's actually correct as-shipped — pushing back rather than applying the fix.

`buildSlabProfile` only applies a radius at **exterior** corners; interior corners are always squared (radius 0). In a 2×2 split each corner piece has exactly one exterior corner, and canonical-pair partners sit at *different* exterior positions:

- A1 ↔ B2: exterior corners are **bl** vs **tr**
- A2 ↔ B1: exterior corners are **tl** vs **br**

When `parent.bl ≠ parent.tr` (asymmetric `cornerRadii`), the two pieces need physically different radii at their respective single exterior corner. They cannot share a generated mesh — rotation at placement cannot reconcile a radius of 3 at one corner with a radius of 2 at another.

Worked example with the suggested canonicalization applied to `{tl:1, tr:2, bl:3, br:4}`:
1. A1 and B2 collapse to one fingerprint.
2. A1 is generated first; its mesh has `bl=3, others=0` (only A1.bl is exterior).
3. B2 clones A1's mesh and applies 180° rotation at placement → mesh's `bl` moves to world `tr`.
4. World `tr` now has radius 3, but `parent.tr = 2`. Wrong geometry.

So the silent-degradation behavior is intentional and correct: asymmetric `cornerRadii` simply emits two STLs for these pairs. The Gotcha note in `src/features/baseplate/README.md` (commit `d55681f`) already documents this, and I just pushed `52bdc6ee` adding two regression tests to lock it in:

- Asymmetric `cornerRadii` → 4 groups (no dedup).
- 180°-symmetric `cornerRadii` (e.g. `{tl:1, tr:2, bl:2, br:1}`) → 2 groups (dedup works).

PR #1653, opened 2026-05-15
fix(generation): cut divider notches through lip in split slotted bins

### LINKED ISSUE #1652: Removable dividers with enabled stacking lip
**Problem**
The removable dividers can no longer be inserted from above when the stacking lip is enabled on the base.

Image

**Proposed Solution**
It would be better if there were a way for the divider notch in the mold wall to extend through the stacking lip.

andymai:
## Summary

- Fixes #1652 — removable dividers couldn't be inserted from above when the stacking lip was enabled on split bins
- Adds `buildLipSlotCuts()` in `slotBuilder.ts` that produces only the lip-zone cutters with a configurable position inset
- Applies the lip cuts to the separately-built `lipSolid` in `splitBinBuilder.ts` before per-piece fusing, using `slotPositionsEdgeInset=0` so cuts align with the body's wall slots

## Root cause

`splitBinBuilder` generates the bin body with `stackingLip: false` to work around an OCCT boolean-intersection crash at the lip-wall junction. The lip is then built fresh via `buildTopShape()` and fused per-piece — but never went through the slot-cut pipeline. Non-split slotted bins already cut the lip correctly via `slotCutsFeature`.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] Targeted scenario test: `binGenerator.scenario.split-slotted-lip.test.ts` (new) verifies slotted+lip+split pieces have more lip-zone vertices than unslotted equivalents, on single-axis, both-axis, and across wall thicknesses 0.95mm / 1.6mm / 2.0mm
- [x] Regression: `binGenerator.scenario.split-robustness.test.ts`, `binGenerator.scenario.split-lip-trim.test.ts`, `slotBuilder.test.ts` all pass

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.split-slotted-lip.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.split-slotted-lip.test.ts
@@ -1,0 +1,146 @@
+// @vitest-environment node
+/**
+ * Scenario tests for slotted bins with stacking lip when split (#1652).
+ *
+ * Verifies that the separately-built lip in split bins gets cut at slot
+ * positions, so removable dividers can slide in from the top without being
+ * blocked by the lip's inner overhang.
+ */
+import { describe, it, expect, beforeAll } from 'vitest';
+import { DEFAULT_BIN_PARAMS } from '@/shared/constants/bin';
+import { DEFAULT_SPLIT_CONNECTOR_CONFIG } from '@/features/bin-designer/constants/defaults';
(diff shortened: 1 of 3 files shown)


andymai:
@greptileai Addressed in af24d530 — made `slotPositionsEdgeInset` a required parameter and expanded the docstring to spell out the contract (must match the caller's body wall slot inset, split callers pass 0, non-split callers would pass `lipOverhang`). The splitBinBuilder caller already passes `0` explicitly so no caller change needed.

PR #1660, opened 2026-05-15
feat(cutouts): split scoop into per-axis radii with per-edge toggles

### LINKED ISSUE #1659: Scoop on Cutout editor split axis
**Problem**
At the moment we can only select a single scoop dimension that applies to both axis.
It does work well but if I have a cilinder shape that I want to store, the aditional axis will push the cilinder to the middle or in the edge case I have to add an aditional grid slot to accomodate the extra curve.

Image

You can see in the printscreen that the bigger scoop has its inner lines much closer to each other on both axis.

**Proposed Solution**
Split the scoop with an advanced toggle that lets you select how much scoop in each axis you want

**Alternatives Considered**
If spliting to create strange combinations is hard to implement, it can also be implemented with a disable scoop on specific walls, where the curve would then meet the wall.

andymai:
## Summary

Resolves #1659. Cutouts now have separate **Scoop W** and **Scoop D** radii plus per-edge toggles instead of a single uniform `scoopRadius`. A long, narrow cutout can scoop only its short ends — letting a cylinder sit flush against the long walls instead of being pushed to the center.

- **Rectangle (ungrouped)**: full split + four edge toggles (L/R/F/B in cutout-local frame)
- **Rectangle (grouped)**: per-member axis radii; per-edge toggles hidden (ambiguous on shared edges)
- **Circle / path**: uniform single slider; W/D collapse to one value

## Architecture notes

- Split out `buildUnrotatedCutoutShape` so the fillet runs in the cutout's canonical local frame; axis classification by edge direction (Y-aligned → `scoopRadiusW`, X-aligned → `scoopRadiusD`) is trivial there. Rotation applies afterward.
- Grouped cutouts use the existing per-edge brepjs callback; for owned edges we rotate the direction by `-member.rotation` to canonical frame before classifying. Junction edges (shared between members) keep the existing `JUNCTION_RADIUS_FACTOR` discount, applied to the max axis radius of the involved members.
- `scoopEdges.{left, right, front, back}` gate individual walls via `null` returns from the callback (OCCT skips them).
- Helpers extracted to `cutoutScoopHelpers.ts` to keep `cutoutBuilder.ts` under the 500-line lint cap.

## Migration

`migrateParams` copies legacy `scoopRadius` into both `scoopRadiusW` and `scoopRadiusD` once. Idempotent — re-loading a migrated design is a no-op.

## i18n

Added 9 new keys across all 9 locales: `scoopW`, `scoopD`, `scoopSplit`, `scoopUniform`, `scoopEdges`, `scoopEdge{Left,Right,Front,Back}`. Dutch "Uniform" is a loanword and allowlisted.

## Test plan

- [x] `defaults.test.ts`: migration round-trip + idempotency
- [x] `CutoutScoopControls.test.tsx`: rectangle/circle/group UI gating, expand/collapse, edge toggles
- [x] `FloatingInspector.test.tsx`: existing assertions updated for new label keys
- [x] Scenario tests: split scoop (W-only), edge gate disabling one wall, asymmetric W/D group
- [x] Snapshot tests record valid triangle counts for new scenarios
- [x] `pnpm run typecheck`, `pnpm run lint`, `pnpm run check:i18n`, `pnpm run check:i18n:values` clean
- [ ] Manual: load an old design saved with `scoopRadius: 5`, confirm scoop renders identically; tweak split values and verify only the targeted walls curve


andymai:
## Addressed PR review feedback

### Critical (correctness)
- **`classifyAxisRadius` rotation bug** (Greptile + Copilot) — switched from bounding-box-extent rotation to midpoint-position-in-local-frame classification. Bounding-box extents lose sign, so a 30°-rotated W-edge and D-edge collapsed onto the same axis. Now classifies by which wall the edge midpoint sits on, which is robust at any rotation. Function signature now takes `memberCenter`.
- **Stale `expanded` state in `CutoutScoopControls`** (Greptile + Copilot) — switched from a `useEffect` (which tripped `react-hooks/set-state-in-effect`) to `key={cutout.id}` at both usage sites. Component remounts on cutout swap so the auto-expand initializer re-runs.

### High (per-edge intent leaks)
- **Progressive fallback dropped per-edge gates** (Copilot, 2 spots) — new `applyCallbackFilletWithFallback` scales the per-edge callback's radii uniformly through fallback steps instead of falling back to a uniform `max(W, D)` that re-enabled disabled walls.
- **Corner arcs bypassed edge gates** (Copilot) — corner arcs on rounded rectangles now require BOTH adjacent walls enabled to round. Without that, disabling "left" still curved the front-left and back-left corners.

### Medium
- **Multi-select circle/path inconsistency** (Copilot) — new `handleScoopAxisBatch` writes both axes for circles/paths in batch updates so the slider stays meaningful (the generator collapses W/D to `max()` for non-rectangles).
- **Accessibility** (Copilot) — edge-toggle buttons now carry localized `aria-label` ("Toggle scoop on left wall", etc.) across all 9 locales in addition to the visible L/R/F/B chips.
- **Stale docstring** (Copilot) — corrected the `CutoutScoopControls` header to match the actual gating: rectangle gets split + edges (ungrouped) or split only (grouped); circle/path uniform only.
- **Missing rotation coverage** (Copilot) — added a grouped scenario with a 30°-rotated member + asymmetric W/D to exercise the new classifier.

### Refactor cleanup (from code-simplifier pass)
- Dropped the `axisAware` flag in `applyAdaptiveScoop` (always passed)
- Tightened `maxOwnerAxisRadius` signature (required `memberScoops`)
- Removed file-top "kept under 500 lines" metacomment + 2 narration comments

### Verification
- `pnpm run typecheck`, `pnpm run lint`, `pnpm run check:i18n`, `pnpm run check:i18n:values` all clean
- 115/115 targeted UI + migration tests pass
- 201/201 scenario tests pass; snapshot updated for `edge gate disables a single wall` (2894 → 2862, reflecting the corner-arc gate fix)

PR #1679, opened 2026-05-16
fix(bin-designer): roll back per-corner lip UI to a single lip color

andymai:
## Summary

The per-corner lip color UI from #1671 isn't working correctly. This PR collapses the panel UI to a single Stacking Lip row while keeping the underlying 4-corner `LipColorConfig` schema intact — so per-corner editing can be restored without a migration when the issue is fixed.

## What changes

- **Panel UI** — `ColorsSection.tsx` drops the `LipZoneRow` (2x2 thumbnail + expand chevron) and the 4 indented corner sub-rows. The lip now renders as a plain `ColorZoneRow`, just like Body / Base / Label Tab.
- **Mirroring writer** — picking a lip color in the single picker mirrors that hex into all four corner slots via `updateFeatureColors({ lip: { frontLeft, frontRight, backRight, backLeft } })`. The 4-corner schema stays valid and the geometry pipeline keeps using its per-corner classification (all four slots simply end up the same color).
- **Canonical display** — the single picker reads `featureColors.lip.frontLeft`. Saved designs with mismatched corners are unchanged on mount; they snap to a single hex on the next user edit, no migration pass needed.
- **Hover wiring** — `ColorZoneRow` (and `ColorPicker`) had their `zone` / `onHover` types widened from `ColorZone` to `HoverableZone` so the lip row can fire the umbrella `'lip'` glow target. All other call sites pass valid `ColorZone` values which remain assignable.
- **Dead code kept** — `LipZoneRow.tsx` and `LipCornerDiagram.tsx` are intentionally left in place (added to `knip.json`'s ignore list) so restoring the per-corner UI is a small, additive change. `lipCornerClassifier`, the 3D preview, materialMapping, and the 3MF exporter are **untouched** — they still classify LIP triangles into four slots, but the slots receive identical colors.
- **Tests** — `ColorsSection.test.tsx` drops the `Lip per-corner expansion` describe block and adds a single-lip-row test asserting the writer patches every corner. `LipZoneRow.test.tsx` removed (component is dormant; its prior coverage doesn't reflect any live UI).

## What does *not* change

- `LipColorConfig` shape, `ColorZone` union (still includes `lip:frontLeft`..`lip:backLeft`), `ZONE_ORDER`, `computeActiveZones`, `getZoneColor`, `lipCornerZone`, `featureTagToColorZone`, `resolveColorMapping`.
- `LipCornerClassifier` and the geometry pipeline (preview + 3MF).
- `ColorsActionsMenu` (Match-all-to-body still writes 4 corners; Apply-palette still loads 4-corner palette data verbatim).
- Per-corner i18n keys (`binDesigner.colors.lip.frontLeft`..`backLeft`, `lip.fourCorners`, `lip.diagramAria`) — kept so locale parity holds and so restoring the UI is a one-file change.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean (1 unrelated pre-existing warning in `useMeshGeometry.test.ts`)
- [x] `npx vitest run src/features/bin-designer src/features/generation/worker/generators` — 2767 / 2767 passing
- [x] `npx knip` clean (after adding `LipZoneRow.tsx` to ignore list)
- [x] `pnpm run check:i18n` clean
- [ ] Manual smoke: stacking lip on, multi-color enabled — verify single Stacking Lip row, picker mirrors to all 4 corners in `featureColors.lip`, 3D preview lip renders as one solid color, 3MF export produces a single lip filament slot
- [ ] Manual: load a pre-PR design with mismatched corner colors — verify the picker shows the `frontLeft` hex and editing it snaps all 4 corners to the new value


andymai:
Addressed in aacdbd06: the lip-mirror test now renders `<ColorsSection />`, clicks the Stacking Lip row, clicks the Red preset, and asserts all four corners in the store equal `#ef4444`. Verified locally that injecting the exact regression you flagged (`updateFeatureColors({ lip: { frontLeft: hex } })` in the section's onChange) now produces a failing assertion.

PR #1711, opened 2026-05-18
feat(seo): add FAQPage and BreadcrumbList schema to /guide and /what-is-gridfinity

andymai:
## Summary

Closes the last two landing-page gaps in our rich-result schema coverage.

After a per-page audit, only `/guide` and `/what-is-gridfinity` were missing FAQPage and BreadcrumbList — every other landing page (`/`, `/gridfinity-generator`, `/gridfinity-bin-generator`, `/gridfinity-baseplate-generator`, `/gridfinity-sizes`) already shipped SoftwareApplication + FAQPage + BreadcrumbList + HowTo.

`scripts/build-content.ts` now supports `faqs` and `breadcrumbs` arrays in markdown frontmatter:

- **Breadcrumbs** render as a visible `<nav aria-label="Breadcrumb">` at the top of the page + matching `BreadcrumbList` JSON-LD.
- **FAQs** render as a visible `<details>` accordion section at the bottom of the page + matching `FAQPage` JSON-LD.
- Both visible HTML and JSON-LD are generated from the same frontmatter source so Google's "schema must match the visible content" requirement can't drift.

8 FAQs added to `/guide` (measuring, grid units, bin sizes, layers, print lists, empty space, printers).
9 FAQs added to `/what-is-gridfinity` (what a bin is, grid units, bin vs baseplate, filament, weight, print time, drawer behavior, buying vs printing, cost).

## Test plan

- [x] `pnpm exec tsx scripts/build-content.ts` succeeds
- [x] All JSON-LD blocks on the two pages parse as valid JSON
- [x] Each FAQ entry has a matching `<details>`/`<summary>`/answer DOM node
- [x] Existing `Article` and `HowTo` schemas preserved on the two pages
- [x] All other pages unaffected (only their `content.<hash>.css` reference bumped)
- [x] Pre-commit hooks (boundaries, i18n × 3, exhaustiveness, component structure, missing tests) pass
- [ ] Google Rich Results Test confirms FAQPage + BreadcrumbList on the two URLs after deploy
- [ ] Search Console shows no schema validation errors after the next crawl

### DIFF
--- a/content/guide.md
+++ b/content/guide.md
@@ -3,6 +3,28 @@ title: How to Plan a Gridfinity Drawer Layout
 description: A practical guide to planning Gridfinity drawer layouts. Measure your drawer, figure out what bins you need, and export a print list.
 keywords: gridfinity planner, gridfinity layout, how to plan gridfinity, drawer organizer planning, gridfinity guide
 schema: HowTo
+breadcrumbs:
+  - name: Home
+    url: https://gridfinitylayouttool.com/
+  - name: Planning Guide
+    url: https://gridfinitylayouttool.com/guide
+faqs:
+  - q: How do I measure a drawer for Gridfinity?
+    a: Measure the inside dimensions of the drawer in millimeters — width (left to right), depth (front to back), and clearance height (bottom to top with the drawer closed). Take measurements at several spots since drawers are rarely perfect rectangles, and use the smallest value for each dimension to be safe.
+  - q: How do I convert drawer dimensions to Gridfinity grid units?
+    a: Divide each dimension by 42mm and round down. For example, a 380mm × 260mm drawer fits a 9×6 grid (378mm × 252mm), leaving small gaps at the edges. The gaps are fine — baseplates don't need to fill every millimeter.
+  - q: What bin sizes should I use for Gridfinity?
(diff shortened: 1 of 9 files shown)


andymai:
Pushed `dbfbe50dd` addressing the legitimate findings:

**Copilot inline #1 (`scripts/build-content.ts` L113)** — fixed. The FAQPage `acceptedAnswer.text` and the visible accordion HTML now derive from the same single render pass (`marked.parseInline(escapeHtml(faq.a))`). The PR description's "can't drift" claim is now actually true rather than aspirational.

**Copilot inline #2 (`public/content.css` L422)** — fixed. Added `summary::marker { content: '' }` so the default disclosure ▶ is hidden in Firefox alongside the existing `::-webkit-details-marker` rule for WebKit.

**Greptile inline #2 (raw HTML through `marked.parseInline`)** — fixed by the same change as Copilot #1. The source is now escapeHtml'd before parsing, so a stray `<tag>` in a future translation renders as text. Markdown syntax is unaffected because `<>&"'` aren't markdown characters.

**Greptile inline #1 (`privacy` and `terms` missing from `handCraftedPages`)** — this is a false positive. `handCraftedPages` is for hand-crafted HTML files in `public/` (the three generator landing pages), which need their CSS hash patched in place. `privacy` and `terms` are markdown-sourced from `content/privacy.md` and `content/terms.md` — the main loop already regenerates them on every build via `processFile`, which writes a fresh `<link rel="stylesheet" href="/${cssFilename}">` into the output. Adding them to `handCraftedPages` would double-process them (regenerate, then immediately patch the same hash back). The `M public/{privacy,terms}/index.html` diffs in this PR are exactly that auto-regeneration in action, not manual patches.

PR #1712, opened 2026-05-18
feat(seo): add de/fr/es/pt-BR locales for /what-is-gridfinity and /guide

andymai:
Stacked on **#1711**. Targets `seo/faq-breadcrumb-schema` — merge that one first, then rebase this onto `main`.

## Summary

Captures the 40% non-USA traffic that the English-only site currently leaves on the table. Two highest-value SEO landing pages (`/what-is-gridfinity`, `/guide`) now have crawlable, server-rendered HTML in German, French, Spanish, and Brazilian Portuguese at locale-prefixed URLs.

8 new translated markdown files under `content/{de,fr,es,pt-BR}/` — translations anchored to the in-app i18n locale glossary so that "Bin" stays as the Gridfinity-specific loanword across all locales, baseplate becomes Grundplatte / plaque de base / placa base / placa-base, etc.

`scripts/build-content.ts` now scans both `content/*.md` (English) and `content/{locale}/*.md` (translations), emitting:
- English at `public/{slug}/index.html` (unchanged behavior)
- Other locales at `public/{locale}/{slug}/index.html`
- Each page declares its locale via `<html lang>`, `og:locale`, and a full bidirectional hreflang cluster (en + 4 locales + x-default)
- Footer, nav CTA, and "Frequently Asked Questions" heading localized via per-locale label maps

`sitemap.xml` expanded from 7 to 15 URLs, each carrying its own `xhtml:link` hreflang alternates per Google's sitemap-hreflang recommendation.

`vercel.json` now sets `Content-Language` headers on the `/de/`, `/fr/`, `/es/`, `/pt-BR/` subtrees, the same `public, max-age=3600, stale-while-revalidate=86400` cache headers as the existing English content paths, and explicit rewrites for the 8 new locale content URLs.

## Intentional non-changes

**Root `index.html` hreflang is unchanged.** Its body is English-only client-rendered React — declaring `hreflang="de"` from `/` to `/` would tell Googlebot the German variant is in English, which suppresses non-English ranking. The pre-existing comment in the head explains this. The locale ranking targets are the real, crawlable, server-rendered locale variants of `/what-is-gridfinity` and `/guide`, which fully self-declare hreflang.

**Privacy and terms stay English-only.** They're legal-compliance pages, not SEO ranking targets, so the limited translation budget went to product-content pages instead.

**Hand-crafted pages (`/gridfinity-{generator,bin-generator,baseplate-generator,sizes}`) stay English-only.** Migrating them to markdown so they flow through the same i18n pipeline is the follow-up PR.

## Test plan

- [x] `pnpm exec tsx scripts/build-content.ts` emits 12 pages across 4 slugs (4 EN + 4 DE + 4 FR + 4 ES + 4 PT-BR... wait, 8 markdown × locales = expected count)
- [x] Every generated JSON-LD block parses as valid JSON
- [x] `<html lang>` matches the locale on every page
- [x] Hreflang cluster on each page lists all 5 locales bidirectionally + x-default
- [x] `sitemap.xml` is valid XML, every locale URL has 6 `xhtml:link` alternates
- [x] `vercel.json` is valid JSON
- [x] Typecheck passes against `tsconfig.scripts.json`
- [x] All pre-commit hooks (boundaries, i18n × 3, exhaustiveness, component structure, missing tests, readme reminders) pass
- [ ] Google Search Console picks up locale variants after sitemap submission post-deploy
- [ ] Manual smoke test of `/de/guide` and `/fr/what-is-gridfinity` in production

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,9 +54,13 @@ scripts/train-bin-recommender/.env*
 scripts/train-bin-recommender/__pycache__/
 scripts/train-bin-recommender/.venv/
 
-# Generated content pages (built from content/*.md)
+# Generated content pages (built from content/*.md and content/{locale}/*.md)
 public/what-is-gridfinity/
 public/guide/
+public/de/
+public/fr/
+public/es/
+public/pt-BR/
 public/content.*.css
 
 # Generated English locale JSON (source of truth is en.ts)
--- a/content/de/guide.md
+++ b/content/de/guide.md
@@ -1,0 +1,144 @@
+---
+title: Eine Gridfinity-Schublade planen — Anleitung
+description: Praktischer Leitfaden zum Planen von Gridfinity-Schubladenlayouts. Schublade ausmessen, passende Bins bestimmen und eine Druckliste exportieren.
+keywords: gridfinity planer, gridfinity layout, gridfinity planen, schubladeneinsatz planen, gridfinity anleitung
+schema: HowTo
+breadcrumbs:
+  - name: Start
+    url: https://gridfinitylayouttool.com/de/
+  - name: Planungsleitfaden
+    url: https://gridfinitylayouttool.com/de/guide
+faqs:
+  - q: Wie messe ich eine Schublade für Gridfinity aus?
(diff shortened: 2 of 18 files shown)


andymai:
Pushed `4d66fe583` addressing the Greptile P2:

- **Non-null assertion in `localesBySlug.get(...)!`** — replaced with the `??` fallback pattern per CLAUDE.md's no-`!` rule. Two call sites fixed (the population loop and the per-entry processing loop).
- Also forward-applied the same FAQ render-once fix from PR #1711 here (schema text now matches visible HTML, raw HTML escaped before parsing, cross-browser `summary::marker` rule). This means when this PR rebases onto main after #1711 lands, the `renderFaqs` function shape converges with no manual conflict resolution.

PR #1787, opened 2026-05-20
fix(settings): tighten physical unit bounds and add reset

andymai:
## Summary

- Tighten grid unit bounds **1-200mm → 20-60mm** and height unit **1-50mm → 3-20mm**. Covers Gridfinity Standard (42/7), half-pitch (21/7), and mini-Gridfinity (25/7); rejects obvious nonsense like a 300mm grid unit.
- Add "Standard Gridfinity: Nmm" helper text below each input in Sidebar, MobileSettingsPanel, and the Settings → Defaults tab.
- Add "Reset to Gridfinity Standard" button (disabled when already at 42/7) in Sidebar and MobileSettingsPanel.

Driven by a real user typing 300/500 into these fields — bounds + helper + reset together solve both *"what does this mean?"* and *"how do I undo this?"*.

## Behaviour notes

- Clamping happens at commit time via `DeferredNumberInput`'s built-in `min`/`max`.
- Reset is atomic via the existing CQRS `batch()` (undoable like any other unit change).
- **Existing layouts with out-of-bound values are not migrated** — clamping only fires when the user next edits the input. Avoids silent mutation of persisted data.

## Test plan

- [ ] Type `300` into grid unit → clamps to `60` on blur
- [ ] Type `500` into height unit → clamps to `20` on blur
- [ ] Reset button restores both fields to `42` / `7`
- [ ] Reset button is disabled when already at `42` / `7`
- [ ] Helper text renders correctly in all 9 locales
- [ ] Loading a legacy layout with `gridUnitMm: 300` still displays `300` (no auto-clamp on load)

### DIFF
--- a/scripts/i18n-values-allowlist.json
+++ b/scripts/i18n-values-allowlist.json
@@ -1,6 +1,7 @@
 {
   "maxShortValueLength": 3,
   "keys": [
+    "settings.gridfinityStandardMm",
     "auth.providerGithub",
     "auth.providerGoogle",
     "dock.syncStatusOffline",
--- a/src/core/constants.ts
+++ b/src/core/constants.ts
@@ -28,6 +28,15 @@ export const CONSTRAINTS = {
   MIN_BIN_HEIGHT: 2, // Minimum bin height in units (1U = base only, no usable cavity)
   MIN_LAYER_HEIGHT: 2, // Minimum layer height in units (1U = socket base only)
   PRINT_GAP_MM: 10, // Gap between bins on print bed
+  // Physical unit bounds. Standard Gridfinity is 42mm grid / 7mm height; the
+  // range here covers half-pitch (21mm) and mini-Gridfinity (25mm) variants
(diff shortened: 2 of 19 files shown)


andymai:
Both of Copilot's findings (re: focus+blur silently clamping out-of-range legacy values) were legitimate — the new bounds turned every blur into a write, which directly contradicted the "leave existing layouts alone" policy spelled out in the PR description.

Pushed `e35749530`: `DeferredNumberInput.commit()` now early-returns when the local text equals the formatted prop, so focus+blur on a legacy value no-ops. Added a regression test covering the focus+blur path. Persisted out-of-range data now stays put until the user actually edits the field.

andymai:
Greptile's P2 on `DefaultsTab.tsx:203` addressed in `3b85a0950` — dropped the redundant `Math.max/Math.min` wrapper since `DeferredNumberInput` already clamps via its `min`/`max` props before invoking `onChange`. Matches the pattern in `Sidebar.tsx`.

PR #1807, opened 2026-05-21
feat(bin-designer): foundation for engraved text on label tabs and cutouts

andymai:
## Summary

Foundation PR for the embedded-text feature brainstormed in [chat]. Adds the data layer, store actions, persistence wiring, server validation, and the labs flag for engraved text on label tabs and cutouts. **Geometry pipeline and editor UI ship in follow-up PRs** — by itself this PR doesn't render anything new; it makes the next two PRs much smaller.

## Why split

The full feature (data layer + worker font loading + textBuilder + label-tab/cutout integration + inspector UI + analytics + ~15 i18n keys) is ~15+ files and several hours of careful work. Landing the foundation first means:
- Data schema is reviewed in isolation
- The compartment-ID-remap fix (a real correctness gap) gets dedicated scrutiny
- Geometry & UI PRs can focus on their own concerns

## What's in
- `BinParams.textDefaults` (font / mode / depth / margin / min-max font size). Default: Atkinson, engrave, 0.4mm.
- `CompartmentConfig.compartmentTexts?: string[]` parallel-to-cells. Replaces a naive `Record<id, string>` that would have **corrupted text→compartment associations on any grid mutation** because `normalizeIds()` renumbers IDs every merge/split.
- `Cutout.engraveLabel?` (off by default — no visual change for existing designs), `textSide?`, `textStyle?`.
- `LabelTabConfig.textStyle?`.
- `FeatureColorConfig.text` + `ColorZone 'text'` + `FeatureTag.TEXT` (14). `computeActiveZones` surfaces it only when text actually exists.
- New helpers `normalizeIdsWithRemap()` and `remapCompartmentTexts()` keep parallel arrays in lockstep with `cells`.
- `paramSlice` actions: `setCompartmentText`, `setTextDefaults`, `setLabelTabTextStyle`; `mergeCells` / `splitCompartment` apply the remap.
- `migrateParams` fills missing `textDefaults`; new `featureColors.text` inherits the `labelTab` color so legacy designs see no shift.
- Server `ALLOWED_PARAM_KEYS` allowlists `textDefaults` so cloud shares don't silently drop the field.
- `embedded_text` experimental labs flag for the upcoming UI.
- i18n: `binDesigner.colors.text` + `binDesigner.colors.zone.text` mirrored to all 8 locales.

## What's NOT in (follow-ups)
- Worker font loading (Atkinson / JetBrains Mono / Allerta Stencil TTFs)
- `textBuilder.ts` (sketchText + extrude + boolean)
- `labelTabBuilder` integration (engraved text on shelf top)
- `cutoutBuilder` integration (text adjacent to cutout, world-aligned)
- Cache-key updates in `labelTabsFeature` and cutout builder
- Inspector text inputs + panel sections
- PostHog analytics events
- README + CLAUDE.md docs

## Test plan
- [x] `pnpm run lint` — 0 errors (5 pre-existing warnings)
- [x] `pnpm run knip` — clean
- [x] `pnpm run check:i18n` — all 8 locales aligned at 1868 keys
- [x] `pnpm vitest run src/features/bin-designer` — 2209 / 2209 pass
- [x] Added 4 new test cases for `normalizeIdsWithRemap`, `remapCompartmentTexts`, `mergeCells with texts`, `splitCompartment with texts`
- [x] Fixed `FeatureColorConfig` fixtures in 7 test files (added `text` zone + previously-implicit `enabled` field)
- [ ] Manual: load an existing design — `textDefaults` filled from migration, `featureColors.text` matches `labelTab`, no visual change
- [ ] Manual: undo/redo across merge/split with `compartmentTexts` populated (will be exercised by the UI PR)

## Notes
- `paramSlice.ts` still has no sibling test (pre-existing warning from the missing-tests hook). The new actions get coverage indirectly via the compartment scenario tests; a focused `paramSlice.test.ts` is queued for the UI PR where it can exercise the user-facing flows end-to-end.
- File `src/features/bin-designer/constants/defaults.ts` is 518 lines (warning threshold 500). Was already 510+ before this PR; moved `DEFAULT_TEXT_STYLE_DEFAULTS` to `types/text.ts` to minimize the bump.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -70,6 +70,7 @@ Bin → position (x,y), size (w,d,h), layerId, category, label, notes, customPro
 3. **Half-Bin Mode**: 0.5 increments. Helpers: `snapToHalf()`, `snapToGrid()`, `isFractional()`. `HALF_BIN_SCALE = 2`.
 4. **Multi-Layout**: Each layout stored by UUID (`gridfinity-layout-{uuid}`). Library index tracks metadata only.
 5. **Wall Pattern Border Rule**: Any feature that cuts through a wall (cutouts, handles, etc.) MUST have corresponding border clipping in `wallPatternBuilder.ts`. Cutout/handle clips use `CUTOUT_BORDER_WIDTH` (1.5mm); divider junction clips use `max(CUTOUT_BORDER_WIDTH, shapeRadius)` so larger hex prisms (4u+ bins) can't bleed into divider walls. Without border clipping, hex prisms overlap the cut region producing jagged edges.
+6. **Compartment IDs are not stable**: `normalizeIds()` renumbers `compartments.cells` on every merge/split, so any parallel per-compartment array (e.g. `compartmentTexts`) must be reindexed in lockstep via `normalizeIdsWithRemap()` + `remapCompartmentTexts()`. Resetting the grid (`setCompartmentGrid`) regenerates IDs from scratch — drop parallel arrays rather than carry ghost values onto unrelated cells.
 
 ### Result Type (`src/core/result/`)
 
--- a/api/lib/designerValidation.test.ts
+++ b/api/lib/designerValidation.test.ts
@@ -645,4 +645,58 @@ describe('validateDesignerShare', () => {
       }
     });
   });
+
+  describe('textDefaults validation', () => {
+    function withTextDefaults(td: unknown) {
+      const payload = validPayload() as ReturnType<typeof validPayload> & {
+        params: { textDefaults?: unknown };
+      };
+      payload.params.textDefaults = td;
+      return validateDesignerShare(payload, JSON.stringify(payload).length);
+    }
+
+    it('accepts the canonical defaults', () => {
+      const result = withTextDefaults({
+        font: 'atkinson',
+        mode: 'engrave',
+        depth: 0.4,
+        margin: 1.5,
+        minFontSize: 3,
+        maxFontSize: 20,
+      });
+      expect(result.valid).toBe(true);
+    });
+
+    it('rejects non-object', () => {
+      expect(withTextDefaults('oops').valid).toBe(false);
+    });
+
+    it('rejects unknown keys', () => {
+      const result = withTextDefaults({ font: 'atkinson', evil: 1 });
+      expect(result.valid).toBe(false);
+      if (!result.valid) expect(result.error.message).toMatch(/unknown key/);
+    });
+
+    it('rejects unsupported fonts', () => {
+      const result = withTextDefaults({ font: 'comic-sans' });
+      expect(result.valid).toBe(false);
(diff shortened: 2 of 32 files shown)


andymai:
Addressed all 4 issues from @greptile-apps and the earlier 3 from Copilot:

**Critical (Greptile #1, #2):** `text` added to `ALLOWED_FEATURE_COLOR_KEYS` **and** the hex-color validation loop. The bug would have rejected every cloud-share for every user once the migration ran — good catch. Regression test added (`accepts the text zone hex` + `rejects a non-hex text zone color`).

**Security (Greptile #3):** server `validateCompartments` now validates `compartmentTexts`: must be an array, entries must be strings ≤ 50 chars, length ≤ cols × rows. Mirrors the client's `TEXT_MAX_LENGTH` so a direct HTTP POST can't bypass it. 5 new test cases.

**Polish (Greptile #4):** `setCompartmentText` no-op short-circuit — comparison happens before `pushHistoryEntry`, so per-keystroke calls from the upcoming UI inspector won't spawn one undo step per character. Same-value writes and empty→empty re-clears are both covered by tests.

**Copilot:** `TEXT_MAX_LENGTH` constant used instead of hard-coded `50`; `trimmed` → `clamped` since it's truncation not whitespace strip; `setLabelTabTextStyle` signature narrowed to `TextStyleOverride | null` so `undefined` can't leak into the set branch; `remapCompartmentTexts` doc + test updated to reflect the actual 1:1 contract.

Quality gate: `pnpm run typecheck` clean, `pnpm run lint` 0 errors, `pnpm run check:i18n` 1868/1868 across all 8 locales, 210 affected tests pass.

andymai:
Addressed both items from the latest @greptile-apps review:

**Issue 1 — `validateTextDefaults`:** Added structure validator at `api/lib/designerValidation.ts`. Enum checks on `font` and `mode`, numeric ranges on `depth` (0-10mm), `margin` (0-50mm), `minFontSize` (0.5-100mm), `maxFontSize` (0.5-200mm). Caps mirror the safe ranges the geometry PR will rely on, so a crafted share can't smuggle in `depth: -1` or `maxFontSize: 1e9`. 7 new test cases covering the attack vectors.

**Issue 2 — `compartmentTexts: readonly string[]`:** Pushed back here. The suggested change breaks the build — Immer's `Draft` wrapper requires mutable element types, so the existing pattern in this file is field-level `readonly` with mutable inner arrays (`readonly cells: number[]`, `readonly inserts: Insert[]`, `readonly cutouts: Cutout[]`). Switching `compartmentTexts` to `readonly string[]` triggers:
> Type 'readonly string[]' is not assignable to type 'string[]'.
> The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.

The right consistency is with the sibling fields on the same interface, not with `ActiveZonesParams` (which is a read-only computed type — different context). Added an inline comment documenting the convention so future readers don't repeat the experiment.

PR #1862, opened 2026-05-23
feat(baseplate): add 9-point padding anchor selector

andymai:
## Summary

- Adds a 3×3 dot anchor selector inside the dashed center area of the baseplate padding schematic
- Picking a dot redistributes the current X (L+R) and Y (B+F) padding sums to bias bins toward that corner/edge/center
- Editing any individual padding stepper flips the anchor back to `'custom'`; new layouts default to `'custom'` (no behavior change)
- Anchor persists in `BaseplateParams.paddingAnchor`, survives save/load, undo/redo, and import; migration preserves it when present

## Design

- 3×3 radiogroup (back-left / back-center / back-right / middle-left / center / middle-right / front-left / front-center / front-right) with arrow-key navigation
- Distribution uses integer-hundredths math so the 0.01mm leftover from odd-sum splits lands on the end side (top-left rounding bias)
- Per-side values are clamped to `[PADDING_MIN, PADDING_MAX]`; clamping surfaces an unobtrusive warning badge on the selector

## Test plan

- [x] Unit tests for `computeAnchoredPaddings` (all 9 anchors, leftover assignment, clamping, total preservation)
- [x] Component tests for `PaddingAnchor` (click + arrow-key navigation, edge clamping, warning badge)
- [x] Integration tests in `PaddingSchematic` (anchor batch-update, edit-flips-to-custom, no-bounce when already custom)
- [x] Typecheck + lint clean
- [x] Playwright spot check: schematic renders inside the dashed area; click → padding values redistribute; edit a stepper → anchor returns to custom
- [x] All 8 non-English locales translated (de, es, nl, pt-BR, fr, nb, uk, sv)


andymai:
Addressed review feedback in three commits:

**P1 (correctness)** — `905c8993a`:
- `setBaseplateParams` v2 schema now includes `paddingAnchor`. Without this, Zod stripped the key and the anchor was silently lost on save/load/undo. Added a test that round-trips it through the handler.
- `showClampWarning` now comes from the redistribution result captured at click time, stored in local component state. The previous derivation re-ran `computeAnchoredPaddings` against the *already-clamped* paddings, which is a fixed point of the function — so `clamped` always returned `false` and the warning never showed. The new approach also implicitly resolves Greptile's `useMemo` concern.

**P2 (a11y)** — `8ab032448`:
- Roving tabindex: only the selected anchor (or `tl` when `value === 'custom'`) is in the tab order. Matches `ThicknessSelector` and the WAI-ARIA radiogroup pattern.
- Arrow keys that would land on the same anchor (grid edges) early-return without firing `onChange`, so holding an arrow against the boundary doesn't pollute the undo stack.

**P3 (test)** — `6831293ee`:
- `'ml' anchor` test now uses non-zero `paddingLeft`/`paddingRight` so the X-redistribution assertion actually exercises the behavior.

Skipped:
- Greptile's "add a README section" — project convention is no doc changes in feature PRs.

PR #1864, opened 2026-05-24
fix(generation): surface cutout boolean failures instead of dropping them

andymai:
## Summary
- The solid-bin cutout cut in `featuresStage` swallowed any `cut()` failure and returned the solid **without cutouts**. The bin generated, but the user's cutouts silently disappeared.
- Surface the underlying error instead, so the failure is visible and recoverable.
- Most visible on **split bins**: `splitBinBuilder` builds the body solid with `stackingLip: false` (lip is added separately per-piece to avoid a different OCCT crash). The lip-less shell has different top-edge topology than the lipped shell, so a cutout the boolean handles fine in the unsplit preview can throw on the lip-less split-path solid — and the user sees a bin with no cutouts on the split pieces.

## What changed
- `src/features/generation/worker/generators/pipeline/stages/featuresStage.ts` — replace silent `catch {}` with `throw new Error('Failed to apply cutouts to solid bin', { cause })`. WASM cleanup of the cutout solid is preserved.
- `src/features/generation/worker/generators/binGenerator.scenario.split-cutouts.test.ts` (new) — three scenario tests pinning the contract that cutouts survive the boolean-intersection split path:
  - Unsplit vs split parity (vertex delta) on a 6×2×3 solid+lip bin.
  - Top-down rectangular cutout straddling the split plane — present in both pieces.
  - Front wall cutout straddling the split plane — survives split.

## Caveat — what this PR does NOT do
This makes the silent failure visible. It does **not** root-cause the underlying OCCT boolean failure for the specific cutout config that triggers the throw — none of the realistic param sweeps I tried (different cutout sizes / positions / depths / connector configs / 6×2 and 8×2 bins) reproduced the throw. Once a user hits the new error message in the wild we'll have the params to write a focused fix.

## Test plan
- [ ] `pnpm exec vitest run src/features/generation/worker/generators/binGenerator.scenario.split-cutouts` passes.
- [ ] `pnpm exec vitest run src/features/generation/worker/generators/binGenerator.scenario.split-robustness src/features/generation/worker/generators/binGenerator.scenario.cutouts src/features/generation/worker/generators/binGenerator.scenario.diagonal-holes src/features/generation/worker/generators/featureBuilder` — 76 tests, no regression.
- [ ] `pnpm run typecheck` clean.
- [ ] Manual: re-trigger the original repro and confirm the cutout cavity now either renders correctly OR a user-visible error surfaces with `cause: <OCCT-error>` instead of producing a bin with missing cutouts.

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.split-cutouts.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.split-cutouts.test.ts
@@ -45,7 +45,7 @@ function makeRectCutout(overrides: Partial<Cutout> = {}): Cutout {
   };
 }
 
-function totalVerts(pieces: { vertices: Float32Array }[]): number {
+function totalVerts(pieces: readonly { vertices: Float32Array }[]): number {
   return pieces.reduce((s, p) => s + p.vertices.length, 0);
 }
 
--- a/src/features/generation/worker/generators/binGenerator.scenario.split-cutouts.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.split-cutouts.test.ts
@@ -6,7 +6,7 @@
  * dropped cutouts from the output.
  */
 import { describe, it, expect, beforeAll } from 'vitest';
-import { DEFAULT_BIN_PARAMS, DISABLED_WALL_CUTOUT } from '@/shared/constants/bin';
+import { DEFAULT_BIN_PARAMS } from '@/shared/constants/bin';
 import { DEFAULT_SPLIT_CONNECTOR_CONFIG } from '@/features/bin-designer/constants/defaults';
 import type { BinParams, Cutout, SplitConnectorConfig } from '@/shared/types/bin';
 import { initBrepjs, getGenerateBin, getGenerateSplitPreview } from './__kernel-tests__/wasmInit';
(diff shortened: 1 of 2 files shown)


andymai:
Addressed both reviewer threads — split into three commits by severity:

**Tier 1 — critical correctness** (`9ceb3c61b`)
- **WASM shell leak on throw** (Greptile #1, Copilot inline on `featuresStage.ts:54`): hoisted `oldSolid` out of the `try` and dispose it (plus `cutoutCuts`) in the `catch` before rethrowing. Defensive try around `oldSolid.delete()` so an already-disposed handle doesn't mask the original boolean failure.
- **`cause` not user-visible** (Copilot inline on `featuresStage.ts:53`): `formatError()` in the worker only serializes `Error.message`, so the wrapped `cause` was invisible. Appended the cause's message to the thrown Error so the OCCT detail surfaces to UI + logs without changing the worker error path.

**Tier 2 — test coverage gap** (`aca15ff00`)
- **Wall-cutout test didn't exercise the fix** (Greptile #2): correct — wall cutouts on a hollow bin take the non-solid branch in `featuresStage`, never reaching the `if (dim.solid)` block. Wall cutout + split is already covered by `binGenerator.scenario.split-robustness.test.ts:192`. Dropped the duplicate. The two remaining tests (unsplit/split parity, straddling cutout) both exercise the fixed solid-bin path.

**Tier 3 — nit** (`17234e260`)
- **`readonly` array type** (Copilot inline on `split-cutouts.test.ts:48`): `totalVerts` now accepts `readonly { vertices: Float32Array }[]` to match `SplitPreviewResult.pieces`.

Local: typecheck clean, 2 scenario tests pass.

PR #1866, opened 2026-05-24
fix(generation): route solid-bin cutouts through booleanStage's cutAllBisect

andymai:
## Summary
Follow-up to #1864, which surfaced the silent failure. This makes the underlying boolean robust so the failure stops happening in the first place.

## Problem
`featuresStage`'s solid-bin branch took every cutout, **fused them into one shape**, then ran a bare `cut(shell, fusedCutouts)`. Two ways this could drop everything:

1. `fuseAll` on the cutout set throws → the whole set is lost (silent before #1864, error after).
2. `cut(shell, fusedCutouts)` throws on one bad spot in the topology → the *entire boolean* aborts, so even cleanly-buildable cutouts in the same set disappear together.

The rest of the pipeline (`compartmentBuilder`, `wallCutoutBuilder`, etc.) goes through `booleanStage`'s `cutAllBisect`, which bisects on failure so a single degenerate tool doesn't poison the others. The solid-bin path was the outlier.

## Change
- `buildCutoutCuts` now returns **`Shape3D[]`** (one logical cutout per entry — ungrouped cutout, grouped-and-scooped union, or engraved label), each individually clipped to the bin interior.
- `featuresStage` pushes them into `ctx.cutTargets` and lets `booleanStage` apply them via `cutAllBisect`. Per-cutout `collectOrigins` calls keep face-tag provenance intact.
- The in-stage manual `cut()` is gone, along with the silent-catch fix from #1864 and its WASM-cleanup follow-ups — `booleanStage`'s existing recovery covers it.
- Export passes pick up `simplify: forExport` automatically (the bare `cut()` did not), which is what slicers like BambuStudio want for clean topology.

## Risk
The bare-cut path was kept "to preserve original behavior" per a 2025 refactor comment, but the rest of the pipeline has been on `cutAllBisect` + simplify since the brepjs 2026 bump (#1795). Routing the solid path through the same shape removes the divergence rather than adding a new path.

## Tests
- `binGenerator.scenario.split-cutouts.test.ts` — three scenario tests using a cavity-floor z-level probe instead of brittle vertex-count deltas:
  - cavity floor appears in unsplit and in split outputs
  - straddling cutout leaves a cavity floor in **both** split pieces
  - **one bad tool does not drop the rest** — pins the new robustness directly
- `featureBuilder.test.ts` — updated `buildCutoutCuts` assertions for the new `Shape3D[]` return type.

## Test plan
- [x] `pnpm exec vitest run src/features/generation/worker/generators/` — 967 pass, 4 pre-existing skips, 0 fail.
- [x] `pnpm run typecheck` clean.
- [x] Lint clean on changed files (one pre-existing `max-lines` warning on `cutoutBuilder.ts` unchanged from main).
- [ ] Manual: re-run the original repro from #1864 (rectangular cavity straddling the split line) and confirm the cavity now renders in both split pieces.

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.split-cutouts.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.split-cutouts.test.ts
@@ -129,10 +129,14 @@ describe('split + top-down cutouts', () => {
     }
   }, 120000);
 
-  it('one bad tool does not drop the rest', () => {
-    // cutAllBisect's bisect recovery: even if one tool is degenerate, the
-    // remaining cutouts should still survive. A zero-size cutout is the
-    // simplest pathological tool we can construct.
+  it('a degenerate cutout in the set is skipped without dropping the others', () => {
+    // Zero-size cutouts are filtered before they reach the boolean (the
+    // shape builder returns null for non-positive dimensions), so this
+    // pins the filter behaviour rather than cutAllBisect's bisect path.
+    // A real bisect-recovery regression test needs a tool that builds
+    // successfully but fails inside cut() — none of our 20+ permutations
+    // surface one consistently, so we cover the architecture via this
+    // weaker but realistic property.
     const generateBin = getGenerateBin();
     const bad: Cutout = makeRectCutout({ id: 'bad', x: 100, width: 0, depth: 0 });
     const good = makeRectCutout({ id: 'good', x: 20, y: 20, width: 30, depth: 30, cutDepth: 8 });
@@ -145,7 +149,7 @@ describe('split + top-down cutouts', () => {
     );
     expect(
       [...mixed].filter((z) => !base.has(z)),
-      'the good cutout should still carve a cavity even when sharing the set with a degenerate tool'
+      'the good cutout should still carve a cavity alongside a filtered-out degenerate tool'
     ).not.toHaveLength(0);
   }, 120000);
 });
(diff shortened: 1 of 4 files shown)


andymai:
Addressed all three reviewer threads, split by severity:

**Tier 1 — correctness** (`e55dc43ba`)
- **`cutTargets` leak when `ctx.solid` is null** (Copilot inline on `featuresStage.ts:46`): `booleanStage` early-returns when `ctx.solid` is null, so unconditionally populating `cutTargets` would leak the tools we'd never apply. Added an explicit `if (!ctx.solid) return ctx;` guard before building any tools.

**Tier 2 — test honesty** (`5550c3223`)
- **Misnamed bisect-recovery test** (Copilot inline on `split-cutouts.test.ts:135`): the zero-size cutout I used is filtered upstream by the shape builder (returns null for non-positive dimensions), so it never reached `cutAllBisect`. Renamed the test to "a degenerate cutout in the set is skipped without dropping the others" and added a comment explaining that I couldn't construct a tool that *builds successfully but throws inside `cut()`* in 20+ permutations, so the architecture is covered by this weaker but realistic property.

**Tier 3 — defensive memory** (`7204fffb5`)
- **`rawShapes` leak if `box()` throws** (Greptile #1): pulled the `box()` call into the outer `try` and added a `catch` that disposes both `rawShapes` and any tools already added to `clipped`. Defensive against WASM heap allocation failures.

Local: typecheck clean, scenario tests still pass.

PR #1901, opened 2026-05-27
feat(label-tabs): add adjustable shelf height

### LINKED ISSUE #1898: Add support for a "tuck-under ledge"
**Problem**

I want to store unruly jumper wires in narrow bins:

Image

The wires (especially Dupont cables) rarely lay flat. Some tape stretched across the top can help keep them tucked into the bin, but that's ugly.

**Proposed Solution**

I want to be able to add "shelves" near the ends of the bin, just under the stacking rim (so I can still stack bins on top), that I can tuck the ends of the wires under. They should span the entire width of the bin, and I want to be able to specify how wide they are and how far from the end they start. They will require supports for printing, unfortunately.

Does this make sense? If not, I can try to CAD up what I'm envisioning.

**Alternatives Considered**

Ugly tape lol. 

Actually, I might be able to print up a little bridge piece to insert into the bin after the fact. But not as nice as a ledge that's part of the structure.

Another alternative might be some slots I could slide a ledge into across the bin opening.

andymai:
## Summary

- Adds an optional **`height`** field to `LabelTabConfig` — the Z position of the shelf top above the cavity floor (mm). Absent = anchor at wall top (legacy behavior, byte-identical output).
- Lowering `height` drops the entire tab assembly (shelf + gusset) down the back wall, leaving a **tuck-under pocket** between the rim and the shelf. Solves #1898 (springy jumper-wire bins where the wires won't stay tucked).
- UI: third stepper joins Width / Depth in the parametric row, labeled "Tab height". Dynamic min = `tabDepth + 1`, dynamic max = interior wall height. `setTabDepth` clamps height upward when needed so the geometry layer doesn't silently render nothing.
- API validation, all 9 locales, and the scenario suite are updated in lockstep.

Closes #1898.

## How the change is contained

- **Type** (`LabelTabConfig.height?: number`) — optional, so existing serialized designs round-trip identically.
- **Builder** — single anchor swap: `shelfTopZ - tabHeight` replaces `wallHeight - tabHeight`. Gussets keep their original size; the assembly slides as a unit.
- **Cache key** — unchanged (`v3`). The existing `stableSerialize(params.label)` already differentiates absent vs. present `height`.
- **API** — new `MIN/MAX_LABEL_TAB_HEIGHT` constants mirror the client; validator rejects out-of-range values when the field is present.

## Test plan

- [x] `pnpm run lint` — 0 errors (only pre-existing warnings)
- [x] `pnpm run check:i18n` — all 8 JSON locales match `en.ts`
- [x] Full vitest suite (855 files / 11901 tests) — passing
- [x] New scenario `2×2×6 label dropped (height = 20mm)` — distinct snapshot from existing label scenarios
- [x] New builder tests — Z position, default-at-top, returns null for invalid configs
- [x] New hook tests — bounds, depth-invalidates-height clamp, no migration of legacy designs
- [ ] Manual: open the bin designer, enable label tabs, drop the height stepper, verify a pocket appears between rim and shelf in the 3D preview

### DIFF
--- a/api/lib/designerValidation.ts
+++ b/api/lib/designerValidation.ts
@@ -52,6 +52,12 @@ const CONSTRAINTS = {
   MAX_LABEL_TAB_DEPTH: 20,
   MIN_LABEL_TAB_WIDTH: 10, // %
   MAX_LABEL_TAB_WIDTH: 100, // %
+  // Label tab height is the Z of the shelf top above the cavity floor (mm).
+  // Static bounds for API payloads — the client's UI uses a dynamic max
+  // tied to the current bin's interior height. Floor (9) = MIN_LABEL_TAB_DEPTH + 1;
+  // ceiling (140) = MAX_HEIGHT * 7mm (heightUnitMm).
+  MIN_LABEL_TAB_HEIGHT: 9,
+  MAX_LABEL_TAB_HEIGHT: 140,
   MAGNET_MIN_DEPTH: 2.0,
   MAGNET_MAX_DEPTH: 4.0,
   MAX_INSERTS: 20,
@@ -496,6 +502,14 @@ function validateLabel(label: unknown): string | null {
       !['left', 'center', 'right'].includes(label.alignment as string)
     ) {
       return 'label.alignment must be "left", "center", or "right"';
+    }
+    // Optional field; absent = anchor shelf at the wall top (legacy behavior).
+    if (
+      label.height !== undefined &&
+      (!isNumber(label.height) ||
+        !inRange(label.height, CONSTRAINTS.MIN_LABEL_TAB_HEIGHT, CONSTRAINTS.MAX_LABEL_TAB_HEIGHT))
+    ) {
+      return `label.height must be ${CONSTRAINTS.MIN_LABEL_TAB_HEIGHT}-${CONSTRAINTS.MAX_LABEL_TAB_HEIGHT}`;
     }
   }
   return null;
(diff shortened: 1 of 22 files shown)


andymai:
Pushed `833ea5de` addressing the review feedback:

**Copilot #1 — `tabHeightMax` could exceed `wallHeight`:** Fixed. Max is now strictly `wallHeight`. When the depth-derived floor (`depth + 1`) exceeds the ceiling, min collapses onto max — the stepper is fixed at the wall top, which is a degenerate but valid state the geometry layer accepts. Added a regression test (`tabHeightMax never exceeds wallHeight even when depth + 1 would`).

**Copilot #2 — API cross-field validation:** Added `label.height <= label.depth` guard alongside the range check. Payloads like `{ depth: 20, height: 9 }` are now rejected with a descriptive message instead of silently dropping the tab.

**CI quality failure:** The `ja` (Japanese) locale was added on `main` after this branch started. Rebased onto `7e100602` and added the matching `binDesigner.tabHeight` key.

andymai:
Pushed `b467e38a` addressing both Greptile findings:

**Greptile #1 — `setTabDepth` height-clamp overflow:** Fixed. The clamp now wraps with `Math.min(depth + 1, wallHeightMm)` so it can never write a height above the actual wall ceiling. Locked down with a regression test (`setTabDepth caps height clamp at wallHeightMm`).

**Greptile #2 — `GhostLabelTabs` ignored `label.height`:** Fixed. The ghost mesh now positions itself at `floorZ + (label.height ?? wallHeight) + 0.2` instead of pinning to the rim, so the new height stepper produces immediate visual feedback while the BREP regenerates. Mirrors the same anchor the builder uses.

PR #1904, opened 2026-05-27
feat(label-tabs): tuck-under ledges for wire bins

andymai:
## Summary

Closes the wire-tucking gap reported in #1898. Previously the label-tab feature could only anchor a single tab to the back wall — which doesn't help when you want to tuck Dupont/jumper cables under a ledge at the *front* of the bin, or both ends. This PR adds:

- **`edges` picker (back / front / both)** in the Label Tabs panel. 'both' renders a tab at each end of the compartment — the wire-tucking shape the issue asks for.
- **`inset` mm offset** slides the tab inward from its anchor wall. For 'both', both tabs move inward symmetrically so they work for wires shorter than the full bin.
- **Depth cap raised 20 → 50mm** so the ledge has enough material to retain springy contents.

UX cleanup along the way:
- The alignment control is **hidden when width = 100%** (the default), where it had no visible effect. JetForMe explicitly called this out as confusing in their issue reply.
- Tooltip on the alignment label clarifies its purpose when shown.
- Inline warning surfaces when `edges='both'` on a compartment too shallow for both tabs (front tab silently drops, warning tells the user why).

## Implementation notes

- Per-row tab construction was extracted into `buildTabsAtRow(scope, params, row, anchor, dims, collidingFrontIds)` in a separate prep commit (byte-identical output — all scenario snapshots match). The feature commit then parameterizes it with a `depthSign` factor threaded through shelf draw, gusset profile, fillet profile, text-center, and world translation.
- Defense in depth: UI clamps depth ≤ innerD-1, geometry rejects with silent null on violations, API validates enum + range (no cross-field — matches the existing silent-fallback pattern for `tabDepth >= innerD`).
- New `compartmentHasTiltedFrontWall` helper mirrors the existing tilted-back-wall skip, so front tabs respect angled-divider boundaries.
- Cache key bumped `v3 → v4` belt-and-suspenders.

## What changed

| Layer | Files |
| --- | --- |
| Schema | `types/index.ts`, `constants/defaults.ts`, `constants/gridfinity.ts`, `shared/types/bin.ts` |
| Geometry | `labelTabBuilder.ts`, `filletProfile.ts`, `utils/compartments.ts` |
| UI | `LabelTabsSection.tsx`, `useLabelTabsSection.ts`, `GhostLabelTabs.tsx` |
| Estimates | `printEstimates.ts` |
| API | `api/lib/designerValidation.ts` |
| i18n | en.ts + en/de/es/fr/ja/nb/nl/pt-BR/sv/uk JSONs (7 keys × 10 locales) |
| Tests | Scenario tests for front/both/inset/collision; unit tests for the new helper, hook bounds, validation |
| Docs | `bin-designer/README.md` |

## Follow-up

`printEstimates.computeLabelTabVolume` uses `cols × (…)` rather than walking back-edge groups — already wrong for merged compartments, now also under-counts for `edges='both'` until that's fixed. Pre-existing, not regressed; tracking issue will follow.

## Test plan

- [x] `pnpm run quality` (typecheck + lint + knip) clean
- [x] All 11924 tests pass; 4 new scenario snapshots written
- [x] i18n key/value/interpolation/unused checks pass for all 10 locales
- [ ] Manual: open Bin Designer, enable Label Tabs, switch to 'both', verify two ledges render at front + back. Try raising depth to 30mm + inset 5mm on a 1×1 bin — front tab should silently drop and warning should appear.
- [ ] Manual: width 100% → alignment control hidden. Width 80% → control reappears with tooltip on hover.
- [ ] Manual: export 3MF of a 'both' design and load in OrcaSlicer — confirm both ledges print cleanly.

Refs #1898.


andymai:
Addressed in f1c381de2:

1. **Ghost overlay collision filter + tilted-wall skip** — `GhostLabelTabs.tsx` now mirrors `buildTabsAtRow`'s suppression logic: skips compartments with tilted anchor walls, drops the front quad when `edges='both'` and the per-compartment depth check would trigger, and applies the same `tabDepth + inset > compartmentDepth` per-compartment guard. The 3D preview now matches what the worker produces. Also dropped the misleading `Math.min(label.depth, cellD)` clamp in favor of the new guards.

2. **`tabInsetMax` now uses `cellD`** — `useLabelTabsSection.ts:184` switched from global `innerD` to `innerD / rows`. On a 4-row bin the prior ceiling exposed values that immediately fired the warning. Conservative for merged compartments but matches what the collision guard actually enforces. Added a hook test asserting the multi-row case.

3. **Single-edge `inset + depth > compartmentDepth` guard** (Copilot's catch on `labelTabBuilder.ts:125`) — added a per-compartment guard in `buildTabsAtRow` that silently drops the tab when the body + inset would exceed compartment depth. The UI clamps prevent reaching this via the stepper, but cloud-share payloads can carry the combo. Mirrors the existing `tabDepth >= innerD` bridge guard.

Scenario tweak: the collision-drop scenario now uses `depth=20` instead of 18 so that, in a 1u-deep bin (innerD ≈ 39mm), the 2·20 = 40mm condition actually triggers collision — the prior `depth=18` was below threshold and the test was passively passing as a both-tabs case.

All 11925 unit tests pass; quality gate clean.

PR #1922, opened 2026-05-28
feat(split-bin): wall connectors to keep tall split pieces aligned

### LINKED ISSUE #1869: Locking Mechanism for side walls of split bins
**Problem**
The side walls of split bins aren't very sturdy, especially on taller bins.

**Proposed Solution**
It would be helpful if they had optional alignment connectors, like the ones on the bottom.
The walls might need to be reinforced slightly to accommodate a locking mechanism.

Image

andymai:
Closes #1869.

## What this does

Split bins can now add **wall connectors** to their exterior side walls, so tall pieces stay aligned at the seam instead of splaying once assembled. The connector is a slim press-together key — a tongue on one half, a matching groove on the other — on each perimeter wall the cut crosses.

Whenever a bin is large enough to split, two **independent** toggles appear under the split options:

- **Alignment connectors** — the 45° floor scarf lap that registers the base.
- **Wall connectors** — the new side-wall keys.

Either can be enabled without the other. (Previously the wall option was nested under alignment connectors and couldn't be used on its own.)

## How it works

- **Self-supporting print.** The key's cross-section is constant up the build (Z) axis and the protruding tongue has a 45° chamfered underside, so it prints without supports. The halves press together horizontally — no undercut that would force a vertical drop-in past the groove and the stacking lip.
- **Thicker walls reuse their own material.** The key is anchored a fixed skin behind the _outer_ face rather than behind the full wall thickness, so a thicker wall envelops it. The reinforcing pilaster is added **only when the wall is too thin to host the key** — once the wall is thick enough it adds nothing inward and the cavity is preserved. The inward intrusion shrinks to zero by ~3.5 mm walls; thin/default walls behave essentially as before.
- When present, the pilaster thickens the wall **inward only**, preserving the Gridfinity external footprint so bins still tile.
- The key spans ~80% of the interior wall height and **stops clear of the rim**, so the stacking lip is never disturbed.

## Verification

- Scenario tests (real OCCT worker): valid full-height pieces, connector geometry actually added (guards the silent boolean-drop path), **stacking-lip Z unchanged**, inert when disabled, **wall connectors work with the floor scarf off**, and the **thick-wall skip-pilaster path** fuses/cuts cleanly.
- `wallKeyGeometry` unit tests: fixed outer skin, inset no longer grows with wall thickness, inward intrusion shrinks to zero as the wall thickens.
- UI + hook tests for the two independent toggles.
- One-shot Playwright visual check: pixel-diff confirms the toggle changes the rendered split preview (store → worker → canvas).
- Full suite green (11977 passed, 3 skipped); typecheck, lint, and i18n parity across all 9 locales pass.

## Notes

- Wall connectors are **opt-in** (default off).
- The inline print-testing helper text has been removed.
- **Follow-up:** if a decorative wall pattern or handle lands exactly where a connector sits, the pattern isn't yet clipped around the pilaster (per the `CUTOUT_BORDER_WIDTH` border rule). Not addressed here.


andymai:
Addressed the two items from the Greptile review (commit 37be8064d):

1. **Y-axis coverage** — added a depth-direction split scenario block (`width:2, depth:8`, `CUT_Y=[0]`) exercising the `makePlanPoint`/`bossDims` axis swap: valid full-height pieces, connector geometry added, and lip-Z unchanged. All 7 wall-locking scenario tests pass — the axis-swap path is correct, no latent bug.
2. **README** — documented the joint in `bin-designer/README.md` (new gotcha #15: two joints, perimeter-only, inward-only boss, default-off pending print tests) and updated the `splitConnectorBuilder.ts` line in `generation/README.md`.

Copilot review came back with no comments.

PR #1923, opened 2026-05-29
fix(baseplate): restore top-down fit-in-view as the default camera

andymai:
## What

The default baseplate preview opened **zoomed-in at a 3/4 angle** instead of the top-down, fit-in-view framing it used to show. This restores the expected default: the whole baseplate framed top-down on open.

| Before | After |
| --- | --- |
| Zoomed-in 3/4 perspective, grid receding into the distance | Top-down, baseplate fitted in view |

## Why

The camera-rig refactor (#1870 / #1877) replaced the single Canvas camera with a dual perspective/orthographic rig that swaps the real camera in via drei's `makeDefault` — a **layout effect** that only takes hold on the next render.

`CameraController` frames the camera in a **passive effect**, whose closure still held R3F's throwaway initial camera from the first render. It framed that throwaway, latched `initializedRef`, and the real camera arrived too late to be framed — left stuck at its seed position `[100, -100, 80]`.

## Fix

Read the **live** default camera from the R3F store (`get().camera`) inside the init effect instead of the stale closure value. By the time the passive effect runs, drei has already set the real camera as default. Framing still runs once, so projection toggles stay handled by `CameraRig`.

## Testing

- Added a regression test that asserts the live default camera is framed and the stale closure camera is not. Verified it fails on the pre-fix code and passes with the fix.
- `pnpm run typecheck`, ESLint, and all BaseplatePreview tests pass.
- Manual visual check (Playwright): default view now opens top-down and fitted.

### DIFF
--- a/src/features/baseplate/components/BaseplatePreview/CameraController.test.tsx
+++ b/src/features/baseplate/components/BaseplatePreview/CameraController.test.tsx
@@ -1,31 +1,101 @@
-import { describe, it, expect, vi } from 'vitest';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render } from '@testing-library/react';
+import { createRef } from 'react';
+import type { RefObject } from 'react';
+import type { OrbitControls as OrbitControlsType } from 'three-stdlib';
 
 vi.mock('three', () => ({
-  Vector3: vi.fn().mockImplementation(() => ({
-    normalize: vi.fn().mockReturnThis(),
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the Greptile finding: the `setCameraFar` effect now reads `get().camera` (the live default) instead of the stale closure camera, matching the framing init. `camera` stays in the deps so a projection swap re-runs it. Extended the regression test to assert the far-plane effect targets the live camera (fails on the pre-fix code).

PR #1928, opened 2026-05-29
fix(bin-designer): fit default 3D preview to screen at any bin size

andymai:
## What

The bin designer's default 3D preview now opens **framed to fit the viewport** in isometric view — centered, with the dimension and name labels fully visible — and it scales correctly with bin size. Previously the default view was zoomed in tight: the bin nearly touched the toolbar and the labels were clipped at the edges, regardless of how large or small the bin was.

## Why it was broken

The size-adaptive isometric framing (`calculateIdealDistance` + the `FRAME_FILL` margin) already existed, but it never reached the visible camera because of a camera-identity race:

- React Three Fiber mounts a transient default camera, then drei's `makeDefault` swaps in the real one.
- `CameraController`'s framing was a boolean one-shot (`initializedRef`) — it fired against the **throwaway** camera, so the real camera kept its fixed birth position `[100, -100, 80]`.
- That fixed position is a constant distance that ignores bin dimensions, so the view was never actually "fit to screen."

This is why tuning the `FRAME_FILL` constant had zero visible effect — the framing math wasn't touching the rendered camera at all.

## The fix

Track the **camera identity** instead of a one-shot flag. Frame whenever the active camera changes, with one exception: a projection (perspective ↔ orthographic) swap, where `CameraRig` already copies the pose across — re-framing there would discard the user's current orbit/zoom.

The default view now matches what "Reset view" (`R`) produces and adapts to bin size.

## Verification

Confirmed in a real browser (Playwright screenshots at 1440×900):

- ✅ Default view is byte-identical to the "Reset view" framing — centered, labels visible with margin
- ✅ Size-adaptive: a 7×7 bin zooms out to fit; a small bin fills appropriately
- ✅ Projection toggle to orthographic preserves pose/scale (no snap-reframe)
- ✅ Auto-frame-on-resize animation unchanged
- ✅ `typecheck` + ESLint clean; full unit suite green (12006 passed / 3 skipped)

## Notes

`previewCanvasCamera.tsx` has no sibling test (it didn't before either) — the behavior depends on the live R3F camera-swap timing, which isn't reproducible in jsdom (no real WebGL canvas), so it was verified via browser screenshots. Open to adding a focused `calculateIdealDistance` framing test if preferred.

### DIFF
--- a/src/features/bin-designer/components/PreviewCanvas/PreviewCanvas.tsx
+++ b/src/features/bin-designer/components/PreviewCanvas/PreviewCanvas.tsx
@@ -364,6 +364,7 @@ export function PreviewCanvas() {
             <CameraController
               controlsRef={controlsRef}
               invalidateRef={invalidateRef}
+              projection={projection}
               width={width}
               depth={depth}
               height={height}
--- a/src/features/bin-designer/components/PreviewCanvas/previewCanvasCamera.test.ts
+++ b/src/features/bin-designer/components/PreviewCanvas/previewCanvasCamera.test.ts
@@ -1,0 +1,93 @@
+import { describe, it, expect } from 'vitest';
+import { Vector3 } from 'three';
+import { calculateIdealDistance, calculateBinCenter } from './previewCanvasCamera';
+
+const GRID = 42;
+const HEIGHT = 7;
+const FOV = 45;
+// Mirrors the module-private FRAME_FILL — the live preview frames the bin's
+// bounding sphere to fill this fraction of the viewport.
+const FRAME_FILL = 0.65;
+
+function expectedDistance(
+  w: number,
+  d: number,
+  h: number,
+  fov = FOV,
+  grid = GRID,
+  height = HEIGHT
+): number {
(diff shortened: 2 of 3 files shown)


andymai:
Addressed both review comments in fdca874:

- **Copilot (in-flight animation on projection swap):** Good catch — confirmed the `useFrame` loop would keep applying a stale position/zoom animation (computed in the previous camera's basis) to the camera adopted on a projection swap, defeating `CameraRig`'s pose copy. Both adopt/re-frame branches now clear `posAnimRef`/`zoomAnimRef`. Verified in-browser by toggling projection mid auto-frame (after a width change): the adopted orthographic camera keeps its pose, no drift.
- **Greptile (missing sibling test):** Exported `calculateIdealDistance`/`calculateBinCenter` and added `previewCanvasCamera.test.ts` covering positive/finite distance across sizes, the FRAME_FILL bounding-sphere formula, monotonicity, FOV inverse-scaling, custom grid/height units, and bin-center placement.

Full unit suite green; typecheck + lint clean.

PR #1936, opened 2026-05-29
feat(bin-designer): add synced organization tags to saved designs

andymai:
## What & why

Adds an optional **`tags`** field to saved bin designs so users can organize a growing library. Production cloud-sync data shows design counts are long-tailed — median 1, but **p90=10, p95=15, max 69** per user — so the designs library needs real organization, not just a flat list.

This PR is the **data-model foundation** (no UI yet). It's stage 1 of 3:
1. **Tags foundation** ← _this PR_
2. Designs-manager rebuild (wider grid, tag chips + filtering, bulk delete/export/tag)
3. Header thumbnail quick-switch + management-only layout modal

## How tags work

Tags **sync across a user's devices** by riding alongside the design `name` (a sibling of `params`, not inside it):

- `SavedDesign.tags` + a shared `normalizeTags` helper (trim, dedupe case-insensitively, cap at 12 tags × 32 chars)
- Persisted in IndexedDB via `saveDesign` / `duplicateDesign`; new `updateDesignTags` helper
- Carried through `DesignSyncPayload` and the design sync adapter — LWW: a remote tag array (even empty) wins; a legacy payload with no tags falls back to local
- Sanitized + stored server-side in the design envelope (`sanitizeTags`); client and server caps are kept identical as a documented cross-boundary contract

## Safety

- `tags` is **optional everywhere** (absent = untagged) → no destructive migration; existing saved designs are untouched.
- The sync-admin drift accounting is unaffected: `tags` appears with the same key+value on both the blob envelope and the index-size calc, so it cancels in `blob.size − sizeBytes` exactly like `name`. Verified by new regression tests **and** a clean read-only production audit (`0 findings`, 731 blobs / 122 users).

## Tests

- `normalizeTags` / `sanitizeTags` unit tests (client + server caps)
- Storage round-trip, duplicate carry-over, update + clear, preserve-on-omit
- Sync adapter: tag carry-through in list/get/applyRemote incl. LWW + legacy fallback
- Sync API endpoint: tag sanitize + round-trip + always-array envelope
- sync-admin: tagged & untagged designs are drift-free

typecheck ✅ · lint ✅ (0 errors) · knip ✅ · affected tests ✅


andymai:
Thanks for the review! Addressed in dbfe88a:

- **Control-char parity (the contract gap):** `normalizeTags` now strips `[\x00-\x1F\x7F]` exactly like the server's `sanitizeTags`, so a tag the client keeps can't be rewritten by the server on the next sync pull. Added a test covering interior-tab / null-byte tags.
- **P2 — double normalization:** `updateDesignTags` now passes tags through raw and lets `saveDesign` normalize, matching `updateDesignName`.
- **README:** documented the `tags` field, `updateDesignTags`, and the sync semantics (name-sibling envelope, LWW, omit-preserves / `[]`-clears, client↔server caps must stay identical).

PR #1939, opened 2026-05-29
feat(bin-designer): organize saved designs with tags, filtering, and bulk actions

andymai:
## What & why

Rebuilds the **saved-designs manager** so it scales with the power-user tail. Production cloud-sync data shows design counts are long-tailed — median 1, but **p90=10, p95=15, max 69** per user — and the old `max-w-lg` single-column dialog didn't keep up.

Stage 2 of 3 (builds on the synced `tags` field from #1936):
1. ~~Tags foundation~~ ✅ #1936
2. **Designs-manager rebuild** ← _this PR_
3. Header thumbnail quick-switch + management-only layout modal

## Changes

- **Wider surface** — `max-w-lg` → `max-w-4xl` with a denser auto-fill thumbnail grid (≈5 columns on desktop), so 10–70 designs are scannable.
- **Tags on cards** — each design shows its tag chips; an **Edit tags** action opens a tag editor (chips + add field) backed by `updateDesignTags`.
- **Tag filter bar** — chip toggles above the list (AND-match, with **Clear filters**); hidden when no design has tags.
- **Bulk-selection mode** — a **Select** toggle reveals checkboxes; the bulk bar can **Delete** (single confirm with count), **Export**, or **Tag** a whole batch.

## New, each unit-tested
`useDesignSelection` reducer · `tagFilter` (collect/filter) utils · `DesignTagChips` · `TagInput` · `TagEditDialog` · `TagFilterBar` · `BulkActionBar`. Existing items/dialog/actions extended (chips, selection, edit-tags) with their tests updated.

## Verification

- Unit/component: **91 tests** across the bin-designer manager (logic + rendering).
- i18n: 23 keys added to **all 9 locales**; `check:i18n` (keys/interpolation/values/unused) green.
- Playwright visual spot-check at desktop + tablet — normal grid, tag-filtered view, and bulk-selection mode all render correctly (screenshots below).
- typecheck ✅ · lint ✅ · knip ✅ · design-system ✅

### Screenshots
| Normal (wider grid + chips + filter bar) | Filtered by `kitchen` | Bulk-selection mode |
| --- | --- | --- |
| _attached in review_ | _attached in review_ | _attached in review_ |

## Notes
- Bulk **Tag** adds the entered tags to each selected design (union); per-design removal is via the single-design editor. Bulk remove can follow if needed.

PR #1941, opened 2026-05-30
feat(layout-library): header thumbnail quick-switch for layouts

andymai:
## What & why

Switching layouts was a modal round-trip. This adds a **`LayoutQuickSwitch`** to the header so it's a one-click, **thumbnail-led** action. Production data shows most users keep **1–2 layouts (82% ≤2)**, so recognition-by-shape in the header fits how layouts are actually used — and you flagged early that *the thumbnail must stay*.

This is stage 3 of 3:
1. ~~Tags foundation~~ ✅ #1936
2. ~~Designs-manager rebuild~~ ✅ #1939
3. **Header layout quick-switch** ← _this PR_

## Changes

- **Header trigger** = the active layout's thumbnail + a chevron (replaces the old text "Layouts" button).
- **Dropdown** lists every layout (preview + name, **✓ on the active one**) and switches on click; plus **New layout** and **Manage layouts**.
- **Manage** opens the existing layout manager modal (now reached via the switch rather than a standalone button).
- **Inline header rename is preserved** — the editable layout-name button is untouched, so quick rename still works.
- Removes the now-orphaned `header.layouts` / `header.openLayoutManager` strings.

## Scope notes
- The layout manager modal keeps its **full management UI** (rename/duplicate/delete/share/import/export). Switching from it still works as a secondary path — no regression. I deliberately did **not** strip its grid/sort, since power users with many layouts benefit from it there.
- A mobile full-width sheet variant is a possible follow-up; on small screens the dropdown still works.

## Verification
- Unit: `LayoutQuickSwitch` (6 tests — render/open/switch/active-noop/new/manage) + updated `Header` tests (27).
- i18n: 2 keys across **all 9 locales**; `check:i18n` (keys/interpolation/values/unused) green.
- **Playwright visual check**: header trigger + dropdown render correctly; the dropdown lists layouts with the active ✓, New Layout, and Manage layouts, styled consistently with the app's other dropdowns.
- typecheck ✅ · lint ✅ · design-system ✅ · module-boundaries ✅


andymai:
Thanks! Addressed in 4abfcd6:

- **Barrel import (P2):** `LayoutQuickSwitch` is now exported from `src/features/layout-library/index.ts` and imported via the barrel in `Header.tsx`. I checked the bundle concern (Header lazy-loads `LayoutManagerModal`): the barrel's static re-export does **not** change the main chunk — `LayoutManagerModal` is already in main via a prefetch in `usePrefetchChunks.ts`, and main-bundle gzip is identical (78.9 kB) either way.

Plus a couple of self-review improvements in the same commit:
- The active layout now uses `aria-current` (the check icon is `aria-hidden`), so screen readers announce the current item once instead of via a labelled SVG.
- Extracted `Icon`/`Thumb`/`MenuAction` helpers, removing ~50 lines of duplicated inline SVG/menu markup.

Re-verified: typecheck ✅ · lint ✅ · 33 tests ✅ · Playwright visual check (dropdown renders identically post-refactor).

andymai:
Good catch — fixed in ea219fe.

**Shared-preview trigger:** in shared-layout-preview / embed mode the active layout is `SHARED_PREVIEW_ID` (not a library entry), so the `find() ?? entries[0]` fallback mislabeled the trigger with the first library layout. The trigger now derives its name + thumbnail from the live layout store (`computePreview`) whenever the active layout isn't in the library, so it always reflects what's on screen. The dropdown's switch list is unchanged, and the leading divider is now suppressed when the library is empty. Added a unit test for the shared-preview path.

(The three comments were facets of this one issue — the imports concern is resolved since the fix adds `useMemo` / `useLayoutStore` / `computePreview`.)

Verified: typecheck ✅ · lint ✅ · 34 tests ✅ · module-boundaries ✅ · knip ✅

PR #1981, opened 2026-06-03
feat: polygon + slot cutout shapes for bit/socket organizers

andymai:
## What & why

The cut editor could only make rectangles, circles, and freeform pen paths — so building a **hex-bit, socket, or Allen-key organizer** meant hand-tracing every hexagon. This adds the two primitives those layouts actually need:

- **Polygon** — a parametric regular N-gon (3–12 sides, **flat-top hexagon** by default) sized by **across-flats** so you can type a hardware spec straight off the datasheet (1/4" hex = 6.35mm). A **common-sizes** preset menu offers the 1/4" hex bit and metric Allen sizes.
- **Slot** — a stadium/capsule (rounded rectangle with fully-rounded ends) for drivers and tools laid flat.

Both insert shapes — **plus circles, for SAE/metric socket trays** — gain an optional **insertion-clearance** offset (default +0.2mm) so a part cut to spec actually drops in, and circles get socket-drive / diameter presets.

## How it fits the existing editor

A polygon's vertices are derived to fill its `width × depth` bounding box, so **every existing system reuses unchanged** — bounds, resize handles, rotation, alignment, clamping, and label placement all keep reasoning in width×depth space. Only three places learn about the new shapes:

- **Generator** (`cutoutBuilder`) — builds the polygon wire / slot rounded-rect and applies clearance.
- **2D boolean preview** (`cutoutToPolygon`) — so grouped Pathfinder ops match the exported mesh.
- **Editor renderer** — polygons route to a points-driven mesh (the pen-path renderer's core, extracted to `outlineShapeGeometry`); slots render as SDF rounded boxes.

The polygon/slot vertex math lives in one shared source (`src/shared/utils/cutoutPolygon`) imported by both the worker and the editor.

## UX

- Toolbar buttons for **Polygon** and **Slot**, with `G` / `S` shortcuts.
- Property panel gains **Sides**, **Across flats**, **Clearance**, and a **Common sizes** preset menu, shown per shape.
- Full i18n across all locales.

## Verification

- New unit tests for the polygon math, outline derivation, presets, helpers, boolean-preview rings, distance-field bake, and the property controls.
- Real-kernel generator scenarios for hexagon, hexagon-with-clearance, and slot — all produce valid geometry (clearance correctly enlarges the cut).
- Manual visual check (Playwright): a hexagon renders flat-top in the 2D editor and generates a clean hexagonal cutout in the 3D preview.
- `typecheck`, `lint`, and all i18n checks pass.

## Roadmap

[...]

PR #1982, opened 2026-06-03
feat: entry chamfer for cutouts (self-centering bit insertion)

andymai:
> **Stacked on #1981** (polygon + slot shapes). Base is `feat/cutout-polygon-slot-shapes`; retarget to `main` once #1981 merges. The diff below is chamfer-only.

## What & why

Bits, sockets, and Allen keys are fiddly to drop into a straight-walled hole — they catch on the rim. This adds an optional **entry chamfer**: a ~45° bevel at the top of the cut that flares the opening outward so parts **self-center and slide in**.

Off by default, so every existing design is cut identically.

## How

The chamfer is built as a **3-section ruled loft**:

1. nominal profile at the bottom (`z = 0`),
2. nominal profile again at `z = cutDepth − chamfer` (the straight wall section),
3. an **outset** profile at the top rim (`z = cutDepth`).

The outset grows each bounding dimension by `2·chamfer` and the corner radius by `chamfer`, so the vertical drop equals the horizontal flare — a true 45°. A single shared profile builder feeds both the straight-extrude path and the loft, so they can't drift.

- Applies to **rectangle / circle / polygon / slot** (`CHAMFER_SHAPES`). Freeform paths are excluded — a constant-offset outset of an arbitrary bézier isn't well-defined.
- Clamped so a straight wall always remains below the bevel (`cutDepth − chamfer > 0`); the panel slider is capped by cut depth.

## UX

- Property panel gains an **Entry chamfer** slider (mm), shown for the parametric shapes and hidden when the cut is too shallow to bevel.
- Full i18n across all locales.

## Verification

- Real-kernel generator scenarios for chamfered hexagon, circle, and rectangle — all produce valid geometry, with the flare adding the expected triangles (chamfered circle 2654 vs plain ~2548).
- Updated property-panel unit tests cover the chamfer field appearing per shape and disappearing at shallow depth.
- `typecheck`, `lint`, and all i18n checks pass.

## Roadmap

Second of three staged PRs: shapes (#1981) → **chamfer (this)** → live parametric array.


andymai:
Superseded by #1985, which combined the full stacked cutout initiative (shapes + entry chamfers + editor UX + parametric arrays) into a single squash-merge to `main` (c6517a589) to ship as one deploy. All of this PR's work is included there.

PR #1984, opened 2026-06-03
fix(generation): include overhang expansion in innerW/innerD

andymai:
## Summary

- `innerW` and `innerD` in `deriveDimensions` (pipeline context) were computed from nominal outer dimensions, without the per-side overhang expansion
- `boxBuilder` already added the expansion to its inner footprint, so the physical shell cavity was larger than what all interior feature builders (compartments, scoops, label tabs, wall patterns, handles, inserts) thought was available
- Interior features were therefore constrained to the smaller nominal space and didn't use the room created by the overhang

## Fix

Resolve overhang before computing `innerW`/`innerD` and include `addW`/`addD` from the overhang expansion:

```ts
// Before
const innerW = outerW - 2 * params.wallThickness;

// After
const overhang = resolveOverhang(...);
const ovhExp = hasOverhang(overhang) ? overhangExpansion(overhang) : null;
const innerW = outerW + (ovhExp?.addW ?? 0) - 2 * params.wallThickness;
```

No-overhang bins are unaffected (`addW` / `addD` = 0). Polygon-mask bins are also unaffected (overhang is already suppressed for them). The duplicate overhang resolution that appeared later in `deriveDimensions` is removed.

## Test plan

- [x] `context.test.ts` — new unit tests verify `innerW`/`innerD` grow by `addW`/`addD` with overhang and are unchanged with zero overhang
- [x] `binGenerator.scenario.overhang.test.ts` — new geometry regression test builds a bin with overhang + scoop, asserts structural validity and that the mesh differs from the no-overhang counterpart (confirming scoop uses the expanded interior)
- [x] All 6 overhang scenario tests pass
- [x] All 14 context unit tests pass
- [x] TypeScript and ESLint clean

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.overhang.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.overhang.test.ts
@@ -113,6 +113,37 @@ describe('overhang with interior features', () => {
     expect(scoopDeltaWithOverhang).toBeGreaterThan(0);
     expect(scoopDeltaWithOverhang).not.toBe(scoopDeltaNoOverhang);
   });
+
+  it('asymmetric overhang: interior features are structurally valid (centering offset applied)', () => {
+    const generateBin = getGenerateBin();
+    // right-only overhang: cavity centre shifts +X by 5mm (offsetX = 5).
+    // Before the centering fix, scoops would extend 5mm into the left wall.
+    // After the fix, all features translate by (innerOffsetX, innerOffsetY).
+    const result = generateBin(
+      buildParams({
+        width: 2,
+        depth: 2,
+        overhang: { left: 0, right: 10, front: 0, back: 0 },
+        scoop: { enabled: true, radius: 'auto' },
+      }),
+      undefined,
+      true
+    );
+    assertStructurallyValid(result, 'asymmetric overhang with scoop');
+
+    // front-only overhang: cavity centre shifts +Y by 5mm (offsetY = 5).
+    const result2 = generateBin(
+      buildParams({
+        width: 2,
+        depth: 2,
+        overhang: { left: 0, right: 0, front: 0, back: 10 },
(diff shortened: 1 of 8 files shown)


andymai (commented):
Addressed both review comments:

1. **Regression test** — replaced the whole-bin  comparison with a delta-based assertion: scoop-delta (with-scoop − without-scoop) under no-overhang vs with-overhang. Before the fix both deltas would be identical (innerD was nominal in both cases); after the fix the overhang case has a larger cavity so the scoop carves more geometry.

2. **Asymmetric centering** — confirmed the issue is real. Added a NOTE comment in `context.ts` documenting the limitation: for asymmetric overhang `offsetX/Y ≠ 0` and feature builders still centre at origin, so there's up to `|(R−L)/2|` of positional error. The proper fix needs `innerOffsetX/Y` in `BinDimensions` threaded through ~8 feature builders — that's a meaningful follow-up task, not a quick line. The current PR fixes the primary bug (SIZE) which is correct for symmetric overhang (the common case).

andymai (commented):
Addressed both review comments:

1. **Regression test** — replaced the whole-bin triangle count comparison with a delta-based assertion. The test now computes scoop-delta (with-scoop minus without-scoop) under no-overhang vs with-overhang conditions. Before the fix both deltas would be identical (innerD was nominal either way); after the fix the overhang case has a larger cavity so the scoop carves different geometry.

2. **Asymmetric centering** — confirmed the issue is real. Added a NOTE comment in `context.ts` documenting the limitation: for asymmetric overhang, offsetX/Y != 0 and feature builders still center at origin, producing up to |(R-L)/2| of positional error. The proper fix requires carrying `innerOffsetX/Y` through BinDimensions and updating ~8 feature builders — scoped as a follow-up. This PR fixes the primary SIZE bug which is fully correct for symmetric overhang (the common case).

PR #1993, opened 2026-06-03
feat(cutouts): dock the cutout inspector as a resizable side panel

andymai:
## What

Replaces the free-floating cutout inspector overlay with a **docked properties column** between the 2D canvas and the 3D preview. First of a 3-PR overhaul of the cutout editor's properties UI.

| | Before | After |
| --- | --- | --- |
| Placement | Floats over the canvas, covering the shapes being edited | Fixed column to the right of the canvas — never overlaps |
| Width | Fixed 220px | Resizable (drag left edge), persisted |
| Reclaim space | — | Collapse to a thin rail, persisted |

## Why

The floating panel obscured the very shapes you were editing (visible in the original screenshot where it covered the hex grid). Docking it removes the obstruction entirely and gives the editor a desktop-class, predictable layout. Chrome mirrors the main `ParameterPanel` sidebar for visual consistency.

## How

- New `InspectorDock` — docked, resizable, collapsible column; width + collapsed state persisted via `inspectorDockStorage` (localStorage, same pattern as `splitRatioStorage`).
- New `InspectorContent` — the selection-aware body (single + multi-select + lock badge + empty placeholder), extracted from the old `FloatingInspector` with all positioning/jitter-lock logic dropped.
- `CutoutWorkspace` renders the dock as a sibling column instead of a canvas overlay.
- Deleted `FloatingInspector` (+ test); ported content coverage to `InspectorContent.test`.

**Controls are unchanged** — this is purely a layout move. The number-first control kit + hardware-preset chips (PR2) and the multi-select / empty-state / actions work (PR3) follow.

## Verification

- `typecheck` clean, lint clean, 14 new unit tests pass (storage round-trip, content states, collapse/resize/persist)
- Playwright spot-check confirmed: dock renders beside the canvas (no overlap) and collapses to a rail reclaiming canvas width
- i18n: 6 new keys added across all 9 locales (parity check passes)

PR #1994, opened 2026-06-04
feat(cutouts): number-first inspector controls + hardware-size chips

andymai:
## What

Second PR of the cutout-editor overhaul. Replaces the inspector's sliders with **number-first controls** and promotes hardware presets to **quick-pick chips**.

> Stacked on #1993 — review/merge that first. The base branch is `feat/cutout-inspector-dock`.

## Why

The sliders made precise values fiddly — their knobs jammed at the left of the track (visible in the original screenshot where Sides/Across-flats/Scoop/Clearance all sat at min). A bit/socket organizer is all about exact spec sizes, so the number should be primary.

## How

- **`CompactNumberInput`** (already the X/Y/W/H control) gains:
  - **drag-scrub** — drag the label to adjust coarsely (cursor `ew-resize`)
  - **modifiers** — `Shift` = ×10 step, `Alt` = fine (÷10), on both scrub and arrow keys
  - **step-snapping** + click-to-type-exact (kept)
  - It now drives rotation, cut depth, corner radius, scoop W/D, clearance, and chamfer too — laid out in tidy 2-col grids.
- **Polygon sides** → design-system `Stepper` (discrete +/−), which fits a small integer range far better than a slider.
- **Hardware presets** → `CutoutPresetChips`: one-tap chips for the common sizes (active size highlighted), full catalog behind a **"More…"** dropdown (reuses `CutoutPresetMenu`).

These controls are shared with the sidebar `CutoutPropertyPanel`, so the improvement lands there too.

## Verification

- `typecheck` + lint clean; 40 unit tests pass (23 `CompactNumberInput` incl. new scrub/Alt-fine/slider-role coverage, preset chips, updated scoop/shape/inspector tests)
- Playwright spot-check of a polygon selection: Stepper for Sides, compact number fields, and hardware chips all render correctly in the docked panel
- i18n: `sizePresetMore` key added across all 9 locales

### DIFF
--- a/src/features/bin-designer/components/CutoutWorkspace/CutoutScoopControls.test.tsx
+++ b/src/features/bin-designer/components/CutoutWorkspace/CutoutScoopControls.test.tsx
@@ -3,8 +3,8 @@ import { render, screen, fireEvent } from '@testing-library/react';
 import { CutoutScoopControls } from './CutoutScoopControls';
 import type { Cutout } from '@/features/bin-designer/types';
 
-vi.mock('@/features/bin-designer/components/controls/SliderInput', () => ({
-  SliderInput: ({
+vi.mock('@/shared/components/CompactNumberInput', () => ({
+  CompactNumberInput: ({
     label,
     value,
     onChange,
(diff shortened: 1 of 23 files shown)


andymai:
Rolled into #1995 (single squash-merge / one deploy).

PR #1997, opened 2026-06-04
feat(cutout-editor): polish side-panel controls + smart fit defaults

andymai:
## What & why

A polish pass across the cutout-editor side panel — making the controls read correctly, removing rough edges, and giving new cutouts print-ready defaults.

### Controls that look like what they are
- **Segmented controls** now use the design-system recessed-track + raised-pill treatment, so the wall-thickness, array-mode, grid-size, label-tab, and text-side pickers read as segmented controls instead of a loose row of chips. The accent on/off fragments are kept for genuine boolean toggles (scoop edge chips, eyedropper).
- **Slider handles no longer clip** at the track ends — the Slider primitive is inset by the thumb radius so scroll/overflow ancestors (the inspector dock) can't shear the round handle at 0%/100%. Fixes it everywhere sliders are used.

### Steppers where steppers belong
- **Array settings** (cols / rows / pitch / count / radius / angle) are now `+/-` steppers instead of sliders.
- **Tolerance + entry chamfer** are now 0.2 mm steppers that still accept off-grid fractional values typed by hand, for precise tuning.

### Physically feasible by construction
- Array **counts, pitch, and radius are clamped to feasible bounds** — you can't grow an array past the bin footprint (or the instance cap). Each field is bounded against the others as-is, so editing one never silently rewrites another.

### Print-ready defaults
- New **insert-style holes** (circle / hex / polygon / slot) now seed a **self-centering entry chamfer**, scaled to the hole's tightest dimension (~10%) and clamped to a tasteful **0.4–0.8 mm** — a crisp edge-break on small holes, never a funnel on large ones. Tolerance continues to seed at 0.2 mm. Existing saved designs are untouched (additive, creation-time only).

### Discoverability + clarity
- Number labels now carry a **persistent After-Effects-style scrub affordance** (always-on dotted underline + `ew-resize` cursor, brighter on hover) so drag-to-scrub is discoverable at rest.
- The hardware-sizes overflow control is now a **count + rotating chevron disclosure** instead of a `+N` chip that read like "add a size."

## Notes
- `defaultEntryChamfer` / `maxEntryChamfer` and `arrayFieldBounds` are pure and unit-tested.
- Full suite green (12,766 tests), typecheck + lint clean.


andymai:
Thanks for the reviews — addressed in `f799689`:

**Copilot**
- **Staggered `maxCols` / `maxPitchX` with `rows === 1`** (`cutoutArray.ts`): the half-pitch X shift only exists once there's an odd row to shift, so it's now gated on `rows > 1`. Single-row staggered arrays get the same ceilings as a plain grid. Added a regression test asserting parity.
- **`maxRadius` ignored master position**: now bounded by the master box's smallest edge clearance (`min(x, y, binW−(x+w), binD−(y+d))`), so an off-center master can't allow a ring past the nearest edge. Test updated to cover centered vs off-center.
- **Fit steppers rounding typed fractionals (`0.25` → `0.3`)**: real display/data mismatch. Added an optional `inputDecimals` passthrough (`StepperControl` → `DeferredNumberInput`); the fit fields render at 2 decimals so off-grid typed values display accurately. Default stays 1dp everywhere else.

**Greptile (P2)**
- `SEGMENT_PILL_INACTIVE` vs `SEGMENT_INACTIVE` identical strings: added a comment noting the duplication is deliberate (shared "off" state; only the *active* treatments differ) so a future cleanup doesn't collapse the two systems.

Also fixed the **Conventional Commit** check by retitling the PR to `feat(cutout-editor): …`. Full suite green (12,767), typecheck + lint clean.

PR #2006, opened 2026-06-04
fix(cutouts): translate path vertices for array instances

andymai:
## Problem

A **path** (pen-tool) cutout with a parametric **array** rendered all its instances stacked on top of the master in the 2D cut editor, instead of spread across the grid/ring. "Flatten array" baked that collapse into real overlapping cutouts.

Follow-up to #2004 (noted there as a separate out-of-scope bug).

## Root cause

`expandCutoutArray` (`src/shared/utils/cutoutArray.ts`) shifts each non-master instance's `x`/`y`/`rotation` but carried the `path` array verbatim. Path vertices are **absolute** coordinates, so every instance's path stayed at the master's location while its bounding box (`x/y/width/depth`) moved away.

Consumers that read `instance.path` directly therefore collapsed:
- **Editor** — `CutoutArrayMeshes` → `CutoutShapeMesh` → `PathShapeMesh` centers on the path's own bounds, so all instances drew at the master.
- **Flatten array** — `flattenCutoutArray` (`cutoutHelpers.ts`) bakes `expandCutoutArray` output into independent cutouts, persisting the collapse.

The **generation worker** was already correct: `buildArrayUngroupedCutouts` builds geometry once from the *master* and only uses instances for x/y/rotation, so it never read `instance.path`. (This is why the live 3D bin looked fine while the editor preview didn't — a 2D/3D mismatch.)

## Fix

Translate path vertices by the same per-instance offset as `x`/`y`. Relative bezier handles are unaffected. One-line-per-point map in `expandCutoutArray`; the worker is unchanged and stays correct.

## Tests (TDD)

`cutoutArray.test.ts` — a grid array of a path cutout must translate each non-master instance's vertices to follow its position (and leave the master + relative handles untouched). Watched it fail (path stayed at master) before the fix.

## Verification

- `pnpm typecheck` clean, lint clean, all pre-commit gates pass
- `cutoutArray` suite 24/24, `CutoutArrayMeshes` + `cutoutHelpers` suites green
- worker path-cutout scenario still green (worker behavior unchanged)

### DIFF
--- a/src/shared/utils/cutoutArray.test.ts
+++ b/src/shared/utils/cutoutArray.test.ts
@@ -151,19 +151,28 @@ describe('expandCutoutArray', () => {
       width: 10,
       depth: 10,
       path,
-      array: cfg({ mode: 'grid', cols: 2, rows: 1, pitchX: 30, pitchY: 30 }),
+      array: cfg({ mode: 'grid', cols: 2, rows: 2, pitchX: 30, pitchY: 30 }),
     });
+    // Instance order is row-major: [master, +x, +y, +x+y].
     const out = expandCutoutArray(c);
 
-    // Second instance shifts +30 in x; its path must shift with it.
+    // Instance 1 (col 1, row 0) shifts +30 in x; its path must shift with it.
     expect(out[1].x).toBe(40);
     expect(out[1].path?.map((p) => [p.x, p.y])).toEqual([
       [40, 10],
       [50, 10],
       [45, 20],
     ]);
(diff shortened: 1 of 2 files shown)


andymai:
Addressed in 9f708713 — extended the array-path test to a 2×2 grid so the Y-axis translation branch (`y: p.y + inst.dy`) is exercised with a non-zero `dy` (asserting the +y instance's vertices and handle preservation), in addition to the +x instance and the untouched master.

PR #2021, opened 2026-06-05
feat(labs): graduate faster live preview out of labs

andymai:
## What

Graduates the `manifold_preview` ("Faster Live Preview") Labs flag from **experimental** to **graduated** — the Manifold draft preview is now on for everyone by default. Exports continue to use the exact occt-wasm engine.

## Why

The feature shipped behind the flag in #2009 and has no known fidelity gaps remaining (honeycomb closed in #2014). Time to make the faster live preview the default experience.

## Changes

- `src/core/labs/features.ts`: `status: experimental → graduated`, add `graduatedAt`, `requiresRefresh: false`, drop the toggle warning. The card moves from the Labs toggle list into the "What's New" section automatically.
- Feature READMEs (generation / bin-designer / baseplate): reworded from "Labs flag on" to "graduated, always on". The flag id remains the runtime gate (`isFeatureEnabled('manifold_preview')`, now always true); a draft init failure still degrades gracefully to exact-only.

## Perf — does the draft slow exact generation?

Investigated (the concern was that exact OCCT felt slower) and **refuted**:

- occt-wasm is a **single-threaded, non-pthread build** running in its own worker — it uses one core. The Manifold draft worker is also single-threaded, so on any ≥2-core machine it lands on a different core and can't steal OCCT's cycles. `bridge.generate()` is byte-for-byte identical flag on/off.
- A/B (chromium, PerfOverlay `totalMs` = exact-only, 6 cache-missing builds of a ~1.5s bin): exact median **off 1474.5ms vs on 1479ms (+0.3%, within noise)**; no cold-start regression.

The likely source of the perception: the instant coarse draft makes the (unchanged) exact latency newly visible as it sharpens.


andymai:
Addressed Greptile's note: updated the stale `acquirePreview` JSDoc (and the related field comment) in `BridgeManager.ts` — the flag-off→null path is no longer reachable now that `manifold_preview` is graduated; null now means draft-kernel init failure (best-effort, non-fatal). The defensive flag gate is retained per the standard graduated-feature pattern. Commit 409b35ca9. Copilot had no comments.

PR #2023, opened 2026-06-05
perf(generation): defer base-socket fuse to export for faster cold preview

andymai:
## What

The base-socket → body boolean fuse is the dominant cost of cold bin generation — **~1.8 s of a ~2.1 s 6×6 shell (≈80%)**. This defers it.

The base socket (the grid of feet) is **never cut by features** (compartments/cutouts carve walls + floor, not feet) and only meets the body at a **hidden internal interface**. So on the **preview** path the socket is now kept out of `ctx.solid` (new `ctx.deferredSolid`); the tessellate stage meshes the body and socket separately and concatenates them (`mergeShapeMeshes`) — visually identical, no boolean. The **export** path still fuses for a watertight model.

## Impact

Cold `base` stage on a 6×6 socket bin: **~2119 ms → ~776 ms (2.7×)**. Composes with the Manifold draft preview (draft instant → exact sharpen ~2.7× sooner).

## Correctness (validated)

- **Export geometry byte-identical** — measured volume 373549 before and after (stash A/B); all `forExport=true` scenarios pass unchanged.
- **Preview socket present + aligned** — z-bounds match export exactly; `translateStage` shifts the deferred socket with the body.
- **Full unit suite green** (12809 tests). 109 preview scenario snapshots shifted by ±4–12 triangles (the socket↔body junction is no longer booleaned) — updated; export snapshots untouched.
- **Visual check** (Playwright) — bin renders cleanly with no artifacts.
- Unit test added for `mergeShapeMeshes`.

## How it was found

Per-stage benchmarking isolated the fuse as the bottleneck; OBB acceleration (kernel rebuild) and OCCT glue both proved inert/unsafe for this axis-aligned geometry. Deferring the fuse for preview was the lever that actually moved it.

### DIFF
--- a/src/features/bin-designer/hooks/useGeneration.ts
+++ b/src/features/bin-designer/hooks/useGeneration.ts
@@ -55,6 +55,9 @@ function toMeshPayload(result: BridgeResult): GenerationResult {
  * A monotonic token guards arbitration: a draft is dropped if a newer edit has
  * started or the exact result for its edit has already landed.
  */
+/** Idle gap after the exact result before speculatively warming the export shell. */
+const EXPORT_WARM_IDLE_MS = 2000;
+
 export function useGeneration(): void {
   const bridgeRef = useRef<GenerationBridge | null>(null);
   const previewBridgeRef = useRef<GenerationBridge | null>(null);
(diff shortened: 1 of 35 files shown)


andymai:
Addressed the review:
- **Issue 1 (handle leak on fuse failure)** — fixed in 7d682dab0. The overhang-feet + export fuses now run in a try/catch that disposes `body`/`socket`/`feet` on any throw, restoring the `withScope` exception-safety. (Can't use `withScope` directly here because the preview path needs *two* survivors — body and the deferred socket.)
- **Issue 3 (duplicate concatFloat32)** — fixed: exported from `mesh.ts` and reused in `tessellateStage`.
- **Issue 2 (shellKey carries socket params)** — leaving as-is intentionally. The key is *correct* (just discriminates a bit more than the body strictly needs); a magnet/screw toggle now produces a cheap body cache-miss (~366ms body rebuild) rather than the old full-shell rebuild, so it's no worse than before, and re-deriving a separate body-only key risks a collision bug for a bounded eviction win. Noted as a follow-up optimization.

andymai:
Addressed the review finding (commit 56590a77c): in the export path, the fused shell is now cached and cloned **before** `body`/`socket` are deleted, so if `translate` throws the catch block only ever deletes live handles (no double-delete), and `full` is owned by the cache rather than leaked. typecheck + export-cache + scenario tests green.

PR #2028, opened 2026-06-05
feat(export): determinate progress bar instead of a spinner

andymai:
## What

Export currently shows an indeterminate spinner. This makes it a **determinate progress bar**.

## How

The generation pipeline already emits per-stage `PROGRESS` (used by the live generating indicator); export just never tapped into it. Now:

- **Worker** — `exportBin`/`runExportAttempt` accept an `onProgress`; generation stages map to 2–85% and encode to 100%. `handleExport` + `handleExportCombined` forward it via `reportProgress(exportRequestId, …)`.
- **Bridge** — `PendingExport` gains an `onProgress`; `runExport`/`exportBin`/`exportCombined` register it, and the `PROGRESS` handler routes export-tagged progress to it (export requests live in `pendingExports`, keyed by slot, separate from the live generate request).
- **UI** — `useExport` tracks 0–1 progress and the bin `ExportDialog` feeds the existing `ProgressBar` (the same one multi-piece/split exports already use), with an "Exporting…" label. The spinner stays as the fallback when no progress is reported.

## Notes

- Granularity: single-bin exports are stage-level (a few steps — honest but coarse); the per-piece bar that split/multi-color exports already had is unchanged. The shared dialog's `exportProgress` prop gained an optional `label` so it shows "Exporting…" for the percentage case vs "piece X of Y" for multi-piece.
- Tests: useExport (18) + ExportDialog component (52) + typecheck + lint all green.

### DIFF
--- a/src/features/bin-designer/components/ExportDialog/ExportDialog.tsx
+++ b/src/features/bin-designer/components/ExportDialog/ExportDialog.tsx
@@ -62,6 +62,7 @@ export function ExportDialog() {
     estimates,
     isExporting,
     isExportingBin,
+    exportProgress: exportFraction,
     downloadBin,
     needsSplit,
     splitPieceCount,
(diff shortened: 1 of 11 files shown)


andymai:
Addressed review (f04880b01) — both the PR bot and a local `greptile review` CLI run flagged these:
- **Progress stalls at 95%** → `handleExportCombined` now reports 100% on all three completion branches (single-bin, STEP, multi-piece).
- **Retry regression** → `useExport` resets progress inside the resilience lambda, so a retry restarts the bar at 0 instead of jumping backwards.
- **Deferred-socket handle leak** (CLI-only finding, in the merged defer code) → `tessellateStage` now disposes the deferred socket in a `finally`.

typecheck + useExport (18) + scenario tests green.

PR #2032, opened 2026-06-05
feat(bin-designer): unify wall-cutout and handle panel UI

andymai:
## What & why

The wall-cutout and handle panels in the bin designer had drifted apart visually — the handle panel was recently modernized to the design-system segmented-control look, while the wall-cutout panel still used older hand-rolled buttons. This brings both to one shared standard and does a UX polish pass.

Both panels now render from a single set of shared primitives, so they read identically and can't drift apart again.

## Changes

**New shared primitives** (`src/features/bin-designer/components/panel/shared/`, each with sibling tests):
- **`ShapePicker`** — design-system segmented control (recessed track + raised pill); replaces the wall-cutout panel's custom buttons
- **`SideSelector`** — new **spatial cross** layout: Back top, Left/Right flanking a center bin glyph, Front bottom — replaces the abstract flat L/R/F/B row. Supports per-side disabled state (e.g. the back handle while a label tab occupies that wall)
- **`StepperField`** — standardized label with inline unit, e.g. `Height (mm)` / `Span (%)`
- **`AdvancedDisclosure`** — collapsible secondary controls (cutout position/offset; handle vertical-position/count), auto-expanding when values are non-default
- shared `LinkIcon` / `ChevronIcon` (previously duplicated per panel)

**Behavior parity**
- Handle linked-state now persists across sessions via a new `handlesLinked` setting (was ephemeral `useState`), matching the wall-cutout panel
- Help text shown contextually rather than always-on

## Verification
- `typecheck`, `lint`, `check:i18n` (all locales) green
- Full Vitest suite passes (incl. new primitive tests)
- Playwright visual spot-check confirmed both panels render with identical layout

Handle-specific controls (physical-dimension readout, corner radius, chamfer, support note) are genuine functional differences and were intentionally preserved.

### DIFF
--- a/src/core/store/settings.types.ts
+++ b/src/core/store/settings.types.ts
@@ -195,6 +195,12 @@ export interface UserSettings {
    */
   wallCutoutsLinked: boolean;
 
+  /**
+   * Whether the handles editor edits all active sides together.
+   * Persisted so unlinking to adjust sides independently survives a reload.
+   */
+  handlesLinked: boolean;
+
   // Print view preferences
   printViewSettings: PrintViewSettings;
 
@@ -311,6 +317,7 @@ export const DEFAULT_SETTINGS: UserSettings = {
   layoutManagerViewMode: 'grid',
   designListViewMode: 'grid',
   wallCutoutsLinked: true,
+  handlesLinked: true,
 
   // Print view preferences
   printViewSettings: { ...DEFAULT_PRINT_VIEW_SETTINGS },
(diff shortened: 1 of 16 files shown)


andymai:
Addressed the review feedback in 0071e6a:

- **`SideSelector` disabled normalization** (Copilot): the component now forces a `disabled` side to read as off for both `aria-checked` and styling, regardless of the `active` the caller passes — the docstring promise is now enforced in the component, not just at the call site. `HandleSection` passes the raw `enabled` flag through. Added a test covering `active: true` + `disabled: true` → `aria-checked="false"`.
- **`handlesLinked` persistence coverage** (Copilot): added regression tests mirroring the `wallCutoutsLinked` suite (default linked, toggle writes `settings.handlesLinked`, fresh mount reads the stored value) and reset the settings singleton in `useHandleSection`'s `beforeEach` to prevent cross-test leakage.

Greptile's review check passed with no inline comments. typecheck / lint / i18n / full unit suite green.

PR #2038, opened 2026-06-06
fix(cqrs): build mutations over a live commandBus shim to survive chunk-cycle races

andymai:
## What

Fixes the recurring **\`TypeError: Cannot read properties of undefined (reading 'dispatch')\`** in \`addBin\` — drawing a bin silently breaks for affected users (surfaced again in PostHog error tracking, last seen today on Edge/Windows).

This is the **3rd recurrence** of the chunk-level static-import-cycle / stale-chunk class (#1466, then the lazy-build fix in #1563).

## Why the prior fix wasn't enough

#1563 made the mutations singleton lazy, but \`getCqrsMutations()\` still **snapshotted \`commandBus\` by value** at first read. A race where the binding is briefly \`undefined\` (chunk cycle settling, or a stale chunk that resolves the binding late) poisons the singleton **permanently** — every subsequent mutation closes over \`undefined\` and throws on \`dispatch\`.

## The fix

Build the adapter over a \`liveCommandBus\` shim whose arrow methods **re-read the \`commandBus\` binding on every dispatch** — never a snapshot. The race now self-heals once the binding resolves. When the bus is genuinely unavailable (truly stale chunk), it throws a clear *"stale build — reload the page"* error instead of the cryptic native one, which also improves error-tracking signal.

The object-literal-with-deferred-arrow-read shape is the sanctioned pattern that passes the \`local/no-init-time-imported-call\` ESLint rule added in #1566.

## Tests

New \`MutationsContext.test.tsx\` (3 cases) mocks \`@/core/cqrs\` with a getter-based \`commandBus\` to flip it undefined→defined and proves the singleton no longer poisons.

- ✅ typecheck, eslint clean
- ✅ new tests pass; full suite green

## Also done (out-of-band, no code)

Added two PostHog error-tracking suppression rules scoped tightly by \`\$exception_fingerprint\` for browser-extension noise that was burying real issues: \`Error: undefined\` from the init.ts global handler (232 occ) and \`CustomEvent … isTrusted\` (222 occ).

### DIFF
--- a/src/shared/contexts/MutationsContext.test.tsx
+++ b/src/shared/contexts/MutationsContext.test.tsx
@@ -1,0 +1,80 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import { ok } from '@/core/result';
+import { layerId, categoryId, binId } from '@/core/types';
+import type { Bin } from '@/core/types';
+import type { CommandBus } from '@/core/cqrs';
+import type * as CqrsModule from '@/core/cqrs';
+
+// Controls what the mocked `commandBus` binding resolves to. The shim in
+// MutationsContext reads it live on every dispatch, so flipping it mid-test
+// simulates a chunk-cycle/stale-chunk race where the binding resolves late.
+let currentBus: CommandBus | undefined;
+
+const fakeBus: CommandBus = {
+  dispatch: vi.fn(() => ok({ value: binId('bin_fake'), events: [] })),
+  use: vi.fn(),
+  resetMiddleware: vi.fn(),
+};
+
+vi.mock('@/core/cqrs', async (importOriginal) => {
+  const actual = await importOriginal<typeof CqrsModule>();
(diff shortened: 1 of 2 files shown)


andymai:
Thanks for the reviews — addressed all three in f1a64ad05:

- **Greptile (shim guard consistency):** extracted `requireCommandBus()` and routed `dispatch`/`use`/`resetMiddleware` through it, so every shim method gives the same actionable stale-build error instead of the native `TypeError`. DRY'd into one helper rather than duplicating the guard.
- **Copilot (test masking):** each test now re-imports the module under test via `vi.resetModules()` so the module-scoped singleton is rebuilt per case — the regression test genuinely builds the singleton while the bus is `undefined`. `renderHook` + `useMutations` are co-imported after the reset to keep a single React instance.
- **Copilot (comment wording):** dropped the "arrow methods" phrasing; the shim uses method shorthand.

typecheck + lint clean, 3 tests green. CI re-running on the new commit.

PR #2042, opened 2026-06-06
feat: revamp Ko-fi support prompts in header and export flow

andymai:
## What & why

Moves the Ko-fi support ask to the surfaces that actually convert, and retires the one that didn't.

### Header
- The Ko-fi entry is now a prominent accent-colored **"Support me on Ko-fi"** pill — a native reproduction of the official Ko-fi widget (the real cup logo + its wiggle animation, light label), tinted with the user's chosen accent color instead of a hardcoded blue. (The official `<script>` widget can't run here — CSP blocks third-party scripts — and can't use the accent color, so it's reproduced natively.)
- Trimmed the verbose "Press `?` for help" hint to a compact "Help" to make room.

### Export flow (bin designer + baseplate)
- After a successful export, the dialog shows an inline **success view**: a download confirmation, a low-pressure support ask, and a free **"star on GitHub"** fallback for people who won't tip.
- **Gated for good UX**: never shown on a user's *first* export (too early — they haven't printed it yet), shown from the **2nd export** on, and **at most once per 30 days** so frequent exporters are never nagged. Everyone else just gets a quick "Export complete" toast + auto-close.

### Nudges
- Removed the timer-based Ko-fi nudge toast — it converted poorly and fired on an arbitrary timer rather than at a value moment. The feedback nudge is unchanged; its copy was tightened across all 10 locales.

## Measurement
The export CTA is tagged `kofi_clicked` with `source: bin_designer_export` / `baseplate_export`, so its conversion can be compared head-to-head against the header pill (`source: header`).

## Notes
- New `exportSupportGate` (returning-maker + cooldown logic) lives in `shared/` with its own storage key — no cross-feature import into the engagement slice.
- Mobile Ko-fi links left as-is intentionally (compact accent "Tip" links suit the narrow title bar).

## Testing
- Typecheck, ESLint, module-boundary, and all 4 i18n checks pass.
- Unit tests for the gate, the support prompt, and the nudge hook (489 tests across touched areas).
- Manually verified end-to-end: first-export fast path vs. second-export support view (desktop + mobile), and the accent pill across multiple accent colors.


andymai:
Thanks for the review — addressed in 91d1e37be:

**Fixed**
- **Gate robustness (Copilot, `exportSupportGate.ts`)** — `load()` now validates the persisted shape, so a malformed payload (non-numeric `exportCount`, or an unparseable `lastShown`) can no longer poison the counter with `NaN` and permanently suppress the prompt. Invalid timestamps are treated as "cooled down", and the cooldown boundary is now inclusive (`>=`). Added two regression tests.
- **Barrel export (Greptile P2, `index.ts`)** — `recordExportAndShouldPromptSupport` is now exported from the `ExportDialog` barrel, and both wrappers import it from there instead of the internal path.

**Intentional — keeping as-is**
- **Hardcoded `#fff` on the Ko-fi buttons (Copilot, header + export prompt)** — this is a deliberate product decision: the ask is styled to match the official Ko-fi widget, which always uses light text on the brand color. Using `--text-on-accent` would render *black* text on the light accents (e.g. the default amber), which isn't the intended look. A text-shadow is already applied to preserve legibility. So the theming bypass here is by design, not an oversight.

PR #2055, opened 2026-06-07
feat(baseplate): snap-together clip connectors (#1610)

### LINKED ISSUE #1610: Snap together baseplates
In addition to the dovetail connectors between baseplate seams I would love the option to have snap together baseplates like the ones featured here.

https://www.printables.com/model/430144-gridfinity-base-with-snap-connectors

andymai:
Closes #1610.

## What

Adds a third baseplate **connector style — "Snap clip"** — alongside Dovetail and Dovetail key. It's a separate top-insert **"staple"** that clips two adjacent baseplate tiles together: press it straight down into the seam from the top until it clicks.

Each seam side gets a **blind, ledged pocket**; the printed clip has two legs that **pinch inward** past the pocket throat on insertion, then **splay out** so the outward-facing **barbs catch the chamber ledge** — a positive snap lock, removable with deliberate force. The bridge sits **flush** with the plate top, so nothing protrudes where bins sit.

This was designed interactively with the issue author's "clips in from the top like a staple" framing.

## How it was built (geometry-first)

The snap geometry is the novel/risky part, so it was **proven standalone with `brepjs-verify`** against the real kernel *before* any app integration:

| Check | Result |
|---|---|
| Clip is a valid manifold solid | ✅ 60.45 mm³, flush top, correct bounds |
| Pockets (throat → chamber → ledge) cut clean | ✅ valid, watertight |
| **Seated clip clears the pockets** | ✅ fused volume == tiles + clip, **interference = 0.000 mm³** |

A single shared **`snapClipLevels(totalHeight, fitOffset)`** (`@/shared/constants/connectors`) drives the pocket cut, the standalone clip, *and* the seated-clip preview, so they can't drift. The barb is a fixed-size feature near the leg tip, so **only the leg length scales with slab height** — taller bases (magnet/weighted) get a longer flex beam automatically, and a too-thin slab skips the pockets rather than ship a clip that would snap off.

## Changes

- **Data model:** new `connectorStyle: 'snapClip'` across `core/types`, `shared/types/bin`, the CQRS Zod schema, and the constants normalizer.
- **Worker geometry:** `buildSnapClip` / `buildSnapClipForPrint` (bed-flat) + blind ledged pockets in `baseplateConnectors.ts`, wired into the generator export path, cache keys, split-planner bed budget, and print guide.
- **UI:** "Snap clip" in the connector-style selector; reuses the existing connector **fit-offset slider** for printer tuning (i18n across all 10 locales + values allowlist).
- **Preview:** seated staple markers at each seam junction (`ConnectorKeyMeshes`).

## Tests

- **Real-kernel scenario tests** (`baseplateGenerator.scenario.snapClip`): snap-clip baseplate exports a **watertight STL** (0 non-manifold / 0 boundary edges), the connector key is a valid **bed-flat** clip, and the clip volume **matches the brepjs-verify-proven design**.
- Existing dovetail / dovetail-key suites still green (456 baseplate+generation tests pass).
- Playwright spot-check: the snap-clip style selects and renders the seated markers at the seam junctions with zero console errors.

## Notes / follow-ups

- The connector toggle is still labeled "Dovetail connectors" (pre-existing); now that it gates three styles, renaming it to "Connectors" would be a small follow-up.
- The seated-clip **preview** is a draft accent marker; the exported/generated geometry is the rigorously-proven path.


andymai:
Addressed the review feedback in 7585b4c8:

- **`generateConnectorKeySection` (the main flag):** now branches for `snapClip` — emits `snap_clip`, "press one in from the top until it clicks" (not "Hammer"), and a fit note derived from `SNAP_CLIP_CLEARANCE` (0.20 mm total) instead of `DOVETAIL_KEY_CLEARANCE`. Added a print-guide test asserting the snap-clip copy and clearance.
- **WASM leaks:** `buildSnapClip` and `buildSnapClipForPrint` now `.delete()` their intermediate B-rep shapes after `translate`/`rotate` (matching the `orientForPrint` discipline elsewhere).
- **`countConnectorKeys` JSDoc:** updated — it now counts `snapClip` junctions too.

**Deliberately not changed:** the `splitPlanner` `TONGUE_PROTRUSION` bed reservation for female-only styles. It's pre-existing (already applies to `dovetailKey`) and only *over*-reserves bed budget — never under-reserves — so it's safe/conservative. Fixing it correctly means threading `connectorStyle` into `makeAxisConfig`, which would shift piece counts for existing `dovetailKey` layouts; that belongs in its own change rather than riding along here.

PR #2056, opened 2026-06-08
fix(generation): footprint-aware timeout budget + relax large-bin lip tessellation

andymai:
## Problem

Large-footprint hex-wall bins time out in the **live preview** (reported with "hex walls + removable dividers"). Removable dividers (`slotted` style) were a red herring — slots *disable* the hex pattern on slotted walls, so they make hex bins faster, not slower.

Root cause, established by reproduction + per-stage instrumentation:

- The hot path is the **boolean `pattern_cut`** (subtracting hundreds of hex prisms from the bin) — ~16s of a ~26s generation on a 16×16×6 bin. Its cost scales with wall **area**.
- `computeGenerationTimeoutMs` budgeted for hex, cutouts, and height — **never footprint**. A wide bin got only 45–60s while legitimately needing more; the browser's slower single-threaded WASM pushed it past the budget → timeout.

> Note: the pipeline runner emits progress *before* each stage, so the `merge@0.8` spike is actually the preceding `boolean` stage — tessellation is only ~1.5s. Easy to mis-attribute.

## Changes

1. **`generationTimeout.ts`** — footprint-aware budget: `+250ms` per grid cell above a 4×4 floor, gated on hex (a plain large bin has no pattern cut and finishes fast). Large hex bins now reach the 90s cap; normal bins unchanged.
2. **`tolerances.ts`** — the `hasLip` branch clamped tessellation at `0.06mm` for *any* size, over-meshing large lipped bins at near-export fidelity. Now relaxes with size (cap `0.15mm`, angular `8→12` above 200mm), cutting preview mesh weight ~29% on large bins. This lightens preview memory/transfer/GPU; it does **not** change generation time (the boolean dominates) — the budget fix is what resolves the timeout.

## Testing

- New `tolerances.test.ts` + extended `generationTimeout.test.ts` (TDD, red→green).
- Full bin scenario suite passes in isolation; one deliberate snapshot updated (`8×8 standard`, 22924→22308, validated). Export (STL) fidelity is unchanged.

### DIFF
--- a/src/features/generation/bridge/generationTimeout.test.ts
+++ b/src/features/generation/bridge/generationTimeout.test.ts
@@ -10,11 +10,15 @@ import {
   BASEPLATE_MAX_TIMEOUT_MS,
   HEX_PATTERN_BONUS_MS,
   HEX_PLUS_CUTOUT_BONUS_MS,
+  HEX_FOOTPRINT_BONUS_MS_PER_CELL,
+  HEX_FOOTPRINT_BONUS_FLOOR_CELLS,
   HEIGHT_BONUS_MS,
   MAX_TIMEOUT_MS,
   computeBaseplateTimeoutMs,
   computeGenerationTimeoutMs,
 } from './generationTimeout';
+
+const HEX_ON = { enabled: true, pattern: 'honeycomb' } as const;
 
 function params(overrides: Partial<BinParams> = {}): BinParams {
   return { ...DEFAULT_BIN_PARAMS, ...overrides };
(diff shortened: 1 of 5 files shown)


andymai:
Addressed review feedback in f6f7cdf5:
- **Non-finite dims guard** (generationTimeout.ts): floor NaN/negative `width`/`depth`/`height` before the footprint+height math and clamp the result to `[BASE, MAX]`, mirroring `computeBaseplateTimeoutMs`. Prevents `setTimeout(NaN)` → 0ms immediate cancel during mid-edit UI state. Added a regression test.
- **Comment wording** (tolerances.test.ts): reworded to clarify the relaxed lip tessellation lightens the *preview mesh weight* and does not drive the generation timeout (the hex boolean cut does).

PR #2059, opened 2026-06-08
feat(baseplate): connector fit-sample test print

andymai:
## What

Adds a **Print fit sample** button to the baseplate connector section. It exports a single calibration tray — a "fit card" — so makers can dial in their connector tolerance *before* committing to a full split baseplate print.

The tray is a grid:

| | −0.10 | −0.05 | 0.00 | +0.05 | +0.10 |
|---|---|---|---|---|---|
| **Dovetail** | pair | pair | pair | pair | pair |
| **Dovetail key** | pair | pair | pair | pair | pair | + 1 key |
| **Snap clip** | pair | pair | pair | pair | pair | + 1 clip |

- **Rows** = the three connector styles; **columns** = a fit-offset ladder centered on nominal.
- Each cell is a **mated pair** of small abstract coupons carrying the *real* connector profile at the plate's height. Both halves are **embossed** with their style (`DT`/`DK`/`SC`) and offset so detached pieces stay identifiable.
- The dovetail **key** and snap **clip** are nominal parts (the offset rides on the female pocket), so those rows ship **one shared loose part** the maker inserts into each offset's pair — rather than five redundant copies.

Workflow: print it → find the fit that clicks → set **Connector fit** to that offset → print the real baseplate.

## Why

Connector fit is printer/filament-dependent, which is exactly what the existing `connectorFitOffset` control is for — but today there's no cheap way to calibrate it short of reprinting a whole multi-piece baseplate. This makes calibration a ~10g test print.

## How

- New `connectorSample.ts` worker generator builds the tray. Coupon profiles **reuse** the existing `makeTongue` / `makeGroove` / `makeSnapPocket` builders (now exported) so the printed fit is faithful — the tolerance lives in the female groove/pocket, not the surrounding socket, so a plain block coupon fits identically to a real 42 mm cell.
- New `EXPORT_CONNECTOR_SAMPLE` worker path mirrors the existing connector-key export; reuses the `BASEPLATE_EXPORT_RESULT` response.
- UI: button in the connector section (shown when connectors are enabled) → the shared `ExportDialog` (STL / STEP / 3MF) with inline print tips.
- Labels via the existing `buildTextSolid` emboss path; pieces compound into one ready-to-slice, bed-resting file.

## Tests

- `connectorSample.scenario.test.ts` (real OCCT kernel): 32 valid positive-volume pieces, watertight + bed-resting STL (0 boundary / non-manifold edges), and magnet-hole (taller) coupons stay valid.
- Sibling tests for the hook and the dialog button.
- i18n keys added across all 10 locales; `check:i18n` + `check:i18n:values` pass.

## Notes

- Connector controls (and this button) only appear when the baseplate is split, matching the existing connector UI gating.

### DIFF
--- a/src/features/baseplate/components/BaseplatePanel/BaseplatePanel.tsx
+++ b/src/features/baseplate/components/BaseplatePanel/BaseplatePanel.tsx
@@ -34,6 +34,7 @@ import { PaddingSchematic } from './PaddingSchematic';
 import { SplitViewStrip } from './SplitViewStrip';
 import { CornerRadiusControl } from './CornerRadiusControl';
 import { GridDimensionStepper } from './GridDimensionStepper';
+import { ConnectorSampleButton } from './ConnectorSampleButton';
 import { resolveOverTileStatus } from '../../utils/overTileStatus';
 import { PADDING_MAX } from '../PaddingStepper';
 import { Stepper } from '@/design-system/Stepper';
@@ -371,6 +372,7 @@ export function BaseplatePanel() {
                       }
                       label={t('baseplate.preferIdenticalPieces')}
                     />
+                    <ConnectorSampleButton />
                   </>
                 )}
               </>
--- a/src/features/baseplate/components/BaseplatePanel/ConnectorSampleButton.test.tsx
+++ b/src/features/baseplate/components/BaseplatePanel/ConnectorSampleButton.test.tsx
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { ConnectorSampleButton } from './ConnectorSampleButton';
+import { resetAllStores } from '@/test/testUtils';
 
 vi.mock('@/i18n', () => ({
   useTranslation:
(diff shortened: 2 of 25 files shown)


andymai:
Thanks for the review — addressed in the latest commits:

**Custom filename ignored (Greptile #1–3, Copilot)** — real bug, fixed. `downloadSample(format, baseName?)` now threads the resolved base name from the dialog through to `triggerDownload` (and the 3MF object name), so the saved file matches what the dialog shows. Added regression tests for both the default and a typed custom name.

**Duplicated constants (Greptile P2)** — `FORMAT_EXTENSIONS` consolidated into `@/shared/generation/exportUtils`; `SAMPLE_BASE_NAME` is now a single exported `CONNECTOR_SAMPLE_BASE_NAME` from the hook, imported by the component (removes the two-copies drift that enabled the filename bug).

**Missing `resetAllStores()` (Greptile P2)** — added to `beforeEach` in both new test files.

Also from a self-review pass before the bots: added `sanitizeParams` to the sample exporter (matches every sibling exporter; guards against NaN coupon height), made the tray export free its pieces if `compound`/the bed-lift throws, and added a STEP-branch scenario assertion + bridge-backed STL/3MF hook tests.

Note on the `format === '3mf'` "fall-through" in `connectorSample.ts`: the worker `ExportFormat` is `'stl' | 'step'` only — 3MF is synthesized client-side in the hook (STL→3MF), so the generator can't be reached with `'3mf'` at the type level.

PR #2065, opened 2026-06-08
fix(print-export): correct filament & print-time estimates (were too high)

andymai:
## Problem

Filament and print-time estimates over-reported — badly in the **bin designer / Export dialog** (the surface users actually see for a single custom bin).

## Root-cause method

Rather than trust the in-code "PrusaSlicer calibration" data, I measured the **real generated geometry** with OCCT (`measureVolume(getLastSolid())`). That data turned out to be fabricated — its docstring claimed a least-squares fit with "max error ~1 min," but the shipped constants missed by 10×, and a 1×1×3u bin is actually **6241 mm³ (≈2.59m)** solid, not the 2.3m the old data claimed.

## Bugs found & fixed

| # | Bug | Impact |
|---|-----|--------|
| 1 | Bin-designer estimator treated the bottom **7mm of every bin as a solid slab** (cavity carved only above the base) | **3–6× too high**, worsening with footprint |
| 2 | Print-export model judged against fabricated targets; vs real geometry it **under**-counted tall bins | −17% on 2×2×9u |
| 3 | Print time charged 16-min plate overhead **per distinct bin type** (`rows.length`) | hours of phantom overhead on varied layouts |
| 4 | **Nozzle size changed filament volume** | wrong — the CAD wall is fixed; nozzle affects slicing/time only |

## Changes

- **Rewrite `estimateStandardBinVolume`** as a 3-term model fit to OCCT ground truth (`perimeter·WALL_EFF·height + base/cell + lip`). Reproduces real solid volume within **~2%** across 1×1×2u–4×4×6u. Now nozzle-independent.
- **Consolidate** the bin-designer estimator onto the shared base geometry via the new `standardBinSolidComponents`, layering feature deltas (dividers, labels, scoops, honeycomb) on top; deleted its divergent hollow-box/socket/lip math.
- **`estimatePrintJobs`** — estimate plate count from bed area; use it for overhead instead of `rows.length`.
- **Time model**: derive `MINUTES_PER_METER` from an effective volumetric flow (~11 mm³/s) and document it.
- **Replace fabricated calibration data/tests** with documented OCCT ground truth and a **±10% regression test**.

## Before / after (default 2×2×3u standard bin)

| | Volume | Filament | Mass |
|---|---|---|---|
| Before (designer) | ~67,000 mm³ | ~27.8 m | ~83 g |
| After | ~17,100 mm³ | ~7.1 m | ~21 g |
| **OCCT actual** | **17,094 mm³** | — | — |

## Tests

- New ±10% regression test pins the model to 10 OCCT-measured bins.
- Full suite green: **11,493 passed**; typecheck, lint, knip clean.

### DIFF
--- a/src/features/bin-designer/utils/printEstimates.test.ts
+++ b/src/features/bin-designer/utils/printEstimates.test.ts
@@ -6,7 +6,7 @@ import {
   formatFilament,
 } from '@/features/bin-designer/utils/printEstimates';
 import { DEFAULT_BIN_PARAMS } from '@/features/bin-designer/constants/defaults';
-import { DEFAULT_PRINT_SETTINGS } from '@/shared/printSettings';
+import { DEFAULT_PRINT_SETTINGS, estimateStandardBinVolume } from '@/shared/printSettings';
 import type { BinParams } from '@/features/bin-designer/types';
 
 describe('printEstimates', () => {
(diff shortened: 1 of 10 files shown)


andymai:
Addressed the review feedback in 67dda6a:

- **`estimatePrintJobs` on empty layouts** — now returns `0` when total footprint is `0`, so a cleared/fully-filtered print list no longer accrues a phantom plate of overhead (previously showed 0.3h for nothing). Still clamps to ≥1 for any non-empty layout, and to 1 for an unknown (zero-capacity) bed.
- **Tautological test assertion** — replaced the self-equality check with an explicit empty-layout case (`estimatePrintJobs(0, 36) === 0`) alongside the existing expected-count assertion.
- **Stale doc** — fixed the solid-volume formula comment to reference `BASE_VOL_PER_CELL_AREA · cellArea · cells` (the constant was renamed during the lint-rule fix).

Issue #1676, opened 2026-05-16
Bin export failed: Error


qzte:
**Format:** STL
**Error:** Split export failed: Split piece A2 lost geometry: expected body Z≈49.0mm (lip fused separately), got 26.7mm. This is likely caused by a coplanar cut plane — please report this bug.

**Bin params:**
```json
{
  "width": 3,
  "depth": 10,
  "height": 7,
  "gridUnitMm": 42,
  "heightUnitMm": 7,
  "style": "standard",
  "base": {
    "style": "standard",
    "stackingLip": false
  },
  "wallPattern": null,
  "handles": false,
  "cutouts": 0,
  "inserts": 0
}
```

andymai:
@qzte Looking. Thx

PR #2069, opened 2026-06-08
perf(split): cut only each worker's assigned pieces

andymai:
## What

Split preview and STL export are now **~2× faster** on multi-piece grids, with byte-identical output.

| Scenario | Before | After | Speedup |
| --- | --- | --- | --- |
| 16-piece grid | 22.0s | 9.7s | **2.26×** |
| 8-piece grid | 7.6s | 3.9s | **1.95×** |
| 2-piece split | 3.4s | 2.6s | **1.31×** |

## Why it was slow

The worker pool distributes split pieces round-robin, but each worker's `splitSolidIntoPieces()` regenerated the full bin **and boolean-cut every piece**, then tessellated only its assigned subset. So the cut work — the dominant cost (~1s per piece vs ~40ms to tessellate) — was O(pieces × workers) and got *worse* as more workers were added; only tessellation actually parallelized.

Solids can't cross the worker boundary (the kernel's shape handles are heap-local), which is why each worker regenerates rather than cutting once and shipping pieces out.

## The fix

- Thread `pieceIndices` into `splitSolidIntoPieces` so each worker only `intersect()`s the pieces it will tessellate. The skip is gated on the col-major flat index (`col*numRows+row`) so it still lines up with the pool's round-robin distribution.
- Cap the pool to `min(pieceCount, poolSize)` so a small split doesn't spin up workers that each pay a redundant full-bin regeneration.

Output is geometry-identical — a new parity test asserts labels, triangle counts, and bounding boxes match the full split exactly.

## Bonus: clearer error for grid-split of a hollow bin

A fully-interior grid piece (interior cut on all four sides) of a **hollow** bin genuinely has no walls — it's a thin floor tile over the open cavity — so the geometry-loss guard correctly rejects it. The old message blamed *"a coplanar cut plane — please report this bug,"* sending users to chase a non-existent kernel bug. It now detects the fully-interior case and explains the cavity, suggesting splitting into strips or using a solid bin. (Solid bins split into full grids fine.)

## Tests

- `binGenerator.scenario.split-range.test.ts` — range/full parity, out-of-range rejection, cavity messaging, solid-bin full-grid success
- `splitPerfBreakdown.test.ts` — profiling harness (excluded from CI, mirrors the existing `previewPerfBreakdown`) with real before/after pool timing
- Existing split scenario suite, WorkerPool tests, typecheck, lint — all green

### DIFF
--- a/src/features/generation/bridge/WorkerPool.ts
+++ b/src/features/generation/bridge/WorkerPool.ts
@@ -122,7 +122,11 @@ export class WorkerPool {
   ): Promise<SplitPreviewResult> {
     await this.ensureWorkers();
 
-    const groups = distributeRoundRobin(totalPieceCount, this.bridges.length);
+    // Cap active workers to the piece count: each worker regenerates the full
+    // solid, so spinning up more workers than pieces only adds redundant
+    // full-gen work with no piece to show for it.
+    const activeWorkers = Math.min(totalPieceCount, this.bridges.length);
+    const groups = distributeRoundRobin(totalPieceCount, activeWorkers);
     let completed = 0;
 
     const taskGroups = groups.map((pieceIndices, bridgeIdx) => {
@@ -174,7 +178,9 @@ export class WorkerPool {
   ): Promise<SplitExportResult> {
     await this.ensureWorkers();
 
-    const groups = distributeRoundRobin(totalPieceCount, this.bridges.length);
+    // Cap active workers to the piece count (see generateSplitPreview).
+    const activeWorkers = Math.min(totalPieceCount, this.bridges.length);
+    const groups = distributeRoundRobin(totalPieceCount, activeWorkers);
     let completed = 0;
 
     const taskGroups = groups.map((pieceIndices, bridgeIdx) => {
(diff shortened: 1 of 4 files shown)


andymai:
Addressed the review feedback in 39c98bf:

- **WASM leak (Copilot):** `tessellatePiece` now disposes each piece solid after tessellation. The preview path was leaking the per-piece boolean solids across repeated generations; the export path already disposed via `tessellateAndExportPiece`.
- **Cavity message on solid bins (Copilot):** the "open cavity" diagnosis now only fires for hollow bins (`!params.base.solid`). A solid bin has no cavity, so an interior geometry loss there correctly falls through to the coplanar "please report this bug" path.
- **Test plane nudge (Greptile):** the parity/messaging `cutPlanes` helper now nudges planes `0.27 * GRID` off socket-cell boundaries, matching the perf harness, so the test doesn't depend on OCCT tolerating coplanar planes (#1676).

Typecheck, lint, and the split scenario suite (range parity, messaging, connectors, wall-locking) all green.

PR #2071, opened 2026-06-08
perf(split): render a Manifold draft split on the leading edge

andymai:
## What

Splitting an oversized bin now shows its pieces **almost immediately** instead of waiting seconds for the full exact pipeline. A fast Manifold **draft** split renders on the leading edge of each edit, and the exact occt-wasm result supersedes it when ready.

Builds on #2069 (which made the exact split ~2× faster by removing the per-worker cut redundancy). This PR attacks the *remaining* cost — the split preview was exact-only and never used the Manifold draft path the single-bin preview already has.

## Why it was slow

`useSplitPreview` waited for the main bin generation to finish, then ran the full occt-wasm split (full-bin regen + per-piece boolean cuts) before any pieces appeared. The single-bin designer preview already renders a Manifold draft while occt computes — the split path was simply never wired to it.

## Measured

Running the split path under the Manifold kernel end-to-end (valid geometry, no NaN):

| Case | Manifold draft | occt-wasm exact | Speedup |
| --- | --- | --- | --- |
| socket+lip 8×2 | **334ms** | 2425ms | **7.3×** |
| flat no-lip 12×2 | **12ms** | 84ms | **7×** |

## How

- `useSplitPreview.ts`: fire `previewBridge.generateSplitPreview` (Manifold) immediately on each edit; the exact occt split (pool or single bridge) still runs after main generation completes and supersedes the draft.
- Token arbitration mirrors the single-bin draft→exact path (`genToken`/`finalizedToken`): a draft is dropped if a newer edit started or the exact result for its edit already landed; the exact result finalizes its token so a slow draft can't clobber it.
- A `previewReady` flag re-fires the draft once the Manifold bridge finishes loading (it usually resolves after the first render), so the initial edit doesn't miss its draft.
- Exposes `SplitPreviewResult` on the `@/shared/generation/bridge` facade.

## Tests

- `useSplitPreview.test.ts` — 15 tests: the 12 existing exact-path tests plus draft-renders-first, exact-supersedes-draft, and stale-draft-dropped.
- `split-draft-manifold-visual.spec.ts` — one-shot Playwright visual (oversized bin renders a split preview that updates on edit).

## ⚠️ Visual verification needed

I could **not** render the 3D preview in my sandbox — occt-wasm + WebGL don't initialize headless there (the preview hangs at "Generating mesh…", no canvas). The kernel-level perf and the hook→store integration are verified, but the actual pixel render is not. **Please check the Vercel preview** (oversize a bin so it splits, then edit it) to confirm the draft renders and feels fast before merge. Not arming auto-merge.

### DIFF
--- a/e2e/bin-designer/split-draft-manifold-visual.spec.ts
+++ b/e2e/bin-designer/split-draft-manifold-visual.spec.ts
@@ -1,0 +1,59 @@
+/**
+ * One-shot visual + perf verification for the Manifold draft split.
+ *
+ * Drives the designer to an oversized (splitting) bin and confirms the split
+ * preview renders, then edits the bin and confirms the preview updates — the
+ * draft (Manifold) path renders on the leading edge while the exact occt-wasm
+ * result computes. Captures screenshots for visual review.
+ *
+ * Run with: `pnpm test:e2e e2e/bin-designer/split-draft-manifold-visual.spec.ts`
+ */
+
+import type { Page } from '@playwright/test';
+import { test, expect } from '../fixtures';
+
+test.use({
+  launchOptions: {
+    args: ['--use-gl=swiftshader', '--enable-webgl', '--ignore-gpu-blocklist'],
+  },
+});
+
+async function waitForGenerationComplete(page: Page): Promise<void> {
+  await expect(page.getByRole('status', { name: /generating mesh/i })).toHaveCount(0, {
+    timeout: 30_000,
+  });
+}
+
+test.describe('Manifold draft split — visual', () => {
+  test('oversized bin renders a split preview that updates on edit', async ({ page }) => {
+    test.setTimeout(180_000);
+    await page.goto('/designer');
+
+    const canvas = page.locator('canvas').first();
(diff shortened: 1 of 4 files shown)


andymai:
Addressed the review feedback in ef1d9e2b:

- **Token arbitration / late bridge-acquire (Greptile + Copilot):** the per-edit token is now bumped only on param/maxGrid edits — never by draft dispatch or `previewReady`. Both the draft and exact paths read the same edit token, and the draft is skipped (and dropped on resolve) when `token <= finalizedToken`. So a Manifold bridge that loads *after* the exact result already landed can no longer mint a newer token and downgrade the exact preview to a draft. Added a regression test (`does NOT downgrade the exact result when the draft bridge loads late`).
- **`maxGrid` stability (Copilot):** memoized on its primitive inputs so the effects don't re-run / re-dispatch split work on every render.
- **e2e screenshots (Copilot):** switched to in-memory buffers + `test.info().attach()` instead of hardcoded `/tmp` paths.

16 hook tests pass (incl. the new late-acquire regression), typecheck + lint green.

PR #2080, opened 2026-06-10
feat(connectors): make split/baseplate connectors print on wider nozzles

andymai:
## What & why

Split-bin connectors, wall locks, and baseplate connectors didn't work when printed with a **0.6mm nozzle**. The features were all hand-tuned for a 0.4mm nozzle, so on a wider nozzle:

- connector/lock **walls fell below two perimeters** (structurally weak / skipped),
- the snap-clip **barb (0.45mm) was narrower than a single bead**, so the slicer dropped it entirely,
- tight **press-fit clearances seized** because the fatter extrusion overshoots into the groove.

This scales every connector dimension and fit clearance with the printer's nozzle size so all four features stay printable on 0.6/0.8mm nozzles. **Geometry is byte-identical at the 0.4mm baseline** — existing designs and fits are unaffected.

## Covered features

- Split-bin **wall keys** (wall locks) and the **floor scarf**
- Baseplate **dovetail keys** and **snap clips**

## How it works

- New shared single-source helper `connectorScaling.ts` (`scaleFeature` / `scaleClearance` / `isFeaturePrintable`): values are pinned to their legacy 0.4mm tuning at or below the baseline, and only grow above it (2-bead feature floor + bead-grown clearance). Because nozzle size is a discrete setting, there's no continuity to lose — 0.4mm output is unchanged.
- The worker split + baseplate builders consume the shared scalers via a `nozzleSizeMm` threaded through `SplitConnectorConfig` / `BaseplateParams`.
- The baseplate mesh cache key includes the nozzle so wider-nozzle geometry can't alias the cached 0.4mm build.
- The shared `nozzleSizeMm` (a printer property, kept in settings) is stamped onto the worker payload at every preview/export dispatch — **not persisted per-design**.

## UI

Nozzle size is now editable alongside grid unit / print bed on the **bin-designer Physical units**, **baseplate dimensions**, and **mobile settings** surfaces — one shared value, editing anywhere updates everywhere. An inline advisory appears in the split-options and baseplate-connector panels when the nozzle is above 0.4mm, noting connectors are enlarged to stay printable and very small pieces may omit them. Localized across all 9 locales.

## Verification

- **Unit:** scaling math + builder consumption proven (0.4mm identical; 0.6mm diverges; barb grows from sub-bead to a full bead; wall key ≥2 perimeters).
- **Kernel (real OCCT):** new scenario cases confirm the scaled 0.6mm wall-lock and snap-clip geometry build **watertight, full-height, and are not silently dropped**, and the snap clip's volume grows as the barb enlarges.
- typecheck, lint, and i18n all green. **136 connector/UI tests pass.**

A CLI slice check at 0.6mm wasn't runnable headless in this environment (GUI slicer needs a display); the geometric guarantee (every feature ≥ 2× nozzle) plus the watertight kernel exports cover correctness — final confirmation is a physical 0.6mm test print.


andymai:
Thanks for the review — all four findings addressed:

**Copilot** — baseplate not keyed on nozzle size (`a6ae759`): the regenerate effect now subscribes to `nozzleSizeMm` reactively and lists it in the dependency array, so changing the nozzle re-runs baseplate generation immediately.

**Greptile P2s** (`9304fcd`):
- `baseplateCacheKeys.ts` — use the `NOZZLE_BASELINE` constant for the fallback instead of a hardcoded `0.4`, so the cache key can't drift from the geometry math.
- `MobileSettingsPanel.tsx` — wrapped `setNozzleSizeMm` in `useCallback` for parity with the other handlers.
- `usePhysicalUnitsSection.ts` — `handleNozzleChange` now reads `printSettings` via `getState()` (dep only on `updateSetting`), so it no longer recreates on unrelated print-setting changes, matching the `BaseplatePanel` pattern.

typecheck, lint, and the affected unit tests are green.

PR #2082, opened 2026-06-10
feat(rightpanel): add nozzle size input to the desktop print list

andymai:
## What

Adds the nozzle-size input to the **desktop RightPanel** — the one surface left out of #2080. It lives in the **Bin List** section, between the piece table and the print-time/cost summary it affects.

## Why

#2080 made connectors, wall locks, and baseplate connectors scale with `nozzleSizeMm` and exposed that setting on the bin-designer, baseplate, and mobile dimensions surfaces. The desktop RightPanel was intentionally deferred; this closes that gap so the nozzle is editable from the main layout view too.

## Details

- Bound to the same shared `settings.printSettings.nozzleSizeMm` (editing anywhere updates everywhere — it's a printer property, not per-design).
- `usePrintList` already reads `printSettings` reactively, so editing the nozzle live-updates the print-time estimate; it also feeds the nozzle-aware connector scaling.
- Same `SettingsRow` + `DeferredNumberInput` pattern (0.2–1.0mm, 0.2 step) as the other surfaces.
- i18n tooltip across all 10 locales.

## Verification

- typecheck, lint, i18n all green; 46 RightPanel tests pass.
- Playwright visual check: added a bin, confirmed the "Nozzle size" input renders in the Bin List section (value 0.4mm) between the table and the print summary.

### DIFF
--- a/src/i18n/locales/de.json
+++ b/src/i18n/locales/de.json
@@ -1725,6 +1725,7 @@
   "printBedInput.widthAriaLabel": "Druckbett-Breite",
   "rightPanel.binList": "Bin-Liste",
   "rightPanel.binProperties": "Bin-Eigenschaften",
+  "rightPanel.nozzleSizeTooltip": "Düsendurchmesser deines Druckers. Bestimmt die Druckzeitschätzung und skaliert Verbinder geteilter Bins, Wandsperren und Grundplatten-Verbinder, damit sie auf breiteren Düsen druckbar bleiben.",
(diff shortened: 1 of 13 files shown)


andymai:
Addressed the Copilot review (`4cf0411`): added focused RightPanel tests asserting the nozzle input renders when the print list has rows and that changing it updates `settings.printSettings.nozzleSizeMm`. RightPanel suite now 48 tests, all green.

PR #2101, opened 2026-06-11
feat(bin-designer): show compartment dimensions in mm and tile by size

andymai:
## What & why

A user asked to see interior compartment sizes in millimeters and to lay out compartments by specifying a target size rather than a count. This adds both, for the compartment-grid (divider) interior style.

### See compartment dimensions in mm
- **3D dimension lines** for the hovered/selected compartment, mirroring the bin's own exterior dimension lines. They appear on demand (hover/select a cell in the grid editor) and stay clear at rest — no clutter on dense grids.
- **2D readout**: each compartment cell shows its cavity size (W×D) on hover, and a summary line under the grid follows the hovered compartment (`W × D × H mm`).
- All measurements are the **usable cavity** — divider walls subtracted — so the number reflects what an object actually has to fit into.

### Tile by size
- A **By count ↔ By size** toggle on the grid controls. *By size* lets you type a target compartment width/depth in mm; the grid picks the count whose resulting cavity is closest and shows the **actual achieved** size. Height proxies the existing divider-height control.

## Implementation notes
- New `utils/compartmentDimensions.ts`: cavity math (single-cell + multi-cell span) and a nearest-target count solver, with colocated tests.
- New `preview/CompartmentDimensions` R3F component, mirroring `BinDimensions`; mounted in `PreviewCanvas` (gated like the exterior lines).
- `hoveredCompartmentId` added to the designer UI store to bridge the 2D editor hover and the 3D preview.
- The derived divider-height value is memoized (like `DividerHeightControl`) — calling the slotMath helpers inline in the render body tripped React Compiler's manual-memoization preservation.

## Testing
- Full unit suite green; new util + component tests added.
- Typecheck, lint, i18n (keys + interpolation + values) all pass; translations added across all locales.
- Playwright visual check: confirmed the 3D dimension lines, the hovered-compartment highlight, the 2D readout, and the By-size inputs render correctly.

Scope: compartment-grid style only (the slotted style already works in mm).


andymai:
Thanks for the review — addressed in b699dbffd:

**Copilot — cavity math didn't match the generator** (`compartmentDimensions.ts`, `CompartmentDimensions.tsx`): Correct, good catch. Reworked `compartmentCavity` to follow `cavityCorners` exactly — equal pitch (`innerW/cols`) with a `thickness/2` inset on *interior* edges only, so edge compartments are wider than interior ones and aren't centered in their cell. It now returns true centered extents (`xMin/xMax/yMin/yMax`), and the 3D dimension lines use those directly instead of centering the cavity in the pitch footprint. Added a unit test asserting edge (24.4 mm) vs interior (23.8 mm) cells for a 4-col / 100 mm / 1.2 mm grid. `singleCellCavity` is kept only as the *average* cavity for the by-size solver and the uniform "≈" summary (documented as such); `spanCavity` removed.

**Copilot — `cells[hoverIdx]` out of bounds on resize** (both sites): Guarded with `hoverIdx < cells.length`.

**Greptile P2 — `formatMm` duplicates `formatCompactMm`**: Done, `formatMm` now wraps it.

**Greptile P2 — `resetAllStores()` in the new test**: Added to `beforeEach`.

Also bumped the Total JS size-limit 1755→1760 kB to cover the added feature (CI was 521 B over).

Issue #1949, opened 2026-05-31
Incorrect object split when overhang is set and model is too large for print bed


JonathanHuss:
**Describe the bug**
When creating a bin that has an overhang, if the object also needs to be split to accommodate the print bed size, in addition to splitting the object to fit on the bed, it also appears to split (and delete) the overhang. When the bed accommodation split does not need to occur, no issues.

**To Reproduce**

1. Create a bin that's 1x7
2. Add an 18mm overhang on one side
3. Set the print bed size to 254x254 (in my case, using a Bambu Labs A1)
4. Wait. The object initially renders appropriately, but after a few seconds (presumably following split logic) the overhang is cut from the object.
5. For reference: https://gridfinitylayouttool.com/designer?id=design_1780198487896_f4zspc

**Expected behavior**
For the overhang to remain intact.

**Screenshots**

Immediately following rendering, before the split:

<img width="1612" height="1044" alt="Image" src="https://github.com/user-attachments/assets/3cd5ed00-56e2-4e90-93f0-b9d99e5ab053" />

After the split:

<img width="1219" height="767" alt="Image" src="https://github.com/user-attachments/assets/7f25627e-0ab2-498a-bbf2-7cbb5b47ce1b" />

**Environment:**

- Browser: Edge Version 148.0.3967.96
- Device: Desktop - Windows 11 Enterprise, 26200.8457

JonathanHuss:
Confirming that the issue is resolved - that was fast! Thank you!

PR #2105, opened 2026-06-11
fix(bin-designer): keep stacking lip on split custom-shape and overhung bins

andymai:
## Problem

Splitting a bin that has a **stacking lip** lost or mangled the lip on the pieces whenever the bin had a **custom shape (L/U)** or an **overhang** — a visibly missing/broken lip in the 3D preview and exploded view.

## Root cause

The split builder can't reuse the normal lip-build path: it builds the body and lip **separately** and fuses them per-piece, to avoid an OCCT boolean-intersect crash at the lip-wall junction. That separate lip build had drifted from the canonical `shellStage` path — it called `buildTopShape(width, depth, true, gridUnitMm)` and dropped two footprint arguments `shellStage` passes:

- **`cellMask`** → the lip was built as a full rectangle, so it juts over the cleared notch of an L/U shape.
- **`overhang`** (#1949) → the lip stayed at nominal grid size while the body grew outward, so it falls short of the overhung edge.

## Fix

Pass `params.cellMask` and the resolved `overhang` into the split lip's `buildTopShape` call, mirroring `shellStage`. `buildTopShape` already suppresses overhang for polygon masks, so rectangular bins are byte-unchanged. The remaining diff just hoists the existing `overhang` resolution above the lip block.

## Tests

New `binGenerator.scenario.split-lip-shape-overhang.test.ts` (reproduce-first — both assertions fail before the fix, pass after):

- **L-shape notch:** the right split piece's concave notch corner must stay empty (a rectangular lip left ~99 stray vertices there).
- **Overhang:** each piece's lip-zone X-extent must reach the body's X-extent (pre-fix 209.7 vs body 227.7 on the overhung piece).

Assertions are translation-invariant because `tessellatePiece` re-centers each piece at its own bbox center.

All existing split/lip scenario tests pass (lip-trim, overhang, polygon, slotted-lip, wall-cutout-lip, cutouts, export, manifold-kernel, cell-boundary); `tsc` + ESLint clean.

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.split-lip-shape-overhang.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.split-lip-shape-overhang.test.ts
@@ -12,7 +12,7 @@
  * Both surface as a "missing / broken stacking lip" on the split pieces.
  */
 import { describe, it, expect, beforeAll } from 'vitest';
-import { DEFAULT_BIN_PARAMS, GRIDFINITY } from '@/shared/constants/bin';
+import { DEFAULT_BIN_PARAMS } from '@/shared/constants/bin';
 import { DEFAULT_SPLIT_CONNECTOR_CONFIG } from '@/features/bin-designer/constants/defaults';
 import type { BinParams, SplitConnectorConfig } from '@/shared/types/bin';
 import type { CellMask } from '@/shared/utils/cellMask';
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the review feedback in 66a5e06:

- **`computeWallTopZ` now uses `params.heightUnitMm`** instead of the hard-coded `GRIDFINITY.HEIGHT_UNIT`, so the helper generalises to non-default height units (Greptile + Copilot).
- **Tightened the `buildTopShape` comment** — the overhang case *does* change rectangular bins, so "rectangular bins are unaffected" was inaccurate; reworded to "a no-op for a plain rectangular bin with no overhang" (Copilot).
- **`countVerticesInRect` comment** changed from "strictly inside" to "within … (inclusive bounds)" to match the `>=`/`<=` implementation (Copilot).

Lint, typecheck, and the regression test remain green.

Issue #1968, opened 2026-06-02
Gap in corners when column dividers are added to thin walled boxes


JonathanHuss:
**Describe the bug**
When adding column dividers to a thin-walled box, a gap is created in the corners.

**To Reproduce**
1. Create a box with 2 columns and 4 rows.
2. Set the wall thickness to .8mm (or less).
3. Add a column divider down the center.
4. When the column is generated, a gap is also introduced in the 4 corners of the box.

**Expected behavior**
The corners to remain intact.

**Screenshots**
<img width="1214" height="794" alt="Image" src="https://github.com/user-attachments/assets/367a038c-d331-48d2-86b6-30a5e16a4612" />

PR #2109, opened 2026-06-11
fix(bin-designer): keep scoop and label support inside the rounded bin corners

andymai:
## Problem

Two square-cornered features poked past the bin's **rounded** corners (reported from the bin designer):

| | |
|---|---|
| **Finger scoop** | A full-width straight prism with square corners overshoots the rounded outer wall and sticks out of the bin at the front-outer corners, once `wallThickness < BOX_CORNER_RADIUS·(1 − 1/√2) ≈ 1.1mm` (the thin-wall options below the 1.2mm default). |
| **Label tab support** | The shelf plate rounds its free-end front corners, but the gusset/solid/fillet support still ran to the full square corner — leaving "points" sticking out on centered/partial-width tabs. |

Both are the same class of bug as the compartment **cavity cut** fixed in #1968 (square geometry vs. the rounded `BOX_CORNER_RADIUS − wt` inner contour).

## Fix

- **`scoopRampBuilder.ts`** — clip the fused scoop to a rounded-rectangle inner footprint, guarded by the `< 1.1mm` threshold so the boolean is skipped entirely on normal walls (no perf impact on the default path).
- **`labelTabBuilder.ts`** — clip the support to the shelf outline (full tab height) via `intersect`, only when a free end exists. Preserves the rounded-shelf look while trimming the overhang.

## Tests

- New `binGenerator.scenario.cornerProtrusion.test.ts` — written first, confirmed **failing** on current code (scoop: 0.42mm protrusion; label: 6 spike vertices), now passing. Asserts vertex containment within the intended rounded footprint.
- No snapshot churn: scoops/labels/wallThickness/combinedFeatures (54) pass unchanged — fixes only touch the buggy configs.
- export-integrity (357) passes → clip booleans stay manifold/hole-free.
- typecheck + eslint + prettier clean.

## Known limitation

`GhostScoops.tsx` (the transient translucent preview during regeneration) still draws a square-cornered arc strip, so it may briefly show the poke before the real geometry lands. The real/exported geometry is correct.

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.cornerProtrusion.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.cornerProtrusion.test.ts
@@ -1,0 +1,149 @@
+// @vitest-environment node
+/**
+ * Regression: geometry must not protrude past its intended footprint.
+ *
+ *  1. Finger scoop ramps are full-width straight prisms with SQUARE corners,
+ *     but the bin cavity has ROUNDED corners (radius BOX_CORNER_RADIUS − wt).
+ *     On thin walls (wt < BOX_CORNER_RADIUS·(1 − 1/√2) ≈ 1.1mm) the scoop's
+ *     square front-outer corner pokes through the rounded outer wall and sticks
+ *     out of the bin. (Mirror of the cavity-cut fix in #1968.)
+ *
+ *  2. A centered, partial-width label tab rounds BOTH free-end front corners of
+ *     the shelf plate, but the gusset/solid/fillet support beneath still runs to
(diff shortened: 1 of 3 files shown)


andymai:
Addressed Copilot's feedback in 5e8aa471: both the thin-wall scoop clip and the label-support clip now wrap `intersect` in try/catch and fall back to the un-clipped shape on kernel failure — matching the best-effort boolean pattern already used in these files (text booleans, cutout clips). The clips are cosmetic (sub-mm corner trim), so a boolean failure degrades gracefully instead of failing the whole bin build.

PR #2119, opened 2026-06-12
fix(baseplate): relieve split dovetail tongues from neighbour sockets

andymai:
## Problem

A user reported that on a **split baseplate**, the dovetail joints "do not completely follow the shape of the grid … bins do not sit completely snug"

## Root cause

In `buildBaseplateSolid`, dovetail **tongues are fused onto the slab edge *after* the socket pockets are carved** (`baseplateGenerator.ts`), as flat-topped full-height prisms. The gridfinity socket mouth opens to the **full cell at the slab top** (`INSET_TOP = 0`), so a tongue protruding across the seam pokes into the neighbouring piece's open socket — exactly where the bin foot seats.

It's worst in **"prefer identical pieces" (paired) mode**: to make opposite-corner pieces share one STL, paired mode offsets each tongue by `PAIR_HALF_OFFSET` (4 mm — *exactly* the socket corner radius), landing it on the fully-open straight edge of the socket mouth.

I measured the intrusion (tongue ∩ neighbour bin foot) before fixing:

| Config | Foot overlap | 
|---|---|
| Plain / inverted / magnet (junction-centred) | **0.0000 mm³** |
| **Paired (`preferIdenticalPieces`)** | **2.44 mm³ / tongue** |
| Paired, depth-3 (2 tongues) | **4.88 mm³** |
| Dovetail key | **0.0000 mm³** |

So the bin-blocking case is the **paired-mode integral tongue**. The plain tongue and the dovetail key only graze the 0.25 mm clearance gap (foot overlap 0), so they're left as-is.

## Fix

`buildConnectors` now subtracts the neighbouring piece's socket pockets from each fused tongue, trimming it back to the grid's wall region — the assembled plate's top then matches an un-split plate and bins seat flush. The tongue keeps its sub-funnel material (the socket recedes with depth), so the dovetail still locks. This mirrors `relieveClipForSockets()`, which already solved the same problem for the snap-clip style.

The relief is surgical: paired-mode foot overlap drops **4.88 → ~0 mm³** while the tongues retain **>80%** of their volume.

## Tests

New `dovetailSocketInterference.test.ts` intersects the fused tongues with the actual neighbour bin feet (cell − CLEARANCE) and requires ~zero overlap, plus an over-cut guard. Watched it fail (paired = 4.88 mm³) before implementing.

- ✅ new interference tests (paired, plain, magnet)
- ✅ full generators suite (127 files / 1489 tests), baseplate-feature + connector suites (487 tests) — no regressions
- ✅ dovetail / dovetailKey / snapClip / split-stress watertight scenarios
- ✅ typecheck, lint, knip

### DIFF
--- a/src/features/generation/worker/generators/baseplateConnectors.ts
+++ b/src/features/generation/worker/generators/baseplateConnectors.ts
@@ -51,8 +51,9 @@ import {
   sketch,
 } from './generatorTypes';
 import type { SnapClipLevels } from '@/shared/constants/connectors';
-import { computeCellBoundariesMm } from './cellDecomposition';
+import { computeCellBoundariesMm, decomposeCells } from './cellDecomposition';
 import { buildSingleCellSocket } from './socketBuilder';
+import { getPocketTemplate } from './baseplatePockets';
 
 /**
  * Half the separation between the tongue and groove of a paired connector,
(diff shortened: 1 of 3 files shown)


andymai:
Addressed both review findings in 8be9391e (test-only): freed the pre-translate socket + each `intersect` result to avoid WASM heap accumulation, and replaced the `Math.round(depth)` + GU-step foot placement with `decomposeCells()`-derived centres sized per cell (correct for fractional depths). The production relief is unchanged.

PR #2122, opened 2026-06-12
fix(label-tabs): trim tab corners flush with the bin's rounded outer wall

andymai:
## Problem

On bins with thin walls, a label tab's square corner pokes out past the bin's **rounded** outer corner — a visible ledge "sticking out of the wall." Most noticeable on small (e.g. 1×1) bins where a full-width tab reaches into both rounded corners. (Reported on Reddit.)

## Root cause

The shelf is built as an axis-aligned rectangle anchored to the *nominal flat* inner-wall planes (`±innerW/2`), but the bin's outer corner is rounded (`BOX_CORNER_RADIUS = 3.75mm`). The square corner sits outside the rounded wall whenever:

```
wt < BOX_CORNER_RADIUS · (1 − 1/√2) ≈ 1.10mm
```

i.e. the **0.4 / 0.6 / 0.8mm** wall-thickness presets. The 1.2mm default is just above the threshold, so it isn't always visible. Measured **0.42mm** overhang at wt=0.8 on a 1×1 bin.

## Fix

`clipToOuterFootprint()` intersects the assembled tabs with a prism of the bin's outer rounded-rect footprint, trimming any wall-touching corner flush with the wall. It is a **no-op** for thicker walls and for interior-divider tabs that never reach the perimeter.

## Verification

- New regression test (`binGenerator.scenario.label-corner-overhang.test.ts`): builds the real mesh and asserts the label tab stays within the bin's outer footprint (overhang 0.42mm → ~0), with a guard that the tab is actually generated.
- Existing label-tab scenario snapshots **unchanged** (confirms the clip is a true no-op on the default thick walls).
- `tsc` + `eslint` clean.

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.label-corner-overhang.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.label-corner-overhang.test.ts
@@ -6,8 +6,8 @@
  * inner-wall planes, so a full-width tab's square corner sits outside the
  * rounded wall whenever `wt < BOX_CORNER_RADIUS·(1 − 1/√2) ≈ 1.10mm` — i.e.
  * for the 0.4 / 0.6 / 0.8mm wall-thickness presets. Most visible on small
- * bins, where the tab reaches both corners (Reddit report). `clipToOuter
- * Footprint` trims the slivers flush with the wall.
+ * bins, where the tab reaches both corners (Reddit report).
+ * `clipToOuterFootprint` trims the slivers flush with the wall.
  *
  * Footprint is taken from the NO-label bin (the with-label bbox would be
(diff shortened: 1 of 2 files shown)


andymai:
Addressed Copilot review in 1c9cb6331:
- **Perf:** `clipToOuterFootprint` now early-returns when `wt >= BOX_CORNER_RADIUS·(1 − 1/√2) ≈ 1.10mm`, so the boolean `intersect()` only runs on the thin-wall presets (0.4/0.6/0.8mm) that can actually overhang — the default 1.2mm wall skips it entirely.
- **Docstring:** fixed the line-wrapped `clipToOuterFootprint` reference in the regression test.

Regression test + label-tab scenario snapshots still pass; typecheck + lint clean.

PR #2138, opened 2026-06-13
fix(deps): pin esbuild to 0.28.1 and stabilize lockfile handling for OSV

andymai:
## Summary

**Type**: fix

## Context

OSV Scan on `main` was failing due to vulnerable transitive `esbuild` versions in the dev toolchain.  
Goal: make remediation deterministic and keep the dependency/lockfile changes minimal and stable.

## Changes

- Pinned workspace override to exact `esbuild: 0.28.1` (not a range) so resolution does not float during/after the temporary minimum-release-age exception window.
- Kept temporary `minimumReleaseAgeExclude` entries for `esbuild` and `@esbuild/*` packages while `0.28.1` is within cooldown.
- Updated rationale wording to reflect both advisories (`GHSA-gv7w-rqvm-qjhr`, `GHSA-g7r4-m6w7-qqqr`) as first fixed in `0.28.1`.
- Added `.prettierignore` entry for `pnpm-lock.yaml` to prevent lint-staged/prettier from rewriting pnpm’s native lockfile serialization and creating large unrelated diffs.
- Verified resulting resolution to a single `esbuild@0.28.1`.

## Architecture

- **Stores**: none
- **Features**: none
- **New types**: none
- **i18n keys**: none
- **Storage/schema changes**: none

## Breaking changes

None.

## Test plan

- [ ] `pnpm run build`
- [ ] New code has colocated sibling tests
- [ ] i18n: `pnpm run check:i18n`
- [ ] No `console.log`, `any`, `var`, `==`, `!` introduced
- [ ] CLAUDE.md updated if conventions changed


andymai:
Addressed @copilot-pull-request-reviewer's findings:

1. **Exact pin** — `overrides.esbuild` is now `0.28.1` (was `^0.28.1`), so the remediation is deterministic and won't float to a newer (cooldown-blocked) 0.28.x when the temporary exclude is removed.
2. **Comment accuracy** — the rationale now names both `GHSA-gv7w-rqvm-qjhr` and `GHSA-g7r4-m6w7-qqqr`, phrased as "first fixed in 0.28.1".

Also fixed an unrelated side effect: the lint-staged `*.{yml,yaml}` glob was running `prettier --write` over `pnpm-lock.yaml`, rewriting pnpm's native serialization (huge spurious diff). Added a `.prettierignore` entry so the formatter leaves the lockfile alone. Rebased onto the 4.168.0 release on main; the diff is now just the three intended files.

PR #2141, opened 2026-06-14
perf(bundle): cut eager initial JS 904→353KB by lazy-loading the 3D + collab stacks

andymai:
## What & why

The three.js + drei rendering stack (**358 KB gzip**), the Liveblocks collab client (**63 KB**), and the bin example gallery (**48 KB**) were all downloading on **first paint** for every user — despite each having a lazy boundary. They leaked into the eager critical path through barrel re-exports, a static `three` import in the offscreen thumbnail regenerator, and presence UI statically wired into the always-on Header/Grid.

**Eager initial JS: 904 → 353 KB gzip (−551 KB, −61%).**

## How

- **Barrel re-export leaks** — `@/features/bin-designer` and `@/shared/hooks` barrels (imported by eager App / sync flows / design-linking) re-exported heavy components/hooks, dragging three+drei+Liveblocks eager. Dropped the heavy re-exports; deep-imported the light symbols.
- **Deferred `three`** in `thumbnailRegenerator` via `await import('three')` (the fn is already async, runs on idle); made `cameraFraming` three-free.
- **Lazy-loaded the opt-in collab/presence UI** (`PresenceAvatars`, `CollabCursors/Ghosts/SelectionRings`, `ParticipantsPanel`) behind `Suspense` at its 4 always-on mount sites. Collab is off by default, so this ships zero Liveblocks to most users.
- **Chunking** — switched `manualChunks` → Rolldown `advancedChunks`. The Rollup-compat `manualChunks` *function* was being ignored for shared vendors, gluing react-dom + three + drei into one eager mega-chunk (every UI chunk imported it just to get react-dom). Priority-pinned groups now keep `react-vendor` / `state` eager and the 3D + Liveblocks stacks lazy. Key subtlety: **zustand must outrank `three-render`** (drei depends on zustand) or the shared module anchors the whole 3D chunk eager.
- **size-limit** — added an "Eager initial JS" budget (260 KB) to lock the win; bumped Total 1800 → 1900 KB to cover ~40 KB of finer-splitting overhead.

## Trade-off

Total JS rose ~40 KB (+2%, amortized/cached across the app lifetime) — the cost of finer code-splitting. The metric users actually feel (first-paint JS) dropped 61%.

## Verification

- ✅ typecheck, eslint (changed files), module-boundaries, component-structure
- ✅ 447 unit tests pass
- ✅ size-limit green — main **88 KB** (limit 190), eager-init **152 KB** (limit 260), total **1.84 MB** (limit 1.9)
- ✅ Runtime boot smoke (production build, headless): app mounts cleanly with **no chunk-init cycle**; `three-render` and `liveblocks` confirmed **absent from first paint** and lazy-load on demand (3D toggle pulls `three-render`)

## Note on localization

Investigated per a side request — locales are **not** a bundle lever: already lazy per-locale (only the active one ships), only 1–6 % of values match English, 1.4 % identical across all locales, and English isn't eagerly bundled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,8 @@
       "path": [
         "dist/assets/main-*.js",
         "dist/assets/react-vendor-*.js",
-        "dist/assets/state-*.js"
+        "dist/assets/state-*.js",
+        "dist/assets/core-foundation-*.js"
       ],
       "gzip": true,
       "limit": "260 kB"
--- a/package.json
+++ b/package.json
@@ -72,10 +72,20 @@
       "limit": "190 kB"
     },
     {
+      "name": "Eager initial JS (gzip)",
+      "path": [
+        "dist/assets/main-*.js",
+        "dist/assets/react-vendor-*.js",
+        "dist/assets/state-*.js"
+      ],
+      "gzip": true,
+      "limit": "260 kB"
+    },
+    {
       "name": "Total JS (gzip)",
       "path": "dist/assets/*.js",
       "gzip": true,
-      "limit": "1800 kB"
+      "limit": "1900 kB"
     }
   ],
   "dependencies": {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -107,7 +107,7 @@ const CollabProvider = lazyWithRetry(() =>
   import('@/shell/Collab/CollabProvider').then(namedExport('CollabProvider'))
 );
 const BinExampleGallery = lazyWithRetry(() =>
-  import('@/features/bin-designer').then(namedExport('ExampleGallery'))
+  import('@/features/bin-designer/components/ExampleGallery').then(namedExport('ExampleGallery'))
 );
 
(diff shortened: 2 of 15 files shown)


andymai:
Addressed review feedback in eb512a6a8:

- **vite.config comment** (Copilot): fixed — `three` core is intentionally grouped *into* `three-render` (they're always loaded together, so keeping them in one chunk avoids duplicating three's modules across two lazy chunks). Updated the comment to match.
- **Eager budget omits core-foundation** (Copilot): good catch — added `core-foundation-*.js` to the "Eager initial JS" budget. It's pinned eager for the #1466 cycle guard, so it belongs in first-paint accounting. Budget now measures 155 KB (limit 260).
- **import/first** (Greptile P2): moved the two displaced imports above the lazy `const` declarations.

All verified: prettier, eslint, typecheck, size-limit green.

PR #2149, opened 2026-06-14
feat(baseplate): add stack-printable baseplates

andymai:
## Summary

**Type**: feat

## Context

Adds an experimental baseplate stack-print workflow so users can print drawer quantities faster as flipped vertical towers, with optional sacrificial separator sheets for multi-material 3MF.  
Follow-up review feedback was incorporated to harden export behavior, accessibility, docs accuracy, and stack-print guidance.

## Changes

- Added stack-print planning + mesh replication transforms for baseplates (preview/export share the same stacking model).
- Added `StackPrintSection` controls (enable, mode, sets, gap/sheet thickness, live totals) and stacking-specific UI notices.
- Added stacked preview rendering with separation slider for inspection.
- Export now bakes per-stack geometry, splits over-tall groups by printer-height-derived caps, and supports separator-sheet material indexing in 3MF.
- Enforced sacrificial-sheet behavior for 3MF-only stack exports (defensive guard in export path).
- Improved export performance by parsing source STL once per source/group before capped tower splitting.
- Added stack-aware print guidance to ZIP output for single-piece multi-tower exports (`print-guide.txt`).
- Preserved connector settings by functionally stripping connector-related geometry only while stack mode is active (no destructive param mutation).
- Incorporated review hardening:
  - Correct mode-aware aria labels for gap/sheet controls.
  - Added accessible naming for separator color listbox UI.
  - Guarded `releasePointerCapture` with `hasPointerCapture`.
  - Corrected stack preview/docs wording to centered grid layout (not row).
  - Corrected stack-print docs/comments to reflect actual build-height-based tower capping.
  - Improved print guide wording so air-gap mode does not mention interface-material removal.
  - Fixed degenerate-triangle handling in overhang-audit test utility (0 area/0 normal).

## Architecture

- **Stores**: none
- **Features**: baseplate, print-export, generation
- **New types**: `StackPrintParams` (and `BaseplateParams.stackPrint` usage/docs updates)
- **i18n keys**: 20 `baseplate.stackPrint.*` keys across supported locales
- **Storage/schema changes**: `stackPrint` persisted in `baseplateParams` with migration/CQRS schema updates

## Breaking changes

None.

## Test plan

- [x] `pnpm run build`
- [x] New code has colocated sibling tests
- [x] i18n: `pnpm run check:i18n`
- [x] No `console.log`, `any`, `var`, `==`, `!` introduced
- [x] CLAUDE.md updated if conventions changed


andymai:
- **Connectors restore on disable** — stacking now strips connectors functionally in `buildFullParams` (not by mutating stored params), so connector settings return intact when stacking is turned off (survives reload). Connector controls are hidden while stacking.
- **@copilot** `gapMm` doc — corrected: it's the separation size in *both* modes (air gap / sheet thickness).
- **@copilot** `stackPrint` doc — corrected: the config uses `sets` + per-group quantities + height capping, not a `count` field.

andymai:
Addressed the latest review round (@copilot):
- **Export perf** — `buildStackedFileBlob` now takes an already-parsed soup; each source STL is parsed once per group/source instead of once per capped tower.
- **Single-piece multi-tower ZIP** — now includes a `print-guide.txt` (new `generateStackPrintNote`) with the upside-down + peel instructions, matching the split-export guide.
- **a11y** — the gap/separator-thickness stepper's `aria-label` now follows the selected mode instead of always reading "gap".

(Greptile's earlier 'README not updated' note was already resolved in 796dc561 — the baseplate README now documents the stack-print flow, `StackPrintParams`, connector auto-disable, and the 3MF-only separator constraint.)

andymai:
Addressed the latest review pass (@copilot):
- Synced the stale "centered row" comments (stackPreview header + README) to the grid layout, and corrected the `buildFullParams` Key-Files line to note it strips magnets + corner rounding too.
- Print guide: air-gap mode no longer tells users to "remove interface material" (there's none) — split into a per-mode finish line.
- `StackSeparationSlider`: guard `releasePointerCapture` with `hasPointerCapture` to avoid a throw on lost/cancelled capture.
- overhang-audit test: degenerate (zero-area) triangles now contribute 0 area / 0 normal instead of `len = 1`.

andymai:
Review comments addressed. The bulk targeted the **separator-sheet mode** and **"sets" multiplier**, both of which were removed in the rework (the feature is now air-gap-only, single-material; PLA/PETG separation is documented as a slicer support-interface setting per community practice). That resolves the sacrificialSheet/3MF-forcing/separator-color-picker/sheet-footprint/"centered row" comments.

Remaining nits fixed directly:
- Stale `sets` comments in `useBaseplateExport` removed.
- `STACK_PRINT_MAX_STACK_HEIGHT` doc clarified as a fallback default (real cap is `stackHeightCap()` from build height).
- Orientation now matches community models (bottom plate upright, rest flipped).
- Degenerate-triangle area guard, pointer-capture guard, README sync — all in.

Previously-confirmed: single-piece multi-tower exports DO include the stacking print guide; `buildStackedFileBlob` parses the source STL once per tower set.

PR #2158, opened 2026-06-15
fix(bin-designer): make overhang, multi-color & lid feature toggles; sidebar polish

andymai:
## Summary

Sidebar polish pass for the bin designer — standardizes feature controls on the shared `FeatureToggle` pattern and tightens the typographic/spacing rhythm.

### Feature toggles
- **Overhang**: `Collapsible` → `FeatureToggle`. Adds an explicit `enabled` flag to `OverhangConfig` so per-side values are retained while the feature is off (mirrors how Scoop keeps its radius); `resolveOverhang` returns zero when `enabled === false`, and legacy designs (no flag) still infer enablement from any non-zero side.
- **Multi-color**: dissolved the redundant "Multi-Color" group panel (the panel wrapped a toggle that did the same job) into a standalone `FeatureToggle`. Moved it into the **Base** group between Base and Physical Units, dropped "Enable" from the title, and restyled its Experimental badge to match the former custom-shape badge.
- **Lid**: folded into the Walls section so it's no longer its own bordered block.

### Visual polish
- Removed the Experimental badges from Custom shape and Lid (Multi-color keeps the now-warning-styled one).
- Un-bolded the wall-thickness `SnappingSlider` label to match feature-toggle titles (also normalizes the Lid/Compartment sliders).
- Dropped the `(…mm)` figures from the Shape group header summary.
- Centered the physical-dimension readout under the width/depth steppers.
- Shortened "Set as default for new bins" so it fits on one line at normal size.

### i18n
- Removed the orphaned `binDesigner.multiColor.enableLabel` from `en.ts` + all 9 locale JSONs (the toggle label now reuses `binDesigner.group.colors` = "Multi-Color", already translated everywhere). Regenerated `en.json`.

## Test plan
- `pnpm typecheck` ✅
- `pnpm lint` (changed files) ✅
- `check:i18n`, `check:i18n:unused`, `check:i18n:values`, `check:i18n:interpolation` ✅
- Affected unit tests green (Overhang, Multi-color, ParameterPanel, Dimensions, Shape, Lid, SetDefaultFooter, overhang generator)

### DIFF
--- a/src/design-system/Stepper/Stepper.tsx
+++ b/src/design-system/Stepper/Stepper.tsx
@@ -76,9 +76,9 @@ const inputVariants = cva(
   {
     variants: {
       size: {
-        sm: 'text-xs min-w-[40px]',
-        md: 'text-sm min-w-[48px]',
-        lg: 'text-base font-semibold min-w-[64px] border-stroke',
+        sm: 'text-xs min-w-[34px]',
+        md: 'text-sm min-w-[38px]',
+        lg: 'text-base font-semibold min-w-[48px] border-stroke',
       },
     },
     defaultVariants: {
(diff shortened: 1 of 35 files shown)


andymai:
Addressed the one review finding (flagged by both Copilot and Greptile): the multi-color help deep-link's `surface` was stale after the Multi-Color group was folded into Base. Fixed in `06a9aa4fa` by pointing the help entry at `surface: 'binDesigner:base'` so the existing `setBaseExpanded(true)` handler runs before the dispatcher scrolls/pulses `bd-colors` (avoids firing on an `inert`, collapsed element). Verified with typecheck + ParameterPanel/helpEntries/helpJumpDispatcher tests.

PR #2162, opened 2026-06-15
fix(baseplate): snap-clip resists pull-apart via seam-side retaining wall

andymai:
## Problem

The snap-clip baseplate connector (`connectorStyle: 'snapClip'`) provided only **vertical (lift-out) retention** — it did nothing to stop the two baseplates from being **pulled apart** across the seam, which is the connector's primary job.

Root cause: the seam pockets were cut *open to the seam*, so every pocket wall that touched a leg sat on the **outer** side and retreated as the plates separated. The barbs catch a horizontal (Z) ledge, which resists the clip lifting out — not the plates spreading. With a clearance fit and nothing on the seam side to bear against, pull-apart resistance was ≈ 0.

## Fix

Leave a solid **retaining wall** between the seam and each leg's inner face, in the upper band just under the bridge (the leg root, which barely flexes). The two seam-sides' walls meet at the seam and **nest between the clip's prongs**, so pulling the plates apart rams the wall into the legs — bearing in shear, like a real staple. The lower band stays open to the seam so the leg tip can still pinch inward to seat the barb.

- **Prongs:** legs spread (`GAP_HALF` 0.75→1.0) and thinned (`LEG_W` 1.2→0.95) to make room for the wall while keeping the **same outer footprint** (legOuter 1.95, barbTip 2.4) — so bin-foot clearance is unchanged.
- **Baseplate pocket:** split into bridge-recess / bearing-band / lower-throat / chamber; the bearing band leaves the wall solid.
- **Clip shape, chamfer, and root fillets are unchanged.**

Bearing-band height (`BEAR_DEPTH` 1.2mm) is clamped above the catch ledge (never eats the snap) and leaves enough free leg below to flex. Wall height vs. snap flex is a tunable trade-off.

## Verification

All brepjs scenario tests green:
- **`snapClipPullApart` (new):** seats with clearance, but pulling a plate off the seam is now blocked by the clip (the fix; was 0mm³ interference, fails on the old geometry).
- Watertight + valid bed-flat print · clears the adjacent bin feet · snap zone intact · catch-ledge backed · nozzle scaling.
- 508 baseplate/sample tests · typecheck + lint clean.

Note: the test proves *geometric* capture; real-world hold is "firm staple," not a hard interlock — a print test is the final confirmation.

### DIFF
--- a/src/features/generation/worker/generators/baseplateConnectors.ts
+++ b/src/features/generation/worker/generators/baseplateConnectors.ts
@@ -413,13 +413,21 @@ export function buildDovetailKey(totalHeight: number): Shape3D {
 /**
  * Blind snap-clip pocket on one seam side: a narrow throat (top → ledge) over a
  * wider chamber (ledge → floor). The throat passes the leg but blocks the barb;
- * the barb springs into the chamber and catches the ledge. Returned as two
- * stacked cutters (throat, chamber) to subtract from the slab. Mirror of the
- * snap clip's leg cross-section; pocket carries the clearance.
+ * the barb springs into the chamber and catches the ledge.
  *
- * Levels come from the shared `snapClipLevels` so the pocket, the clip, and the
- * seated-clip preview can't drift. Proven standalone with brepjs-verify: at the
- * thin 5mm slab the seated clip clears the pockets with zero interference.
+ * The throat is cut in three Z-bands so a RETAINING WALL is left solid between
+ * the seam and the leg's inner face — the wall the leg bears against to resist
+ * pull-apart (the two seam sides' walls meet at the seam, nesting between the
+ * clip's prongs):
+ *   - bridge recess (top → −BRIDGE_THK): open across the seam so the flush bridge
+ *     clears it,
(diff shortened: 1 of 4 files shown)


andymai:
Addressed the review feedback in b8720e219 (test-only + a comment fix, no geometry change):

- **`overlap()` helper** (Greptile + Copilot): now deletes the `intersect` result and **throws** on a failed boolean instead of returning `0` — a swallowed error could otherwise let the "seated clearance fit" assertion pass spuriously.
- **WASM cleanup in test body** (Greptile): the `clip`, both plate solids, and `pulled` are now `.delete()`d (and `box` inside `rightPlateWithPocket`), matching the cleanup discipline in `dovetailSocketInterference.test.ts`.
- **Comment accuracy** (Copilot, `connectors.ts`): clarified that the "same outer footprint" invariant holds at the 0.4 mm baseline; on wider nozzles `scaleFeature` floors leg width to 2× the nozzle, so the outer face grows by the ~0.25 mm `GAP_HALF` increase.

Pull-apart test still green; lint/typecheck clean.

PR #2169, opened 2026-06-15
feat(baseplate): refine generator panel, connectors, and stack/fit-sample exports

andymai:
## Summary

A batch of UX/visual refinements to the **/baseplate** generator screen, plus export improvements.

### Panel & steppers
- Renamed baseplate **"Print Settings" → "Physical Units"** (consolidated into a shared `common.physicalUnits` key used by the bin designer and layout sidebars too).
- Moved **nozzle size** into Physical Units.
- Folded the bespoke `PaddingStepper` into the **design-system `Stepper`** — added a vertical orientation variant, custom +/- aria labels, an Enter/blur double-commit guard, hidden native number spinners, and crisp 12px `sm` icons.
- Slimmed the vertical (left/right) padding steppers to a fixed 32px column and added **Left/Right labels**.
- **Magnet holes** and **corner radius** now have bordered separators.

### Connectors
- Replaced the dropdown with an **Interior-section-style card picker**: icon + title + description, and the selected card **expands inline** to its controls (tolerance, invert dovetails, fewer unique parts, print fit sample). Accurate solid-filled SVG icons drawn from the real connector geometry.
- Renamed **"Connector fit" → "Tolerance"**; dropped the dovetail-specific "(rotate to fit)" qualifier.

### 3D preview & stacking
- Footprint grid now renders in **stack mode**, and towers **snap to the scene grid** (single or multiple).
- New **stack fit-sample** export (one tower of two 1×1 plates).
- Hint text explaining vertical stack when enabled.

### Exports
- Connector fit sample now generates **only the selected style** with coarser tessellation — the 31MB / >3min export is now small and fast.
- Export dialog shows a **progress bar** during export.

### Shared
- Extracted an **`ExperimentalBadge`** (on the design-system `Badge`) and replaced all inline experimental badges so they match.

## Testing
- `pnpm typecheck`, all 4 i18n checks, lint, and the affected unit/scenario tests pass locally.


andymai:
Addressed review feedback in dbb820bf1:

- **`stackPreview` `Math.round` → `Math.ceil`** (Copilot) — a fractional tower footprint can no longer undersize the cell.
- **`PaddingStepper` label association** (Greptile) — added an optional `id` prop to the design-system `Stepper` and render a real `<label htmlFor>`, so clicking the Left/Right/Back/Front label focuses the input again.
- **`ConnectorPicker` keyboard nav** (Copilot) — the radiogroup now has roving tabindex + Arrow/Home/End navigation, matching `SegmentedControl`.
- **`StackSampleButton` toast** (Copilot) — uses a dedicated `baseplate.stackPrint.sampleExportComplete` key instead of the connector-sample one.

**On the 18 "translate to <locale>" comments:** the new strings are intentional English placeholders registered in `scripts/i18n-values-allowlist.json`, which is this repo's sanctioned mechanism for new copy pending a translation pass (CI's `check:i18n:values` passes via the allowlist). Happy to do a proper translation sweep across all 9 locales as a follow-up if preferred.

PR #2181, opened 2026-06-15
feat(export): grant exports a far higher timeout ceiling

andymai:
## Why

Exports shared the **live-preview** timeout, which clamps at a 3-minute ceiling. That ceiling is tuned for interactive iteration on a *reference* machine — but an export is a different beast:

- **User-committed** — they clicked Export and will wait.
- **Terminal & cancellable** — there's progress + a cancel affordance.
- **Runs on the user's hardware** — which can be several times slower than the dev machine the budgets were measured on.

A complex bin/baseplate that finishes in ~3 min on a fast machine could legitimately exceed it on a low-end laptop or throttled mobile browser and time out mid-export.

## What

Split the cost model from the clamp in `generationTimeout.ts`:

- Extracted the uncapped **raw budget** (`binRawBudgetMs` / `baseplateRawBudgetMs`) — base budget + all complexity bonuses.
- **Preview** functions (`computeGenerationTimeoutMs`, `computeBaseplateTimeoutMs`) clamp it at the unchanged 3-min ceiling — **behavior identical to before**.
- **Export** functions (`computeExportTimeoutMs`, `computeBaseplateExportTimeoutMs`) apply `EXPORT_TIMEOUT_MULTIPLIER` (×4, modeling slow hardware) and clamp to a new 15-min `EXPORT_MAX_TIMEOUT_MS`.

Every export path in `bridgeExports.ts` now uses the export budget — **bins and baseplates** (`exportBaseplate`, `exportConnectorKey`), plus split/combined/dividers and the connector fit-sample (shares the high ceiling).

A constant **multiplier** (not a flat value) is intentional: the preview budget already encodes *relative* complexity correctly, so ×4 just says "assume up to 4× slower hardware" while keeping a hex bin proportionally longer than a trivial one. The high ceiling is safe precisely because exports have progress + cancel — the clamp's only remaining job is to stop a genuinely-wedged WASM heap, not to bound interactive wait.

## Tests

- Added `describe` blocks for both new functions: multiplier scaling, complexity ordering preserved, runs past the preview ceiling, export-ceiling clamp, NaN/negative guards.
- Bumped the bridge export-timeout test's advance to clear the new 120s default-export boundary.
- Full unit/dom suite green (13,539 passed, 0 failed); typecheck + lint clean.

### DIFF
--- a/src/features/generation/bridge/GenerationBridge.test.ts
+++ b/src/features/generation/bridge/GenerationBridge.test.ts
@@ -440,8 +440,9 @@ describe('GenerationBridge', () => {
       ) as { payload: { requestId: string } };
       expect(exportMsg).toBeDefined();
 
-      // Default bin timeout is at least BASE (30s); advance well past it.
-      await vi.advanceTimersByTimeAsync(120_000);
+      // Export timeout is the complexity budget (BASE 30s for a default bin)
+      // scaled by EXPORT_TIMEOUT_MULTIPLIER (×4 = 120s); advance well past it.
+      await vi.advanceTimersByTimeAsync(180_000);
 
       const result = await settled;
       expect(result).toBeInstanceOf(Error);
--- a/src/features/generation/bridge/bridgeExports.ts
+++ b/src/features/generation/bridge/bridgeExports.ts
@@ -10,7 +10,11 @@
 
 import type { BinParams, BaseplateParams, SplitConnectorConfig } from '@/shared/types/bin';
 import type { WorkerMessage, ExportFormat } from './types';
-import { computeBaseplateTimeoutMs, computeGenerationTimeoutMs } from './generationTimeout';
+import {
+  computeBaseplateExportTimeoutMs,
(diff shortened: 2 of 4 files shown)


andymai:
Addressed the review feedback in b9efe694:

- **`generateSplitPreview` / `generateSplitPreviewRange` should use the preview clamp** (3 comments): Agreed — these feed the interactive split preview (`useSplitPreview.ts` re-runs them from `useEffect` as the user drags cut planes), so they're not user-committed exports. Reverted both to `computeGenerationTimeoutMs` (3-minute preview clamp) so a wedged worker is cancelled promptly instead of waiting out the 15-minute export ceiling. The actual export-download paths (`exportSplitBin` / `exportSplitBinRange`) keep the export budget.
- **Stale baseplate docblock**: Merged the orphaned formula comment into `baseplateRawBudgetMs`'s own doc (it documents the raw, unclamped budget now).

Typecheck + lint clean; timeout unit tests green (31 passed).

PR #2192, opened 2026-06-16
refactor(design-system): migrate shell + cloud-share controls to design system

andymai:
## What

Batch 6 (final) of the design-system migration. Migrates raw `<button>` and **three** `SharePermission` `<select>` elements across **8** `shell` (Mobile + Modals) and `cloud-share` components to `Button`/`IconButton`/`Select`. Drains 9 from `scripts/design-system-allowlist.txt` (8 migrated + the stale, already-clean AppearanceTab entry).

Files: BinContextMenu, MobileCloudSharePanel, MobileLayoutsListItem, MobileLayoutsPanelParts (+ `<select>`), HalfGridModeBlockedModal, DeleteShareConfirm, ShareLinkSection (+ `<select>`), SharePopover (+ `<select>`).

- Buttons → `Button`/`IconButton` (delete-confirm → `Button variant="danger"` with `loading`; close X → `IconButton`). Mobile controls keep comfortable touch targets.
- The three `SharePermission` `<select>`s → `<Select options value onValueChange>`; each gains an `aria-label` (reusing the existing `share.anyoneWithLinkCan` key — no new i18n keys).
- Adds render+interaction smoke tests for the 3 mobile files that lacked sibling tests.

## Verification
- ✅ `check:design-system` 0 new violations · typecheck · lint · all i18n checks clean
- ✅ Full suite: 13,597 passed, 7 skipped, 0 failed

Draft until visually eyeballed + Greptile.

PR #2191, opened 2026-06-16
refactor(design-system): migrate grid-editor/layers/snapshots controls to design system

andymai:
## What

Batch 5 of the design-system migration. Migrates raw `<button>` across **7** components in `grid-editor`, `layers`, and `snapshots`, plus one raw `<select>` → `Select`. Drains 7 from `scripts/design-system-allowlist.txt`.

Files: IsometricPreviewControls (10-button overlay toolbar), LayerLabel, SelectionToolbar (2 buttons + move-to-layer `<select>`), SizeSelectorPopover, HeightCrossSectionDiagram, SnapshotEntry, SnapshotHistory.

- Buttons → `Button`/`IconButton` (delete actions → `dangerGhost`); dense diagram/swatch controls keep size via `!px-0`/`touchTarget={false}`.
- The move-to-layer `<select>` → `<Select options={...}>` **keeping native `onChange`** so the `value=''` reset after `onMoveToLayer` still fires (placeholder preserved as a disabled leading option).
- Pure markup swaps — no logic/behavior changes.

## Verification
- ✅ `check:design-system` 0 new violations · typecheck · lint clean
- ✅ Full suite: 13,583 passed, 7 skipped, 0 failed (no test edits needed — role/name/`selectOptions` queries survive)

## Visual notes (within "normalize, small shifts accepted")
- Move-to-layer Select restyles to the design-system chevron/`bg-surface` look in the dense toolbar — functionally identical.
- Category swatches kept at 16px (`!min-w-0/!min-h-0`); isometric segmented active state kept via tailwind-merge (`bg-accent`).

Draft until visually eyeballed + Greptile.

PR #2190, opened 2026-06-16
refactor(design-system): migrate baseplate buttons to Button/IconButton

andymai:
## What

Batch 4 of the design-system migration. Migrates raw `<button>` elements across **6** `src/features/baseplate` components to `Button`/`IconButton`, and removes the stale **PaddingStepper** entry (already clean). Drains 7 from `scripts/design-system-allowlist.txt`.

Files: BaseplatePreviewControls (13-button view/zoom toolbar), BaseplatePage, ConnectorPicker, EditableDimensions, SplitViewStrip, PaddingAnchor (+ PaddingStepper allowlist removal).

- Dense toolbar icon controls → `IconButton touchTarget={false}`; toggles use `pressed`; pads/segmented/inline-edit triggers → `<Button variant="ghost">` with padding-neutralizing overrides (`!px-0`) to preserve tight grid/inline layouts.
- Pure markup swaps — no logic/behavior changes.

## Verification
- ✅ `check:design-system` 0 new violations · typecheck · lint clean
- ✅ Full suite: 13,583 passed, 7 skipped, 0 failed (no test edits needed)

## Visual notes (within "normalize, small shifts accepted")
- PaddingAnchor 3×3 pad cells and SplitViewStrip mini-map cells kept at original size via `touchTarget={false}` + `!px-0/!py-0` (avoids grid distortion).
- Active toggles on mobile preview controls keep their `bg-accent` styling via tailwind-merge (worth confirming accent fill over the IconButton `pressed` grey).

Draft until visually eyeballed + Greptile.

PR #2189, opened 2026-06-16
refactor(design-system): migrate shared/components buttons to Button/IconButton

andymai:
## What

Batch 3 of the design-system migration. Migrates raw `<button>` elements across **9** `src/shared/components` files to `Button`/`IconButton`. Drains all 9 from `scripts/design-system-allowlist.txt`.

Files: ExperimentalKernelBadge, ExportDialog, ExportSupportPrompt, FeatureToggle, FractionalEdgeToggle, HeaderSupportLinks, PrintBedInput, StickyGroupHeader, UserDock.

- Icon-only → `IconButton`; text → `Button` (ghost/secondary/primary/danger); segmented/toggle/menu-row → `<Button variant="ghost">` keeping existing `className`. Async submit (ExportDialog download) → `loading` prop (drops hand-rolled spinner svgs).
- Pure markup swaps — no logic/behavior changes.

## Verification
- ✅ `check:design-system` 0 new violations · typecheck · lint clean
- ✅ Full suite: 13,583 passed, 7 skipped, 0 failed (no test edits needed — role/name queries survive the swap)

## Visual notes (within "normalize, small shifts accepted")
- Link-style buttons (Labs-settings, GitHub-star) flattened with `px-0/h-auto/rounded-none/font-normal` to stay text-links.
- Ko-fi CTA normalizes to the primary accent (was a hand-branded color); white text/textShadow preserved.
- Full-width bars (StickyGroupHeader, UserDock trigger) use `justify-start rounded-none` to preserve the left-aligned bar look.

Draft until visually eyeballed + Greptile. One of a series of per-area button batches.

PR #2196, opened 2026-06-16
refactor(design-system): combined button/select migration (#2188–#2192)

andymai:
## Combined design-system migration (batches 1–6)

Integration branch bundling the five raw-control → design-system migrations so they can be reviewed together. Each source PR remains independently mergeable; this branch is the combined surface.

Rolls up:
- #2188 — bin-designer/components → `Button`/`IconButton`
- #2189 — shared/components → `Button`/`IconButton`
- #2190 — baseplate → `Button`/`IconButton`
- #2191 — grid-editor/layers/snapshots → `Button`/`IconButton`/`Select`
- #2192 — shell + cloud-share → `Button`/`IconButton`/`Select`

### Beyond the raw migrations, this branch also carries the review + visual-regression fixes
- **Review feedback** from Copilot/Greptile on each PR (dangerGhost hover, link-button hover suppression, `resetAllStores()` in new tests, JSDoc, redundant sr-only, `Select` `fullWidth`, etc.).
- **Visual-regression sweep** (two passes) neutralizing what the design-system primitives inject vs. the original raw elements:
  - active/selected ghost controls no longer flip to neutral-gray on hover,
  - `font-medium` no longer bleeds onto previously-normal-weight labels/child text,
  - inline/link buttons keep zero padding + no hover pill,
  - dense `IconButton`s restored to their compact size (`size="sm"`/`touchTarget={false}`),
  - `justify-center` layout break in `InteriorModeCard` fixed,
  - flush segmented pills get `rounded-none`,
  - `ExportDialog` disabled-format `title` tooltip restored.
- `PrintBedInput` was reverted to a raw `<button>` (its padding-sized toggle can't map cleanly to `IconButton`) and kept in the allowlist — it's the only remaining allowlist entry after this branch.

### Notes
- `scripts/design-system-allowlist.txt` is drained to a single entry (`PrintBedInput.tsx`).
- `tsc --noEmit` clean; design-system usage check clean.


andymai:
Addressed review feedback:

- **Copilot — `HalfGridModeBlockedModal` static `aria-label` during loading**: fixed in c310ea26. The `aria-label` is now omitted while `isRemediating`, so the visible "Loading…" text + `aria-busy` (from the `loading` prop) convey the busy state; the descriptive count label is kept at rest. (Note: this behavior pre-dated the migration — the original raw button had the same static label — but it's a fair a11y improvement.)
- **Greptile — `SnapshotHistory` create/save-checkpoint link buttons**: already addressed in the visual-regression sweep. Both carry `font-normal h-auto !px-0 hover:bg-transparent hover:text-accent-hover` — equivalent to the suggestion (using `!px-0` for a stronger override than `px-0`).
- **Greptile — `SwipeActionBgColor` union note**: intentional. Narrowing `bgColor` from `string` to the 4-literal union is what makes the `hover:` map Tailwind-JIT-safe; all callers pass values in the union (verified, incl. the test file).

All CI checks were green prior to this commit; re-running on the new commit.

andymai:
Addressed the latest review pass (`09aea4fe`):

- **Copilot — `HeaderSupportLinks` legacy `.btn`/`.btn-ghost`/`.btn-primary` on Buttons (×3)**: removed the legacy classes from the Feedback, Help, and Ko-fi `Button`s — they were already overridden by `variant` + Tailwind utilities, so this is visually a no-op and drops the dual-styling smell. The GitHub element is a plain `<a>` (not a design-system `Button`), so it keeps `.btn btn-ghost` for its base styling.

Previously-raised comments, for the record:
- **`HalfGridModeBlockedModal` aria-label** — already fixed in `c310ea26` (aria-label is omitted while remediating).
- **`SnapshotHistory` link buttons** — already carry the suggested `font-normal h-auto px-0 hover:bg-transparent` reset.
- **`SwipeActionBgColor` union** — intentional (keeps the `hover:` map JIT-safe).

PR #2218, opened 2026-06-17
perf(baseplate): draft preview + defer BREP for large plates

andymai:
## Problem

Large magnet baseplates time out the live **preview** BREP build. PostHog `baseplate_preview_timing` (30d) shows ~150 distinct users hitting `success=false`, **mostly warm desktop** (so not cold-WASM) with brep times 10–40× the success median — sitting right at the per-piece magnet budget (`30s + cells×200ms + lightweight`). Biggest bucket is modest 2–9 piece splits (133 users); the 26×26 is the extreme tail.

## Root cause

The OCCT magnet-hole boolean pipeline is intrinsically expensive. Two "free perf" levers were **measured and rejected**: `cutAll({trackEvolution:false})` = 0% (metadata propagation isn't the bottleneck) and raising `BOOLEAN_BATCH_SIZE` = ~3% + reintroduces the OOM risk the 64-cap guards. The one real lever: the **lightweight floor cutters are ~⅓ of build time** on magnet grids — and they're underside-only material removal that the procedural draft mesh already omits.

## Changes (no impact on exported geometry)

**A. Draft-quality preview (measured −35%).** A `draft` flag threaded `generateHandler → generateBaseplate → buildBaseplateSolid` (+ mesh cache key) skips the lightweight floor cutters in the **live preview only**. Export is a separate function (`exportBaseplate`) and keeps full geometry. Measured: 6×6 preview 2688→1735 ms, 8×8 5048→3248 ms. Bonus: removes a direct→BREP underside "pop" (the procedural draft already renders a solid underside).

**B. Defer BREP for large plates.** New `previewComplexity.shouldDeferBrepPreview` (magnet-gated): defers when the largest unique piece ≥64 cells, total unique work ≥200 cells, or the last real BREP took ≥25 s (adaptive to slow hardware). On defer, keep the faithful instant procedural mesh and mark the preview complete — no doomed timeout. Wired into both `runGeneration` and the mount-effect bridge-ready path. Export rebuilds at full fidelity on its own larger budget.

Adds a `deferred` dimension to `baseplate_preview_timing` to measure the fallback rate post-launch.

## Why dedup doesn't already fix this

`groupPiecesByFingerprint` cuts the *number* of generations, but not single-piece cost or the per-piece timeout; the worst bucket (small splits of edge/corner pieces) barely dedups.

## Testing

- typecheck clean; lint 0 errors; `check:i18n` clean (no strings added)
- 547 baseplate tests pass — incl. overhang-audit + winding, which confirm **export** geometry is unchanged
- New tests: `previewComplexity` predictor (9), `baseplateGenerator.draft` (draft skips lightweight / no-op without magnets), cache-key draft keying, analytics `deferred` field

### DIFF
--- a/src/features/baseplate/hooks/useBaseplateGeneration.ts
+++ b/src/features/baseplate/hooks/useBaseplateGeneration.ts
@@ -316,11 +316,15 @@ export function useBaseplateGeneration(): void {
       fullParams: FullBaseplateParams,
       bedWidthMm: number,
       bedDepthMm: number,
-      epoch: number
+      epoch: number,
+      // Callers that already computed the tiling pass it in to avoid a second
+      // computeBaseplateTiling pass (cheap, but redundant on large plates).
+      precomputedTiling?: BaseplateTiling
     ): BaseplateTiling => {
       directMeshStartRef.current = performance.now();
 
-      const tiling = computeBaseplateTiling(fullParams, bedWidthMm, bedDepthMm);
+      const tiling =
+        precomputedTiling ?? computeBaseplateTiling(fullParams, bedWidthMm, bedDepthMm);
       setTiling(tiling);
       setSplitProgress(null);
       setDedupStats(null);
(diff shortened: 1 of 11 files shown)


andymai:
Addressed Copilot's review in dcbe4b5f9:

- **Cache fragmentation** — `meshCacheKey` now only folds `draft` into the key when it can actually change geometry (`magnetHoles && lightweight !== false`); a no-op draft no longer splits the mesh-result LRU. Added a regression test.
- **Redundant tiling compute** — `runDirectMeshPreview` accepts an optional precomputed tiling; `runGeneration`'s deferred and direct-mesh branches pass the tiling they already computed.
- **Doc wording** — `DEFER_MAX_PIECE_CELLS` / `DEFER_TOTAL_CELLS` docs now say "at or above (`>=`)"; `DEFER_LAST_BREP_MS` doc now matches the actual behavior (defers the next magnet plate regardless of its complexity).
- **Harness type** — `GenerateBaseplateFn` in `wasmInit.ts` gains the `draft?` param to match `generateBaseplate`.

Note: this PR fixes the live **preview** timeouts. A separate PR will follow for **export** performance (cheaper lightweight floor cutter + coarser export tessellation + cross-session geometry cache).

PR #2220, opened 2026-06-17
perf(baseplate): faster exports — simpler floor relief + coarser STL tessellation

andymai:
## Goal

Make large baseplate **exports** faster. (Companion to #2218, which fixes preview timeouts; this PR is the export-side work.)

## Why per-piece build is the lever

Split exports dedupe to a small, **saturating** set of unique shapes — a 26×26 (25 pieces) builds only ~9 unique shapes and clones the rest; an 18×18 also ~9. So total export cost ≈ `unique_pieces × per-piece-build`, and the unique count is already bounded by cloning. The way to speed exports is therefore to cut **per-unique-piece build cost** — which helps every shape, every grid size.

## Changes (both measured, both watertight)

**1. Simpler lightweight floor relief.** The cross-shaped underside relief had 4 filleted inner corners. The cut is a vertical prism, so sharp concave corners print identically — but the curved fillet faces roughly *doubled* the per-cell boolean cost. Dropping them cuts build by **16–18%**. (Curved faces are disproportionately expensive in OCCT booleans vs planar.)

**2. Coarser export STL tessellation.** Default linear tolerance 0.01→0.02 mm (angular 5°→6°) — an order of magnitude below any FDM nozzle/layer resolution, so visually and functionally indistinguishable, while roughly halving the tessellation triangle budget on large plates.

## Measured (magnet plates, OCCT, node)

| Plate | exportBaseplate before | after | delta |
|---|---|---|---|
| 6×6 | 2931 ms | 2365 ms | **−19%** |
| 8×8 | 5502 ms | 4389 ms | **−20%** |

## Safety

- Winding scenario suite passes **at the new 0.02 mm tolerance** (directed-edge uniqueness + positive signed volume = watertight, no non-manifold edges).
- Overhang-audit scenario passes — the relief is still overhang-safe (vertical walls unchanged).
- Geometry change is cosmetic (underside relief inner corners); the printed part is functionally identical and still material-saving.

## Follow-up

Cross-session geometry/mesh cache (IndexedDB) for near-instant re-exports is the next, larger piece — tracked separately.

### DIFF
--- a/src/features/generation/worker/generators/baseplateGenerator.ts
+++ b/src/features/generation/worker/generators/baseplateGenerator.ts
@@ -427,8 +427,13 @@ export async function exportBaseplate(
     // STL ourselves — OCCT's StlAPI.Write fails on baseplate geometries.
     // brepjs already produces face-consistent winding so no per-triangle
     // correction is applied (see #1472).
-    const tol = tolerance ?? 0.01;
-    const angTol = angularTolerance ?? 5;
+    //
+    // Default linear tolerance is 0.02mm — an order of magnitude below any FDM
+    // nozzle/layer resolution (~0.1-0.4mm), so it's visually and functionally
+    // indistinguishable from the prior 0.01mm while roughly halving the
+    // tessellation triangle budget on large plates. Callers can still override.
+    const tol = tolerance ?? 0.02;
+    const angTol = angularTolerance ?? 6;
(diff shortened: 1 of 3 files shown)


andymai:
Addressed Copilot in 462242d82 — updated the module header to describe the new sharp-corner relief (it still said 'filleted').

PR #2221, opened 2026-06-17
perf(baseplate): dedupe geometrically-identical split pieces via corner signature

andymai:
## What

Split-export piece dedup (`groupPiecesByFingerprint`) over-distinguished pieces. Whenever corners were rounded, it keyed the fingerprint on **raw edge labels**, so a top-edge, bottom-edge, and interior piece of the same size each got a distinct fingerprint and a **separate BREP build** — even though none of them rounds a corner and they're geometrically identical.

This keys the no-connector case on **which corners actually round** (both adjacent edges exterior) instead of raw edge labels, so those pieces collapse to a single build.

## Why it's safe

- The exterior-corner predicate is extracted into a shared `baseplateCorners` helper used by **both** the dedup fingerprint and `buildSlabProfile`'s rounding, so they can't drift.
- **Connectors unchanged**: they live on join edges, so the connector case still keys on the full edge layout.
- **Latent gap closed**: `canonicalizeEdges` (the 180°-rotation dedup) now only runs in the connector branch, where the placement actually applies the compensating rotation. Previously it was gated on `preferIdenticalPieces` alone while the rotation was gated on `preferIdenticalPieces && connectorNubs` — a mismatch that could have mis-oriented a piece if the flag were ever set without connectors.
- Merged pieces are placed by **translation only** (no rotation), and are byte-identical (same size, padding, magnet/grid config, zero rounded corners). `split-stress`, `winding`, and `dovetail` scenarios pass.

## Effect (magnets, 256mm bed, unpadded rounded)

| Grid | unique builds before | after |
|---|---|---|
| 18×18 | 9 | **5** |
| 24×24 | 9 | **5** |
| 26×26 | 9 | **7** |

**Scope/limitation:** this helps *unpadded* rounded plates. With drawer-fit padding, opposite-edge pieces are genuinely different parts (a top-padded piece can't substitute for a bottom-padded one without flipping it over), so they correctly stay distinct. Deduping those requires a 180° in-plane rotation — i.e. `preferIdenticalPieces` — which is a separate product decision (rotated pieces + assembly-guide implications), discussed in the PR thread / follow-up.

No layout or orientation change. Pairs with #2218 (preview) and #2220 (export build speed).

### DIFF
--- a/src/features/baseplate/utils/pieceFingerprint.test.ts
+++ b/src/features/baseplate/utils/pieceFingerprint.test.ts
@@ -99,6 +99,49 @@ describe('computePieceFingerprint', () => {
     expect(computePieceFingerprint(blCorner)).not.toBe(computePieceFingerprint(trCorner));
   });
 
+  it('treats all-zero cornerRadii as no rounding (edges omitted, no connectors)', () => {
+    // cornerRadii present but every corner 0 → nothing rounds → a corner piece
+    // and an interior piece are geometrically identical and must share a key.
+    const base = makeParams({ width: 3, depth: 3, cornerRadii: { tl: 0, tr: 0, bl: 0, br: 0 } });
+    const corner = {
+      ...base,
+      edges: { left: 'exterior', right: 'join', front: 'exterior', back: 'join' } as const,
+    };
+    const interior = {
+      ...base,
+      edges: { left: 'join', right: 'join', front: 'join', back: 'join' } as const,
+    };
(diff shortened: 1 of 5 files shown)


andymai:
Addressed Copilot in b0968c127 — the corner-rounding fingerprint now keys on `exterior && radius > 0` (matching `buildSlabProfile`/`resolveCornerRadii`) and omits the edge signature entirely when no corner rounds, so explicit all-zero or per-corner-zero `cornerRadii` no longer split identical pieces. Added regression tests.

PR #2224, opened 2026-06-17
perf(baseplate): cross-session IndexedDB cache for split-export pieces

andymai:
## What

Exporting a split baseplate rebuilt and re-tessellated **every unique piece from scratch on every export** — including an identical re-export after a page reload. This persists each unique piece's export bytes in IndexedDB, so a repeat export of the same geometry skips the worker entirely; only cache **misses** go to the pool.

This is the third and last of the export-perf levers (after #2220 build speed and #2221 dedup). Where those cut the cost of *building* a piece, this removes the build entirely for already-seen geometry — a re-export of a 26×26 becomes near-instant.

## Correctness — never serves stale geometry

The cache key is complete by construction:
- **`__GIT_SHA__`** (Vite build-time constant) — any generator change shipped in a new deploy invalidates the whole cache. Within one deploy the pipeline is fixed, so cached bytes are byte-identical to a rebuild.
- **Order-stable serialization of the exact params** passed to `exportBaseplate` — the sole determinant of the worker output — plus the **format** and the **session nozzle size** (which scales connector geometry but isn't carried on the per-piece params object). Including the nozzle explicitly can only ever cause a conservative miss, never a wrong hit.

## Robustness

- **LRU eviction** caps the store at 50MB (monotonic recency stamps make eviction deterministic even within a millisecond).
- **Graceful degradation**: every op is wrapped — if IndexedDB is unavailable (private mode, quota, disabled), reads miss and writes no-op, and the export runs exactly as before.
- Integrated only in the split path's unique-piece generation: cache keys computed → batch read → only misses generated (pool or sequential) → fresh results persisted → assembled in order. Progress reporting counts cached pieces as already-done.

## Tests

`exportCache.test.ts`: key stability under property reordering; key sensitivity to geometry/format/nozzle; put/get roundtrip; batch hits+misses; clear; and LRU eviction over the byte budget (via a test-only cap seam). 541 baseplate tests pass.

## Follow-up

The single-plate (non-split) export path and the connector-key part still rebuild; they're cheaper (one piece) and can adopt the same helper later. The Settings "storage" view could also surface/clear this cache.

### DIFF
--- a/src/core/storage/clearAppData.ts
+++ b/src/core/storage/clearAppData.ts
@@ -16,7 +16,19 @@
 import { clearAllData as clearIndexedDB } from './backends/indexedDB';
 import { pruneAnalyticsData } from '@/shared/analytics/posthog';
 import { clearLabelSizesCache } from '@/shared/analytics/purposeInference';
-import { SETTINGS_STORAGE_KEY } from './storageKeys';
+import { SETTINGS_STORAGE_KEY, BASEPLATE_EXPORT_DB_NAME } from './storageKeys';
+
+/** Delete a whole IndexedDB database, resolving regardless of outcome. */
+function deleteDatabase(name: string): Promise<void> {
+  return new Promise<void>((resolve) => {
+    try {
+      const req = indexedDB.deleteDatabase(name);
+      req.onsuccess = req.onerror = req.onblocked = () => resolve();
+    } catch {
+      resolve();
+    }
+  });
+}
 
 /** Keys that should be preserved during a full data clear. */
 const PRESERVED_KEYS = new Set([SETTINGS_STORAGE_KEY]);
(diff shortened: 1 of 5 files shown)


andymai:
Addressed both review points in eece12c10:

- **`stableStringify` null/undefined collision (Greptile)** — object keys with `null`/`undefined` values are now omitted entirely, so an unset optional param canonicalizes identically whether it's absent, `undefined`, or `null` (e.g. after a JSON/storage round-trip). This matches the generator, which reads every optional baseplate param with nullish semantics (`x ?? default`, `x !== false`, truthiness) — so those forms are byte-identical geometry and *must* share a key. Real values including `false`/`0` are always kept, so a meaningful flag never collapses into "unset". Added tests pinning `absent ≡ undefined ≡ null` and `0 ≢ unset`.
- **Batch timestamp (Copilot)** — `putCachedExports` now stamps each entry with its own `nextTs()`, so intra-batch eviction orders deterministically by recency instead of tie-breaking on key.

PR #2233, opened 2026-06-17
fix(scan): recover eroded-corner cards via min-area rectangle

andymai:
## Problem

Follow-up to #2230 (chroma sweep). A second real photo still failed: a glossy **blue Citi card on dark fabric** reported "No card detected." Reproduced against the merged code.

The chroma sweep now masks the card with its full footprint (bounding box aspect **1.61** ✓), but the card's glossy "citi" logo is a desaturated patch that erodes one **corner** out of the mask. `contourToQuad` fits the quad from 4 extreme contour points, so that one eroded corner drags a vertex inward → a skewed quad → `estimateRectAspect` reads **1.21**, failing the card-aspect gate (needs ≥1.29). A single bad corner sinks an otherwise-correct detection. A glossy/multi-colored card will *always* have regions that drop out of any single threshold, so the detector must tolerate it.

## Fix

Add **`minAreaRect`** — convex hull (monotone chain) + rotating-calipers minimum-area rectangle. Because it's hull-based, it bridges the eroded-corner "pocket" and recovers the card's true rectangle from its surviving straight edges. This is what production scanners use (OpenCV `minAreaRect`, jscanify) — ~90 lines of pure TypeScript, no OpenCV.

Wired as a **strictly additive rescue** in `findBestCardComponent`: the existing inscribed-quad path runs first and wins for every currently-working case (silver-on-wood, tilted shots, the square decoy stay byte-identical). Only when its aspect check fails does the min-area rect try to recover the card — gated on the rect's own aspect **plus a fill ratio** so non-rectangular tools (L-shapes, triangles) can't sneak through. Same additive philosophy as the chroma fix: it can only *recover* misses, never regress a working detection.

## Validation

Ran the detector against **21 diverse real card photos** downloaded from the web (blue/black/gold/silver/metal/holographic cards on wood, fabric, marble, light/dark desks, flat and tilted):

- **~92% detection on representative shots** (a single card lying on a surface with margin — the feature's contract). No regressions; the min-area-rect rescue carried the borderline cases.
- Remaining misses are **out-of-contract** scenes — cards not lying flat, occluded (fingers/wallet/paper), cluttered product shots, or frame-filling flat-lays — which **cannot be sized reliably** (a clipped/occluded card has no recoverable true size). The right response there is guidance, not a wrong measurement.

So the **no-card guidance copy** is reworded from "Add a … card beside your tool" to steer users past the actual failure modes: *lay the card flat, fully in frame, on a plain surface*. (Updated across all 9 locales. The existing steep-angle hint already covers tilted-but-detected cards.)

> Note: I scoped out a "card is cut off" detector that was in the original plan — on investigation the photo that motivated it turned out not to be clipped (it was a cluttered scene), and a border-touching-blob heuristic is fragile (a partially-clipped card often still detects with wrong scale; cluttered scenes risk false "move the card" hints). The copy improvement covers the same intent robustly with zero false-positive risk.

## Tests
- `minAreaRect.test.ts` — hull, axis-aligned + rotated recovery, concave-bite immunity, degenerate input.
- `cardDetect.test.ts` — eroded-corner card: asserts the inscribed quad alone rejects it (aspect out of tolerance) and the min-area-rect rescue detects it.
- `cardRobustness.test.ts` — 9-case matrix: both color polarities (neutral/saturated card × saturated/neutral bg), rotation, eroded-corner rescue.

77 scanTrace tests pass; typecheck, lint, knip, and all 4 i18n checks clean.


andymai:
@greptileai review

PR #2247, opened 2026-06-18
feat(bin-designer): non-bin item foundation + slanted tool rack (labs)

andymai:
## What

Lays the foundation for **non-bin Gridfinity items** in the bin designer — things that sit on a baseplate but aren't walled bins — and ships the first one: an **angled-fin tool rack** (for pliers, tweezers, hand tools). Behind the experimental Labs flag `item_kinds` (off → zero behavior change).

## Architecture

A general `GridfinityItem = { envelope, structure }`:
- **envelope** = concerns shared by every baseplate item (footprint, grid/height units, magnet/screw attachment, colors)
- **structure** = a discriminated union keyed by `kind` (`bin` | `toolRack`)

A **bin is registered as one item kind**, but its existing flat-`BinParams` generation/persistence path is left **physically intact** via pure round-trip adapters (`binParamsToItem`/`itemToBinParams`, identity-tested). New kinds plug in via three boundary-respecting registries (descriptor = shared schema/defaults/migrate/labels; generator = worker/OCCT; UI = designer/React), so the worker never imports React.

Worker dispatch is a generic `GENERATE_ITEM`/`EXPORT_ITEM` message resolved by `structure.kind` — adding a future item type needs no new message.

## Scope
- **shared:** item types, adapters, descriptor registry
- **generation:** `toolRackGenerator` (fins + reused socket), generic dispatch, bridge `generateItem`/`exportItem` with its own dedup cache
- **designer:** kind-dispatched `ParameterPanel`, `ToolRackParameterPanel` (live preview + STL/STEP export), `PreviewCanvas` gates bin-only overlays
- **persistence/store:** `SavedDesign`/`DesignerState` gain optional `kind`/`envelope`/`structure` (bins byte-identical); kind-aware storage
- **safety:** every `BinParams`-coupled consumer (thumbnails, share, design-linking) guards on kind so a rack can't crash/corrupt them
- **i18n:** real translations across all 10 locales

## Verification
- typecheck: 0 errors · lint: clean (no boundary violations — worker pulls no React) · i18n parity + value checks pass
- tests: 6 real-WASM scenario + adapter/registry/store/UI unit tests — all green

## Deferred (follow-up)
- Rack **autosave** (`useAutoSave` watches `params`, static for racks → inert, not broken)
- Folding rack export into the main header Export button (rack currently exports via its own panel buttons)


andymai:
Thanks for the review — addressed in bfeace7e7:

1. **`finCount` schema min 1 vs generator floor of 2** — schema min is now `2` (matches the generator's `Math.max(2, …)` and the panel stepper), so a persisted count is respected rather than silently bumped.
2. **Panel shows `finCount ?? 6` but generator auto-derives from `slotPitch`** — `DEFAULT_TOOL_RACK_STRUCTURE` now sets an explicit `finCount: 6`, so the panel reading and the first preview/export agree (the generator only auto-derives when `finCount` is omitted).
3. **Misleading `loadDesign` comment** — reworded to state that structure migration runs later in the designer store; this layer returns the row unchanged.
4. **`saveDesign` could persist invalid shapes** — added an up-front guard returning a `storageCorrupted` error when a bin design lacks `params` or a non-bin design lacks `envelope`/`structure`.

Also fixed the failing `features.test.ts` check by giving the experimental medium-risk `item_kinds` flag the required `warning`.

PR #2255, opened 2026-06-18
fix(lid): size lids to the overhang-expanded bin body

andymai:
## Problem

Bins with an **overhang** (per-side outward expansion that fills a drawer's centering gap) grow their outer body **and stacking lip** outward, but the lid was still sized from the nominal footprint (`width*42 - clearance`) and ignored `params.overhang` entirely. The result: the lid came out too small and wouldn't seat over the bin.

## Fix

- **`lidInputs.ts`** — `resolveLidInputs()` now resolves the overhang and grows `lidOuterW/D` by the expansion (`addW/addD`), and exposes the asymmetry shift as new `outerOffsetX/Y` fields. Overhang is suppressed for polygon (`cellMask`) bins, mirroring `boxBuilder`.
- **`lidProfile.ts`** — `buildOutlineDrawing()` translates the outline by the offset, so the floor, mating shell, and stack-grid slab all wrap the expanded/shifted body from one place.
- **`lidClickRail.ts`** — rectangular rail placements shift by the same offset so rails ride the actual perimeter.

### Two coordinate families
Perimeter-anchored geometry (outline, mating shell, floor, click rails) follows the overhang-expanded/shifted body. Grid-anchored features (stack-grid pockets, magnet holes) stay pinned to the nominal socket grid at the origin — overhang deliberately leaves the base sockets at nominal footprint, so a stacked bin still mates on the standard gridfinity grid.

## Tests

Three new scenario tests in `lidGenerator.scenario.test.ts`:
- Symmetric overhang grows the lid footprint (+10mm width / +8mm depth).
- Asymmetric overhang shifts the perimeter (heavy side reaches further; opposite edge fixed).
- Polygon bins suppress overhang (identical mesh with/without).

Typecheck clean; full unit + dom suite green (13872 passed / 0 failed).


andymai:
Addressed both review points in 2cd8cba4:

- **Two-coordinate-family invariant (greptile P2):** added a Gotchas entry to `src/features/generation/README.md` (#7) explaining that perimeter-anchored lid features follow `outerOffsetX/Y` while grid-anchored features stay on the nominal socket grid, so future lid work picks the right frame.
- **Polygon-suppression test (Copilot + greptile):** strengthened the assertion with a bounding-box comparison alongside `triangleCount`, so a stray overhang shift/scale on a polygon lid would move an edge and fail the test.

Full unit + dom suite green (13872 passed / 0 failed).

PR #2279, opened 2026-06-20
feat(bin-designer): choose which side the half-unit foot sits on

andymai:
## What

Adds `fractionalEdgeX` / `fractionalEdgeY` to `BinParams` so a bin with a fractional dimension (e.g. a 2.5×2 bin) can place its half-unit foot on **either** side, mirroring the baseplate's existing fractional-edge option. Default `'end'` preserves current geometry.

## Why

From discussion #2271: with a 2.5×2 bin the half-unit segment is hardcoded to one side. The only workaround — print and rotate — fails because the finger scoop always generates on the front face, so rotating moves the scoop to the back. Letting users pick the half-foot side removes the need to rotate, keeping the scoop where they want it.

## How

The underlying `forEachCell` already supported the reversal; this PR propagates one setting to every cell iterator that must agree on foot placement:

- base sockets **and** their magnet/screw holes (`socketBuilder`)
- lightweight base cups + pad/drill loops (`lightweightBaseBuilder`)
- lid magnets, via `LidInputs` (`lidInputs` → `lidMagnets`)
- both socket cache keys (`socketCacheKey`, `baseSocketShapeKey`) — the `frac:x:y` token is appended **only when non-default**, so existing `'end'` designs keep byte-identical keys
- API share allowlist (`ALLOWED_PARAM_KEYS`) so the setting survives cloud share/sync
- UI: reuses `FractionalEdgeToggle` in `DimensionsSection`, shown only when an axis is fractional (no new i18n strings)
- `migrateParams` backfills `'end'` for existing designs

### Scope note
Overhang gap-fill feet keep the default `'end'` tiling — they don't mate with baseplate sockets, so only a cosmetic seam differs at a fractional edge, never socket mating. Documented in a code comment.

## Testing

- `tsc -b` clean, ESLint 0 errors
- New tests: foot placement (`forEachSocketCell`), cache-key behavior, migration defaults, UI visibility/wiring
- Existing `fractional-feet` scenario still passes

Closes #2271

### DIFF
--- a/api/lib/designerValidation.ts
+++ b/api/lib/designerValidation.ts
@@ -90,6 +90,8 @@ const ALLOWED_PARAM_KEYS = new Set<string>([
   'width',
   'depth',
   'height',
+  'fractionalEdgeX',
+  'fractionalEdgeY',
   'gridUnitMm',
   'heightUnitMm',
   'wallThickness',
--- a/src/features/bin-designer/README.md
+++ b/src/features/bin-designer/README.md
@@ -81,6 +81,21 @@ graph TB
   `paramsNeedHalfGridMode` (fractional dimensions OR `hasHalfBinDetail(mask)`),
   so reopening a design or undoing past a dimension change never leaves the
   UI toggles out of sync with the underlying shape.
+- **Fractional foot edge** (`params.fractionalEdgeX` / `fractionalEdgeY`): for a
+  bin with a fractional dimension (e.g. 2.5u), which side the half-unit foot
+  column/row sits on — `'end'` (default) = right/back, `'start'` = left/front.
+  Mirrors the baseplate's drawer-level option and lets a 2.5×2 bin place its
+  half foot on either side **without** rotating the print (rotation would move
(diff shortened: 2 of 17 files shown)


andymai:
Addressed review feedback in 65d3feaf:

- **Swap dimensions (Copilot):** `handleSwapDimensions` now swaps `fractionalEdgeX`/`fractionalEdgeY` alongside `width`/`depth`, so the half-foot preference travels with its axis. Added a test.
- **half-sockets mode (Greptile + Copilot):** in `base.halfSockets` mode every foot is a uniform 0.5u cell, so the edge setting has no geometric effect — the `DimensionsSection` toggle now hides in that mode, and the `globalHalfSockets` branch in `forEachSocketCell` carries a comment explaining the intentional omission. Added tests for both the hook flag and the rendered control.
- **Docs (Greptile):** documented `fractionalEdgeX`/`fractionalEdgeY` (incl. the half-sockets interaction) in `bin-designer/README.md`.

Issue #2346, opened 2026-06-24
perf(generation): adopt brepjs instanceGrid/materialize (gridPattern) for the repeated-cell grids


andymai:
The bin socket grid and the baseplate pocket grid currently hand-roll the array: clone/translate a cell template per position, then `fuseAllBisect` / `cutInBatches` (`socketBuilder.ts`, `pipeline/stages/booleanStage.ts`, `baseplateGenerator.ts`, `baseplatePockets.ts`). The per-cell-loft was fixed in #2332; this is the next-level grid optimization.

**Now newly actionable:** brepjs **18.102.0** ships `instanceGrid(cell, { cols, rows, pitchX, pitchY })` + `materialize({ fuse: true })`, which routes a grid through the kernel's `gridPattern` — a **bulk-copy + nested row/column fuse** instead of N app-space translates + one flat N-way fuse. (The tool is currently on brepjs 18.100.3.)

**Evidence it's worth it:** the brepjs A/B benchmark (andymai/brepjs#1612) measured gridPattern **~20–25% faster** than hand-rolled translate+fuse at 36–64 cells under occt-wasm, and the gap grows with grid size. Notably the win is the **nested-fuse structure, not `sameFace` glue** — under occt-wasm `+sameFace` ≈ plain `fuseAll`, so the tool's reliance on `commonFace` glue isn't the lever here.

## Proposed

- [ ] Bump brepjs to **≥ 18.102.0**.
- [ ] **Bin socket grid**: A/B `materialize(instanceGrid(cell, …), { fuse: true })` (or `gridPattern` directly) vs the current template-clone + `fuseAll({commonFace})`. Adopt the winner; delete the hand-rolled array + fuse where it wins.
- [ ] **Baseplate pocket grid**: same A/B against the `cutInBatches` / `pocketTemplate` path.
- [ ] Keep the existing `*.bench.ts` timing as the gate; measure on **both kernels**.

## Caveats

- **brepkit** returns an **unfused compound** from native `gridPattern` (near-free, but not a single solid). For the fused *export* path under brepkit, verify the result is actually fused (or force the fuse) — occt-wasm fuses. This is exactly the apples-to-oranges the brepjs benchmark flagged.
- This composes with #2334 (direct-mesh draft preview) and the now-shipped #2332/#2333 — it's the fused-export grid path, not the preview.

Part of #2335.

andymai:
[...]

PR #2353, opened 2026-06-24
test(generation): cold-cache benches for socket + baseplate grid builds

andymai:
## What

Adds larger-grid, **cold-cache** benchmarks for the two repeated-cell grid builds:

- **Bin socket grid** (`buildBaseSocket`): 4×4 / 6×6 / 7×7 preview, 6×6 export, 6×6 export+magnets.
- **Baseplate pocket grid** (`generateBaseplate`): 6×6 / 8×8 / 12×12 cold, 8×8 export cold.

## Why

The existing end-to-end benches measured these grids **warm**:
- the whole-socket cache isn't keyed on bin height, so repeated `generateBin` calls hit it after the first iteration;
- the baseplate `paddingLeft` cache-defeat trick quantizes away in the cache key.

Clearing the relevant caches each iteration times the real **cold** grid build (cell loft + fuse + hole/pocket cut) across sizes — the lever for any future grid-fuse change. No production code changes.

## Context

Spun out of investigating #2346 (adopt brepjs `gridPattern` for these grids). The optimization itself measured as a **regression** on occt-wasm and was not adopted — see the findings on #2346 — but the cold-cache bench coverage I added while measuring is independently useful, so it lands here.

## Test

- `pnpm exec vitest bench binGenerator.bench -t "socket grid"`
- `pnpm exec vitest bench baseplateGenerator.bench -t "cold"`
- Both kernels via `BREPJS_KERNEL=wasm`.

### DIFF
--- a/src/features/generation/worker/generators/baseplateGenerator.bench.ts
+++ b/src/features/generation/worker/generators/baseplateGenerator.bench.ts
@@ -106,6 +106,9 @@ describe('baseplate generation', () => {
   // Large-plate cold builds. The unique-paddingLeft cache-defeat above
   // quantizes away for these, so clear the baseplate caches each iteration to
   // time the real cold pocket cut across grid sizes; no magnets ⇒ through-cut.
+  // The clear must live INSIDE the timed fn — vitest's bench() ignores
+  // tinybench's per-iteration beforeEach and runs setup once per phase, so a
+  // hook would leave every iteration after the first warm.
   const coldPlate = (w: number, d: number, forExport: boolean) => () => {
     clearBaseplateCaches();
     getGenerateBaseplate()(defaults({ width: w, depth: d }), noop, forExport);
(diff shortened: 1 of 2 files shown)


andymai:
@copilot Good catch on the cache-clear being inside the timed region — but I verified empirically that moving it to a hook would break these benches. vitest's `bench()` **ignores** tinybench's per-iteration `beforeEach` (probe: `beforeEach` fired 0 times) and runs `setup`/`teardown` only **once per phase** (warmup + run = 2 calls), not per iteration. So a hook would clear once → first iteration cold, every subsequent iteration **warm** (cache hit) — silently defeating the cold-build measurement.

Clearing inside the timed fn is the only way to keep every iteration cold with vitest's bench API. The disposal cost is sub-ms vs 45–12,000 ms builds and, since it's constant, cancels out in the relative A/B these benches exist for. Added a code comment (c12e6dd) documenting the constraint so it isn't "fixed" into a warm bench later.

PR #2390, opened 2026-06-29
refactor(baseplate): replace margin-fill segmented control with a toggle

andymai:
The baseplate margin-fill control was a 3-way segmented toggle (**Padding / Grid / Half-grid**) that read as cramped, and the wording was unclear. This swaps it for the panel's standard **toggle + sub-option** pattern.

### Before → After
- **Before:** segmented control with three mutually-exclusive segments.
- **After:** a `FeatureToggle` switch **"Fill padding with grid tiles"** (off = solid padding, on = grid fill), with a nested **"Use half-grid cells"** checkbox revealed only when the toggle is on.

This collapses the three states into the common two and matches the magnet toggle's layout/behaviour. No change to behaviour or the underlying `overTile` / `overTileHalfGrid` params.

### Copy
- Toggle label `Fill padding with grid` → `Fill padding with grid tiles` (consistent with the existing "grid tiles" hint copy).
- Reworded the half-grid hint so it no longer references the now-removed "Grid mode".
- Removed the three unused `marginFill*` keys; added `baseplate.useHalfGrid`. All 10 locales updated; parity/values/interpolation pass.

### Visual
Verified via a throwaway Playwright capture:
- Off: just the `Fill padding with grid tiles` switch.
- On: switch + indented `☑ Use half-grid cells` + hint.

### Tests
- Rewrote the `BaseplatePanel` "margin fill mode" suite for the switch + checkbox (64 pass). The half-grid checkbox is a sub-option, so the "enable half-grid" case starts from fill-on.
- Updated the `baseplate-overtile` e2e to drive the switch + checkbox instead of the radios.

> Note: the copy is open to refinement — I picked clear wording but happy to adjust the toggle/checkbox/hint text to taste.

---
## Summary by cubic
Replaced the 3-way margin-fill control with a simple toggle plus a half-grid sub-option to match the panel pattern. Also fixes an edge case where the toggle could get disabled when padding was too small.

- **Refactors**
  - Swapped the segmented control for a `FeatureToggle` “Fill padding with grid tiles,” with a nested “Use half-grid cells” checkbox shown only when enabled.
  - Updated copy and hints; added `baseplate.useHalfGrid`; removed unused `marginFill*` keys across 10 locales.
  - Updated unit and e2e tests to drive the switch + checkbox.

- **Bug Fixes**
  - Kept the toggle reachable when padding shrinks: derive state from stored `overTile`/`overTileHalfGrid` flags, only block turning it on when no edge can fit tiles, and show the “too small” note; added a test for this case.

<sup>Written for commit 97c61f0a78d989fabbb567e0a1d3ae61f56dccaa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2390?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Fixed the P2 (Copilot + cubic) in 97c61f0a: the toggle now derives its checked state from the stored `overTile`/`overTileHalfGrid` flags rather than the `canOverTile`-clamped mode, and only *turning it on* is blocked when padding is too small — turning it **off** stays possible (with the "too small" note shown inline). Added a regression test for the fill-on + shrunk-padding case. Also fixed the French P3 (`tuile pleine` → `tuile entière`).

PR #2398, opened 2026-06-30
feat(baseplate): detach margins into separate printable pieces (#2392)

### LINKED ISSUE #2392: Baseplate Margin's to filler plate
**Problem**
Margins of bases plates are currently apart of the base plate.

**Proposed Solution**
offer the option to separate the margins into separate pieces and/or add on connection type if selected.  

**Alternatives Considered**
Adding a separate margin/spacer creator?

andymai:
Closes #2392 (PR 1 of 2).

## What
Adds an opt-in **Detach margins** mode: the drawer-fit padding prints as its own piece(s) instead of part of the baseplate, so a misprinted margin can be reprinted on its own without scrapping the whole plate. This unifies both asks on the issue — separate margin pieces (jldesign) and detachable/reprintable margins (DannyDelott).

This PR ships the **default detached-without-connector** behavior. An opt-in connector between rail and body is the planned follow-up (PR 2).

## How it works
- **Planner** (`emitMargins`): decomposes the padding into butt-joint **rails** — long rails own the plate corners, short rails fit between. Every rounded outer corner the integral plate has is carried by exactly one rail, across all padding patterns (all-four, opposite-pair, adjacent, single-side, asymmetric). Runs independently of split state, so a bed-fitting (unsplit) plate detaches too.
- **Geometry** (`buildMarginSolid`): a rail is a `length × band` slab built centered at the origin (it can't go through `buildBaseplateSolid` — that path rejects zero-grid pieces). Rails reuse the body's frame-cell pockets (filtered to each rail) and corner profile, so over-tile and half-grid margins keep their tiles and the assembled corner matches the integral plate.
- **Body** prints padding-free on detached sides (both the unsplit body and split body pieces).
- **Preview**: rails render at their world offset in every mode (split or not), with an outward explode gap.
- **Export**: rails ship as separate files in the ZIP (STL/STEP/3MF) with a print guide listing each rail.
- **Mutually exclusive with stack-print** (stack-print wins); UI disables the toggle with a reason.

## Design decisions (confirmed up front)
- Default detached, **no connector** (connector is PR 2).
- Butt-joint rails; **all three fill modes** detach.
- stack-print wins on conflict; `MARGIN_MIN_DETACH_MM = 8mm` (reuses the over-tile threshold).

## Architecture note
The rail-mesh path adds a `GENERATE_BASEPLATE_MARGIN` / `EXPORT_BASEPLATE_MARGIN` message that rides the existing mesh/export response unchanged. `MarginPiece` moved to `shared/types` so the generation worker can build a rail without a cross-feature import.

## Tests
- Planner: corner-ownership across every padding pattern + unsplit emission + body padding-zeroing.
- Geometry: BREP scenario tests per fill mode (solid / over-tile / half-grid), long + short rails, magnet height — all mesh to non-degenerate solids with correct bounding boxes.
- Migration (stack-print precedence), print guide, MarginMeshes, plus the full baseplate + generation suites (2569 tests) green. Typecheck + knip + i18n (10 locales) clean.

## Known minor gaps (deferred)
- Rails aren't counted in `SplitViewStrip` / `bedLoads` / analytics `pieceCount` (they export and render correctly, just aren't tallied in those split-only UIs).

---
## Summary by cubic
Adds an opt‑in Detach margins mode that prints drawer‑fit padding as separate rail pieces so you can reprint a bad margin without reprinting the whole plate. Rails now segment per body piece for bed fit and explode with their piece; detached edges square the body’s corners to meet the rail cleanly. Closes #2392.

- New Features
  - Toggle: “Detach margins” in the Baseplate panel. Sides detach only when padding ≥ 8 mm; disabled while stack‑print is on (stack‑print wins) and reflects the stored opt‑in as on‑but‑disabled.
  - Planner: emits rail segments per outer body piece (long rails own corners; short rails fit between). Works for split and unsplit plates.
  - Body: prints padding‑free on detached sides; body corners on detached edges are squared so the rail owns the outer rounding; rails reuse frame‑cell pockets and corner radii so over‑tile/half‑grid align.
  - Preview: rails render at their world offset, explode with their piece plus a small outward gap; never alongside the stack preview.
  - Export: rails ship as separate STL/STEP/3MF files in a ZIP, plus a margin guide listing each rail with dimensions; single‑body exports become ZIPs when rails are present.
  - Generation/Bridge: adds `generateMargin`/`exportMargin` and worker messages `GENERATE_BASEPLATE_MARGIN` / `EXPORT_BASEPLATE_MARGIN`.

- Bug Fixes
  - Export: body now exports padding‑free on detached sides (STEP/STL/3MF), matching preview and split exports; cache keys use the padding‑free params.
  - Migration/UI/Preview: preserves the stored detach‑margins opt‑in; the toggle shows as on‑but‑disabled under stack‑print; rails hide in stack preview and honor the xray toggle; panel explains when margins are too small to detach.
  - Rails: extend over integral perpendicular padding so outer corners are fully covered; fixes a small corner gap on mixed‑threshold layouts.

<sup>Written for commit 2f97950e597d1649be9fe1b760c7bc511524f839. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thanks @copilot — addressed in 832960d6c:

- **Export double-counted padding (the real bug):** the unsplit STEP and STL/3MF branches now export the body via `bodyParamsForDetach(fullParams)` (and key the export cache on it), so the body prints padding-free and matches both the preview and the split path. Good catch.
- **Scenario tests:** switched to `forExport=false`, the path production actually uses (`handleGenerateBaseplateMargin` / `exportMargin`).
- **Docs:** reattached the `emitMargins` doc block that had been orphaned onto `detachedSides`.

andymai:
Thanks @cubic-dev-ai — triaged all 12. Addressed in 7f0bf6173:

**Fixed**
- **P2 migration** — preserve the stored `detachMargins` opt-in; mutual exclusion with stack-print stays a runtime concern (so turning stacking off restores intent).
- **P2 stack preview** — rails no longer render while stack-print is active (gated on `stackEnabled`).
- **P3 panel** — shows the "too small to detach" explanation when the toggle is on but no side meets the threshold.
- **P3 xray** — rails honor the xray toggle (transparent like body/split).
- **P3 print guide** — rail dimensions at 0.1mm precision (matches the piece table).
- **P3 store** — `MarginMeshEntry.side` derives from `MarginPiece['side']`.
- **P3 i18n** — fixed uk ("друк стосами", was "пошаровий друк") and nl ("gestapeld printen") conflict messages.
- **P3 test** — magnet-height assertion checks the full Z span, not just the top.

**Already addressed before your run (P1, conf 9):** the STEP/STL detached body export uses `bodyParamsForDetach` as of 832960d6c — your run reviewed the pre-fix diff. The draft path also already clears `marginMeshes` (line 342), so no stale rails during the draft window.

**Deferred (noted):** P3 on `generateHandler.ts:92` — margin generations feeding the shared adaptive-debounce window. Real but niche; it touches the shared worker timing path and is better as a small standalone follow-up than bundled here.

andymai:
Thanks @copilot — second pass triaged:

**Fixed (310a9aa5f) — the corner-ownership comments (splitPlanner.ts:558/566/574/585):** all four describe the same real gap — when a side detaches but a perpendicular side has nonzero *sub-threshold (integral)* padding, the rail stopped at the grid edge and left an uncovered notch at the shared corner. Rails now span the full outer extent: long rails extend over any perpendicular padding, and short rails extend over a perpendicular side's padding when that side isn't itself a long rail. The rail sits outside the grid on its band axis while the body's integral padding sits inside it, so they abut without overlap. Added a regression test; all-four/opposite/adjacent/single-side layouts are unchanged.

**Not a bug — `baseplateMargin.ts:178` (`forExport=false`):** this is intentional and matches the body export. `exportBaseplate` also builds with `forExport=false` — the full-detail multi-section loft (`forExport=true`) produces BREP topology OCCT can't reliably tessellate/export (see the comment on `exportBaseplate`), and it's geometrically equivalent for printing. Using `false` here keeps the rail export on the same proven path as the plate; `true` would be the unexportable one.

andymai:
Thanks @copilot:

- **`BaseplatePanel.tsx:241` (toggle hides stored opt-in)** — fixed in 2f97950e5. The switch now reflects the stored flag, so when stack-print suppresses detach it reads as on-but-disabled (with the stack-print reason) instead of off.

- **`baseplateMargin.ts:177` (`forExport=false`)** — not a divergence: the baseplate export also builds with `forExport=false` (`exportBaseplate` → `buildBaseplateSolid(params, false)`), so the rail and the body use the **same** simplified pocket cutter on export — they match. `forExport=true` is the full-detail loft that OCCT can't reliably export, so it's intentionally avoided on both paths. No change.

Issue #2439, opened 2026-07-03
Shadow Board style Multi-Color Support


jnierodzik:
**Problem**
Creating shadow board/box style bins using the Custom Cutouts feature
Reference image - https://i.ytimg.com/vi/K45Y8rKlYDY/hqdefault.jpg?sqp=-oaymwEmCOADEOgC8quKqQMa8AEB-AH-CYAC0AWKAgwIABABGGQgZChkMA8=&rs=AOn4CLCzIizefoR2eX6qrXE6ua--Wqh9_w

**Proposed Solution**
You would be able to assign a color to a cutout in the editor. Ideally you could select wether to assign a color specifically to just the floor of the cut-out or the floor + interior walls.

**Alternatives Considered**
None

PR #2440, opened 2026-07-04
feat(bin-designer): shadow-board colors for custom cutouts

andymai:
Give any custom cutout its own filament color, so a tool lifted out of a shadow-board bin reveals a contrasting silhouette underneath — and the whole board still prints as one multi-material 3MF.

Implements #2439.

## What you get

In the Cutout Editor, each cutout gains a **Color** section: a toggle, the shared color picker, and a **Floor** vs **Floor + walls** choice.

- **Floor + walls** paints the whole cavity (the classic full-contrast recess).
- **Floor** paints only the bottom, leaving the walls body-colored (a literal "shadow backing").
- Grouped (boolean) cutouts share one color; parametric-array instances inherit their master's.
- Recoloring is **instant** — it repaints from the existing mesh without regenerating geometry.

Colors flow through both the 3D preview and the 3MF export (per-piece, so oversized boards that split across the print bed still color correctly).

## How it works

Each cutout cavity is stamped at generation time with a per-unit face tag (`CUTOUT_COLOR_TAG_BASE + ordinal`, from a single shared `enumerateCutoutColorUnits` used by both the worker and the paint layer). At paint time the tag identifies the cutout and the triangle's normal decides floor vs wall. Because every cutout is tagged regardless of color, turning a color on/off is a pure read-side change — the setter is cosmetic and never bumps the generation epoch.

Cutout colors occupy material slots appended after the fixed `ZONE_ORDER`, so existing zone indices (and hover-glow) are untouched.

## Notes

- Extends the existing (graduated) Multi-Color feature; setting a cutout color auto-enables multi-color output.
- Server-side validation rejects a non-hex color or unknown scope on both share and sync (cutout color reaches exported 3MF material colors).
- i18n added across all 9 locales.

## Testing

- Store: group-spread, auto-enable, cosmetic-epoch, undo (unit).
- Worker: real-WASM scenario — distinct per-cutout tags with floor + wall faces, group collapse, uncolored-still-tagged.
- Paint: preview + export material mapping (floor+walls vs floor-only vs single-color gating), verified end-to-end on the real kernel.
- UI: inspector controls (unit).
- Live visual check in the app: toggling color turns the cavity red; Floor scope paints only the bottom, walls stay body-colored.

---
## Summary by cubic
Add per-cutout shadow-board colors so removing a tool shows a contrasting silhouette in preview and in exported 3MF, without regenerating geometry. Implements #2439 and auto‑enables multi‑color when a cutout is colored.

- **New Features**
  - Cutout editor: new Color controls (toggle, shared picker, Floor vs Floor + walls). Works with multi-select.
  - Grouping + arrays: grouped cutouts share one color; array instances inherit their master.
  - Instant recolor: reads baked face tags; no geometry regen; undo supported. Labels stay in the text zone.
  - Preview + 3MF: paints floor-only or floor+walls per triangle normal; material slots are appended after fixed `ZONE_ORDER`.
  - Worker tagging: each cavity is stamped `CUTOUT_COLOR_TAG_BASE + ordinal` (shared enumeration); uncolored cutouts are still tagged for instant on/off.
  - Validation + i18n: server rejects invalid hex colors, unknown scopes, or non-object cutouts on share/sync; strings added across all locales.

- **Bug Fixes**
  - Multi-object 3MF export (bin + lid/dividers) now paints cutout colors, matching the single-piece export path.
  - Correct floor vs wall detection in split‑lip 3MF exports so the Floor scope paints correctly.
  - Locked cutouts: the color picker is fully disabled (inert) so it can’t be focused via keyboard.

<sup>Written for commit d75a1b0606239c4b05743c15bab6c7cf7d2a287e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2440?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thanks for the reviews — addressed all findings across two commits (`d3467ad2b`, `8a247639d`).

**Correctness (the two user-visible export bugs Greptile flagged):**

- **Multi-object 3MF export dropped cutout colors** (bins with a lid/dividers). `buildMultiObject3MF` now widens its gate with `anyCutoutColored` and passes `enumerateCutoutColorUnits`, mirroring the single-piece path.
- **Floor scope broke in export on a split lip grid.** `materialMapping` was cross-producting three identical vectors from the `faceNormals` buffer (always ~0). It now reads normal-z directly from that buffer on the split path (as the preview already did), and derives it from `vertices` only on the in-place path.

**Quality / hardening:**

- Skip normal math for non-cutout triangles (check the tag ordinal first) in both preview and export.
- `CutoutColorControls` is now keyed by cutout id, so the picker popover resets when switching cutouts; the color row is gated by `disabled`.
- Cutout color validation is hex-only now (rejects legacy `slot1…4` ids — this is a new field with no migration path).
- Server: reject non-object `cutouts` entries; moved the `validateDesignerShare` JSDoc back above its function.
- Dropped an unverified "TEXT tag preserved" bullet + its dead counter from the tag scenario test.

All tests, typecheck, lint, and i18n checks pass locally.

Issue #2241, opened 2026-06-17
baseplate


tbarritt:
baseplates with dovetails. The dovetails are not a tight fit.

andymai:
Thanks for the feedback. I agree that could lock together better.

PR #2484, opened 2026-07-08
fix(baseplate): export puzzle connector fit sample as puzzle, not dovetail

### LINKED ISSUE #2481: Unable to export puzzle print fit example
If I try to export the "puzzle print fit example" it is exported a dovetail instead of a puzzle

andymai:
## What

Selecting the **Puzzle** connector and exporting the baseplate **Print fit sample** produced a *dovetail* calibration tray instead of a puzzle one. Fixes #2481.

## Why

The connector fit-sample tray (`connectorSample.ts`) predates the Puzzle style (added in #2241 as its own `connectorStyle`). Its `SAMPLE_STYLES` table and union type only listed `dovetail` / `dovetailKey` / `snapClip`, and the builder degrades an unknown style to `SAMPLE_STYLES[0]`:

```ts
const matched = SAMPLE_STYLES.filter((s) => s.key === selectedKey); // [] for 'puzzle'
const styles = matched.length > 0 ? matched : [SAMPLE_STYLES[0]];    // → dovetail
```

So a maker calibrating a puzzle-jointed split baseplate got the wrong profile — useless for dialing in the fit.

## Fix

- Add `puzzle` as its own integral tongue/groove row in the sample tray, reusing the real plate geometry builders (`makePuzzleTongue` / `makePuzzleGroove`, now exported) so the printed coupon carries the **exact** female groove a full split baseplate produces. Puzzle shares the dovetail's base groove clearance; the fit offset rides on the groove.
- Corrected a stale comment on `ConnectorSampleButton` (it claimed the tray swept "all three" styles; it builds one row for the selected style).

Verified the sibling **stack** "Print fit sample" is unaffected — it strips all connectors for uniform tiles, so no style can leak.

## Tests

Extended `connectorSample.scenario.test.ts` (real OCCT kernel):
- piece-count assertion now covers `undefined` / `dovetail` / `puzzle` (all 10 pieces, no loose part)
- watertight/bed-resting STL export now runs for both dovetail **and** puzzle (the filleted lobe is the boolean-risky part)

All 6 scenario cases pass; `typecheck` and `eslint` clean.

---
## Summary by cubic
Fixes the baseplate “Print fit sample” for the Puzzle connector so it exports a puzzle calibration tray, not a dovetail. Adds puzzle to supported styles and reuses real plate geometry for accurate fit tuning.

- **Bug Fixes**
  - Added `puzzle` to sample styles and built puzzle tongue/groove using exported `makePuzzleTongue` and `makePuzzleGroove`.
  - Kept dovetail-equivalent base groove clearance; fit offset applies on the groove.
  - Expanded tests to cover `undefined`/`dovetail`/`puzzle` piece counts, watertight STL export for dovetail and puzzle, and a 1.0 mm nozzle edge case for puzzle (near-collapsed neck).
  - Fixed a stale comment in `ConnectorSampleButton` about which styles the tray sweeps.

<sup>Written for commit 33f3ab324a23968c5945ab2c007021aa9776e352. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2484?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/baseplate/components/BaseplatePanel/ConnectorSampleButton.tsx
+++ b/src/features/baseplate/components/BaseplatePanel/ConnectorSampleButton.tsx
@@ -2,7 +2,7 @@
  * "Print fit sample" control for the baseplate connector section.
  *
  * Opens the shared ExportDialog to download a one-file calibration tray that
- * sweeps all three connector styles across a fit-offset ladder, so makers can
+ * sweeps the selected connector style across a fit-offset ladder, so makers can
  * dial in the fit that clicks before committing to a full split baseplate.
  */
 
--- a/src/features/generation/worker/generators/baseplateConnectors.ts
+++ b/src/features/generation/worker/generators/baseplateConnectors.ts
@@ -565,7 +565,7 @@ function puzzleOutline(
  * cross-section, so the protrusion prints as a self-supported prism with no
  * overhang — in either orientation, so it stack-prints cleanly too.
  */
-function makePuzzleTongue(
+export function makePuzzleTongue(
   pt: (wall: number, bp: number) => [number, number],
   w: number,
   bp: number,
@@ -578,7 +578,7 @@ function makePuzzleTongue(
 
 /** Puzzle groove (female): the tongue outline grown by `cl` on every face, carved
  *  in `−d` (into the piece) and extended beyond the wall and in Z. */
-function makePuzzleGroove(
+export function makePuzzleGroove(
   pt: (wall: number, bp: number) => [number, number],
   w: number,
   bp: number,
--- a/src/features/generation/worker/generators/connectorSample.scenario.test.ts
+++ b/src/features/generation/worker/generators/connectorSample.scenario.test.ts
@@ -146,6 +146,27 @@ describe('connectorSample — fit-sample tray', () => {
   );
 
   it(
+    'exports a watertight puzzle tray at max nozzle (near-collapsed neck)',
+    async () => {
+      // Worst case for the puzzle groove: the 1.0mm nozzle grows the base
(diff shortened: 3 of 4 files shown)


andymai:
Re: Greptile's *Puzzle neck can collapse* finding (`connectorSample.ts`) — addressed in 33f3ab3 with a regression test rather than a geometry change, because the collapse can't actually occur within the enforced bounds:

- Nozzle is clamped to a **1.0mm max**, and the sample offset ladder tops out at **+0.10**.
- Worst case: `scaleClearance(TONGUE_CLEARANCE=0.15, 1.0) = 0.45`, then `effectiveClearance = 0.45 + 0.10 = 0.55`.
- So `nP = max(0, PUZZLE_NECK_PROTRUSION − clear) = max(0, 0.60 − 0.55) = 0.05mm` — the neck narrows but never reaches zero, and brepjs clamps the 0.3mm armpit fillet on the short segment.

The new test `exports a watertight puzzle tray at max nozzle (near-collapsed neck)` builds the puzzle tray at nozzle 1.0mm (exercising the +0.10 column) and asserts the export is **watertight and manifold** (0 non-manifold edges, 0 boundary edges, no NaN). It passes — confirming the `puzzleOutline` clamp keeps the geometry valid, as its comment states. This matches the shared production baseplate builder, so the sample now mirrors real-plate behavior at the extreme too.

PR #2486, opened 2026-07-08
feat(bin-designer): add extra lid height for tall lids over short bins

### LINKED ISSUE #2482: New lid variant
**Problem**
The existing lid works but if I have a bin for toothpicks the bin needs to have the height of the toothpick and the lind only closes the bin, to access the toothpicks I need to try to snag them with my fingers inside the bin. 

**Proposed Solution**
The solution to this niche problem would be to have a bin of 60% of the height of the toothpick and a lid that has 40% of the remaining height, so when the bin is open the toothpick is sticking out of the bin for easy pickup. This could be a new lid variant, keeping the bin logic the same, but adding a modifier for the lid, something like "percentage/mm height" for the lid, similiar to what wall cutout has. Also a toggle to have to small connection points on the lid and the bin, so when the lid is over the bin there is a "snap" to make it more secure

**Alternatives Considered**
Some alternative solutions are:
- Adding a heigh option for the lid, no new variant, keeping the same shape of existing lid just allowing to have something like this 60/40 ratio between bin and lid

andymai:
## What

Adds an **Extra lid height** control (mm) to the click-lock lid, so contents that stick up out of a short bin — toothpicks, skewers, long drill bits — are enclosed when the lid is on. Resolves #2482.

`0` = the standard one-grid-unit lid, so **every existing design is byte-for-byte unchanged**.

toothpick holder example

## How it works

The user keeps the (short) bin exactly as they design it today; the lid grows to cover what pokes out. Geometrically the value feeds `heightUnitMm + extraHeightMm` into `lidAnchorZ` / `lidWallBottomZ`, which deepens the mating-shell cavity **above the lip**. The lip-mating band and click rails stay pinned at the bottom, so:

- the fit clearance is unchanged,
- the existing per-side **click-lock snap** (already the lid's snap mechanism) still engages exactly as before,
- only the plain wall above the lip lengthens into a tall hollow tube.

Threaded through `resolveLidInputs`, the STEP export lift, and the main-thread preview mirror (the cross-thread agreement test enforces the formula match).

## Design decisions (confirmed with maintainer)

| Decision | Choice |
|---|---|
| Model | Height field on the existing lid (not a new variant) |
| Unit | Absolute mm, 0–100 (clamped) |
| Snap | Reuse existing click-lock rails — no new mechanism |
| Bin sizing | User sets bin height manually; bin logic unchanged |
| Rollout | Ship directly (additive, default 0 = no change) |

## Changes

- `LidConfig.extraHeightMm` (0–100mm) + default 0; migration clamps legacy/corrupt values
- `LidSection` mm stepper + dimensions readout reflects total height
- `lidCompatibility`: warns when a tall lid sits on a marginal 1U bin (long lever arm on a shallow grip)
- **Toothpick Holder** gallery example (short bin + tall lid) with generated thumbnail + mesh
- i18n across all 10 locales

## Tests

- Real-WASM lid scenario: tall lid builds watertight/valid; deepens Z by exactly the added mm; footprint unchanged; **`extraHeightMm: 0` reproduces the standard lid exactly** (same triangle count + bbox)
- Panel: control renders, commits, and clamps over-range input
- Compatibility warning fires only on a tall lid + 1U bin
- Type/default/bounds
- Cross-thread preview-mirror agreement test passes

Full gate green locally: typecheck, lint, knip, all four i18n checks, and every lid test.

---
## Summary by cubic
Adds an extra lid height control (mm) to click‑lock lids so tall items that stick out of short bins are enclosed; 0mm keeps the standard lid and all existing designs unchanged. Closes #2482.

- **New Features**
  - New `lid.extraHeightMm` (0–100mm, 1mm step) with a stepper in `LidSection`; the dimensions readout reflects the added height.
  - Geometry: feeds `heightUnitMm + extraHeightMm` into `lidAnchorZ`/`lidWallBottomZ` across worker and previews; STEP lift mirrors it.
  - Compatibility: warns on `tallLidShortBin` when a tall lid sits on a 1U bin.
  - Example and i18n: Toothpick Holder preset (short bin + tall lid) with thumbnail/mesh; strings added in all locales.
  - Tests: scenario + panel + compatibility + type bounds; confirms 0mm reproduces the old lid, Z deepens by the added mm, XY footprint unchanged.
  - Backward compatibility: migration clamps persisted `extraHeightMm` to 0–100mm; defaults keep saved designs exactly the same.

- **Bug Fixes**
  - Aligned tall‑lid positioning in all preview paths: `StackPlateMesh` and thumbnail regeneration now pass `extraHeightMm` to `lidAnchorZ`, matching `LidMesh` and export.
  - UI/CI polish: stepper onStep uses the configured step; added migration clamp/default tests and tightened the whole‑millimetre step assertion.

<sup>Written for commit 58995defb2e52e414d435743bd26cae7ced33cbd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2486?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed all findings in 58995de:

**Greptile (4/5 — separate stack-plate preview mismatch):** Real bug, fixed. `StackPlateMesh` positioned the glue-on plate on the default-height lid anchor, so a tall lid + `separateStackPlate` rendered the plate at the wrong Z (the STEP export was already correct — it reuses the tall `lidZ`). While fixing, I swept every `lidAnchorZ`/`lidWallBottomZ` call site and caught a second one: the saved-design **thumbnail regenerator** (`thumbnailRegenerator.ts`) had the same omission. Both now thread `extraHeightMm`, matching `LidMesh` and the export lift.

**CI failure (`defaults.test.ts > migrateParams > passes through fully-specified lid config`):** Updated the round-trip expectation for the new `extraHeightMm` field, and added migration coverage (default 0 when absent; clamp out-of-range; fall back on non-numeric).

**Copilot:**
- `LidSection` `onStep` now multiplies the delta by the configured step (correct if the step ever changes from 1), matching the WallCutouts pattern.
- Tightened the "whole-millimetre step" test to assert `Number.isInteger`, not just positivity.

Typecheck, lint, and the affected unit/preview tests pass locally.

PR #2490, opened 2026-07-08
fix(bin-designer): stop parameter panel from scrolling the whole page

andymai:
## What

In the bin designer you could scroll the whole page down and reveal blank space below the parameter panel and 3D preview. This removes that stray scroll.

## Why

The parameter panel's scroll container was `position: static`. The `sr-only` (`position: absolute`) checkbox inputs inside its toggle Switches therefore resolved their containing block all the way up to the initial containing block (`<html>`) rather than the panel. An `overflow: hidden`/`scroll` ancestor only clips an absolutely-positioned descendant when that ancestor *is* the descendant's containing block, so all the layout clips were bypassed. A hidden input's static position deep in the tall scroll area (~1818px down) leaked into `<html>`'s scroll height and made the entire document scrollable.

This is why it was intermittent — it only appeared once the panel grew tall enough (expanded sections / added compartments) that a hidden toggle input's static position fell below the viewport.

## Fix

Add `relative` to the panel's scroll container so it is the containing block for those absolutely-positioned descendants. `overflow-y: scroll` then clips them and the document no longer overflows.

## Verification

Measured the real dev build with Playwright at 5 viewport sizes (desktop, short desktop, landscape, tablet-portrait stacked, mobile). Before: `documentElement.scrollHeight` 1818 vs 1000 viewport, page scrolled 818px. After: scroll height equals viewport height and the page cannot scroll in any layout.

---
## Summary by cubic
Stops the bin designer from scrolling into blank space below the parameter panel and 3D preview. Sets the panel scroll containers and `Switch`/`Checkbox` labels to `relative` so sr-only inputs are clipped, fixing overflow in both panels and app‑wide.

<sup>Written for commit 5e712b56916f345ee29b5ec86ede0fcf58c90eb2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/design-system/Checkbox/Checkbox.tsx
+++ b/src/design-system/Checkbox/Checkbox.tsx
@@ -241,7 +241,10 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
       <label
         htmlFor={id}
         className={cn(
-          'inline-flex items-center gap-2',
+          // `relative` contains the sr-only (position:absolute) input below so it
+          // can't resolve its containing block to the ICB and leak its static
+          // position into the document scroll height inside static scroll containers.
+          'relative inline-flex items-center gap-2',
           disabled ? 'cursor-not-allowed opacity-50' : 'cursor-pointer',
           className
         )}
--- a/src/design-system/Switch/Switch.tsx
+++ b/src/design-system/Switch/Switch.tsx
@@ -153,7 +153,10 @@ export const Switch = forwardRef<HTMLInputElement, SwitchProps>(
       <label
         htmlFor={id}
         className={cn(
-          'inline-flex items-center gap-2',
+          // `relative` contains the sr-only (position:absolute) input below so it
+          // can't resolve its containing block to the ICB and leak its static
+          // position into the document scroll height inside static scroll containers.
+          'relative inline-flex items-center gap-2',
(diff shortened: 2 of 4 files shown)


andymai:

- **Root cause at the source (Copilot):** `Switch` and `Checkbox` now set `relative` on their `label`, so the `sr-only` (`position: absolute`) input is contained by its own label instead of resolving its containing block to the ICB. This closes the leak for every switch/checkbox app-wide.
- **Tool-rack panel (Greptile):** added `relative` to `ToolRackParameterPanel`'s scroll container, mirroring `ParameterPanel`. The container-level guard also covers non-switch `sr-only` descendants (e.g. `CopyButton`/`Spinner` spans), which the component fix alone wouldn't.

Verified with Playwright on the real dev build (desktop + mobile): `documentElement.scrollHeight` equals the viewport height and the page can't scroll. Switch/Checkbox/panel unit tests pass (73) and typecheck is clean.

PR #2520, opened 2026-07-11
fix(design-linking): infer half-unit edge from drawer & warn on mismatch (#2518)

### LINKED ISSUE #2518: Half-size edge positions are not inferred correctly from layout's own properties
**Describe the bug**
Half-size edge position is incorrectly inferred for bins generated from the layout, could result in incorrect orientation.

**To Reproduce**
Steps to reproduce:

1. Set layout to include half-size edges in both X/Y directions.
2. Where the 0.5x0.5 slot is assigned, draw a 1.5x1.5 bin
3. Create design from the bin on the layout
4. Observe that it may be in the wrong orientation

**Expected behavior**
Half-unit edge position for a bin should be initially inferred from the layout's own Half-unit edge position property, should still be possible to change afterwards on the bin, but there should be a warning, and if it exists should be part of the general sanity check of the layout and evaluate whether it makes sense architecturally to show a warning in the bin design workspace to provide early user feedback.
See screenshots and layout.pdf uploaded for reference.

**Screenshots**
Image

Image

Image

Layout:
[layout.pdf](https://github.com/user-attachments/files/29916735/layout.pdf)

**Environment:**

- Browser: Chrome

andymai:
Fixes #2518.

## Problem

When a user creates a bin design from a placed layout bin, the design's half-unit foot edge (`fractionalEdgeX`/`fractionalEdgeY`) always defaulted to `'end'` instead of being inferred from the drawer's own edge orientation. In a drawer configured with a `'start'` half-unit edge, a fractional (e.g. 1.5×1.5) bin was generated with its half foot on the wrong side — a wrong-orientation print.

## What changed

**1. Infer the edge (the bug fix)**
`navigateToCreateDesign` now threads the drawer's `fractionalEdgeX/Y` through the create-from-bin URL, and `useCreateFromBin` applies it to the new design — but only for axes that are actually fractional.

**2. Track deliberate intent**
A new persisted `fractionalEdgeManual` flag is set when the user toggles the edge in the designer. Inference on create leaves it `false`, so a *later* drawer-edge change surfaces a warning, while an *intentional* override stays silent.

**3. Warn on mismatch (with a one-click fix)**
- In-designer banner when the open design's edge disagrees with a linked drawer, with a **Match drawer** action.
- Per-bin indicator in the inspector's linked-design section, driven by the design edge now carried on the lightweight registry ref, with the same one-click fix.

The mismatch logic lives in shared pure helpers (`hasFractionalEdgeMismatch` / `computeMatchedEdges`) so both features can use it without crossing the `bin-designer` ⇄ `design-linking` module boundary.

## Decisions
- Edge is copied from the drawer property (the drawer holds a single fractional slot, so a fractional bin's foot belongs on that edge).
- Inference runs only on create-from-bin; it never overrides a manual choice afterward.
- A deliberate manual edit suppresses the warning via the `fractionalEdgeManual` flag.

## Tests
Unit tests cover the inference (URL threading + apply), the `hasFractionalEdgeMismatch`/`computeMatchedEdges` helpers, the manual-flag toggle, registry edge round-tripping, and the banner. i18n keys added across all 9 locales. `pnpm run quality` is green (0 errors); targeted suites for the touched areas pass (812 tests).

---
## Summary by cubic
Infer the half‑unit edge from the drawer when creating a design from a layout bin, and warn when a linked design’s edge disagrees with the drawer with a one‑click “Match drawer” fix. Prevents wrong‑orientation prints for fractional bins and keeps linked designs aligned with drawer edge settings.

- **Bug Fixes**
  - Infer `fractionalEdgeX/Y` from the drawer on create‑from‑bin, applied only to fractional axes.
  - Hardened behavior: normalize missing/invalid drawer edges to `'end'`, announce the banner with `role="alert"`, and omit `fractionalEdgeManualX/Y` from user defaults to avoid propagating overrides.

- **New Features**
  - Per‑axis manual flags `fractionalEdgeManualX/Y` suppress warnings only for the axis the user changed; “Match drawer” realigns the fractional axes and clears just those flags.
  - Designer banner and inspector warning use shared helpers `hasFractionalEdgeMismatch` and `computeMatchedEdges`; registry refs now carry edge and manual fields via `registryEdgeFields`.

<sup>Written for commit d389529d3714e3089ee23c2278849a3ccaeac93f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed the review feedback in 3d20e0f:

- **Greptile (per-axis suppression):** Split the single `fractionalEdgeManual` flag into `fractionalEdgeManualX`/`fractionalEdgeManualY`. A deliberate override on one axis no longer hides a real mismatch on the other — which matters for the reporter's both-axes-fractional (1.5×1.5) case. `Match drawer` now clears only the flags for the axes it realigns.
- **Copilot (English wording):** Reworded the inspector warning to `Half-unit foot doesn't match this drawer's edge.` to drop the dangling possessive.
- **Copilot (pt-BR):** Fixed the `da desta` duplication (and re-aligned all locale translations with the new wording).
- **Copilot (parseEntry):** Cache the parsed edge values instead of calling the `edge(...)` helper twice per field.

andymai:
Round-2 review addressed in d389529:

- **cubic P2 (user default leakage):** Added `fractionalEdgeManualX/Y` to `STYLE_DEFAULT_OMIT_KEYS` so a manual edge choice saved as the "new bin" default no longer carries into every new design and mutes warnings.
- **cubic P2 (legacy/invalid edges):** `drawerEdge` now normalizes any non-`'start'` value (unset, legacy, corrupt) to `'end'`, so comparisons can't spuriously mismatch and "Match drawer" can't persist an invalid edge.
- **cubic P2 (a11y):** Added `role="alert"` to the mismatch banner.
- **Copilot (consistency):** Swapped `!Number.isInteger(...)` for the shared `isFractional(...)` helper in the create-from-bin URL threading.
- **Copilot (coverage):** Added unit tests for `matchDesignEdgesToDrawer` (realign + clear flag; error toasts on load/update failure).

The earlier pt-BR / English-wording / `parseEntry` Copilot threads were already resolved in 3d20e0f — those were stale re-posts.

PR #2560, opened 2026-07-13
fix(baseplate): thicken magnet pad walls

### LINKED ISSUE #2559: Fix for 0.8mm nozzles wasn't a fix
**Describe the bug**

I thought I looked at the 3mf and slicer output but I think I messed up. It needs more of a buffer with a 0.8mm nozzle.

**To Reproduce**
Steps to reproduce:

1. Use magnets
2. Set nozzle size to 0.8mm
3. export 3mf
4. be sad after wasting filament

**Expected behaviour**

More of a wall between the magnet hole and the edge

**Screenshots**
If applicable, add screenshots.

Image
This is what's being generated now

Image
With these modifications. Tested a print, has a 2.5mm edge this time.

**Environment:**,

- Browser: 150.0.7871.115
- Device: MacOS 26

vjanelle:
## Summary

  **Type**: fix

  ## Context

  Fixes #2559.

  The 0.8 mm nozzle path left too little material between the magnet pocket and the outside edge. Gridfinity magnet centers remain unchanged; only the lightweight-floor relief contracts to retain a thicker pad wall.

  ## Changes

  - Preserve existing Gridfinity magnet-hole coordinates.
  - Add wider-nozzle outside-wall margins to lightweight relief cuts.
  - Include magnet-floor nozzle geometry in the mesh cache key.
  - Add regression tests for coordinates, margins, partial tiles, and cache separation.

  ## Architecture

  - **Stores**: none
  - **Features**: baseplate, print-export
  - **New types**: none
  - **i18n keys**: none
  - **Storage/schema changes**: none

  ## Test plan

  - [x] Focused Vitest suites: 56 tests passed
  - [x] `tsgo -b --noEmit`
  - [x] Prettier check
  - [x] `pnpm run build`
  - [x] New code has colocated sibling tests
  - [x] i18n: `pnpm run check:i18n`
  - [ ] No `console.log`, `any`, `var`, `==`, `!` introduced
  - [ ] CLAUDE.md updated if conventions changed

### DIFF
--- a/src/features/generation/worker/generators/baseplateCacheKeys.test.ts
+++ b/src/features/generation/worker/generators/baseplateCacheKeys.test.ts
@@ -112,6 +112,20 @@ describe('meshCacheKey — nozzle-dependent floor pads (issue #2559)', () => {
       );
     expect(solid(0.4)).toBe(solid(0.8));
   });
+
+  it('does not key draft geometry on a nozzle used only by skipped relief', () => {
+    const a = meshCacheKey(
+      base({ connectorNubs: false, magnetHoles: true, nozzleSizeMm: 0.4 }),
+      false,
+      true
+    );
+    const b = meshCacheKey(
+      base({ connectorNubs: false, magnetHoles: true, nozzleSizeMm: 0.8 }),
+      false,
+      true
+    );
+    expect(a).toBe(b);
+  });
 });
 
 describe('cache keys with outlines (issue #2528)', () => {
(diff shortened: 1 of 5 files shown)


vjanelle:
Apologies! Actually printed this time.

[image](https://github.com/user-attachments/assets/f602c84b-7ac4-4544-97f4-2ce4aabe6fb9)

vjanelle:
Looked at it with 42x42 and it's fine.. 68x68 though it's suspicious

vjanelle:
IMG_0152

Better (still 2.5mm), I think anything further would be something I would have to deal with with my print settings.

PR #2587, opened 2026-07-15
feat(bin-designer): add per-label size control (#2586)

### LINKED ISSUE #2586: Add a label size property
**Problem**
Fix #2584 resolved the label vanishing on small cut-outs, but now that labels are unbounded they can grow disproportionally large compared to the cutout. 

Image

**Proposed Solution**
Add a label size property to the interface.

andymai:
## What

Adds a **Label size** control to cutout labels and label-tab text, so an oversized label can be reined in without shrinking every other label.

Closes #2586.

## Why

#2584 fixed labels vanishing on narrow cutouts by removing the cutout-span cap on the auto-fit band. The side effect: with the cap gone, a short label auto-grows all the way to the design's `maxFontSize` (20mm default) and dwarfs a small cutout (see the screenshot in #2586).

## How it works

- Each cutout label (and label-tab text) defaults to **Auto** (today's auto-fit behavior).
- Turning Auto off reveals a size slider (range = the design's `minFontSize`..`maxFontSize`) that locks an explicit mm size.
- The explicit size is **clamped to the available band and still respects the legibility floor**, so it can only ever make a label smaller than auto-fit would, never larger, and never bleeds over a neighboring cutout.

The size cap is enforced in three places that must agree: the worker engraver (`buildTextSolid`), the 2D editor preview (`cutoutLabelFit`), and the server-side share validator (bounded so a crafted share can't smuggle a worker-crashing size).

## Tests

- Worker scenario test (real WASM geometry) asserting the override shrinks the built solid and clamps to the band.
- Preview fit-sizing unit tests mirroring the worker.
- Server validation tests for the new `textStyle` field on cutouts and label tabs.
- Helper unit tests for the set/clear override logic.

All quality gates green: typecheck, lint, the four i18n checks (keys added across all 10 locales), knip.

---
## Summary by cubic
Adds a per‑label “Label size” control for cutout labels and label‑tab text, letting you cap auto‑fit on a label-by-label basis. Prevents short labels from ballooning to the design’s max size on small cutouts (closes #2586).

- New Features
  - “Auto” by default; turning it off shows a size slider (min..max font size in mm) stored in `textStyle.fontSizeOverride`. Toggling seeds the override at max; toggling back clears it.
  - Override caps auto-fit and is clamped to the available band and floored at `minFontSize`; it only shrinks, never enlarges or bleeds.

- Refactors
  - Extracted shared `LabelSizeControl` used by cutouts and label tabs; preview sizing moved to `cutoutLabelFit`.
  - Floored sub-min overrides in both the worker (`buildTextSolid`) and 2D preview; added validation via `validateTextStyleOverride` and tests to keep worker, preview, and server in sync.

<sup>Written for commit 33ad78fa40884fa659fece0cdec2059ae37d0474. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2587?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed the legibility-floor finding in 33ad78fa4: the `fontSizeOverride` is now clamped into `[minFontSize, fit.fontSize]` in both the worker (`textBuilder.ts`) and the 2D preview (`cutoutLabelFit.ts`), so a crafted/imported sub-floor override renders at `minFontSize` instead of below it. Since the band always fits at least `minFontSize`, it clamps up rather than dropping the label. Added below-floor coverage to both `textBuilder.scenario.test.ts` and `cutoutLabelFit.test.ts`.

PR #2589, opened 2026-07-15
fix(sidebar): make baseplate, drawer shape, and stack solver match the sidebar

andymai:
The Active baseplate, Custom drawer shape, and Fit unit to stack sections each looked out of place next to the panels around them. Each had opted out of the sidebar's established patterns in a different direction, leaving **four heading tiers and two different controls for the same boolean** in a 288px column.

## What changed

| Section | Before | After |
| --- | --- | --- |
| Active baseplate | Bare uppercase micro-label, no chevron, inline "Manage…" | `Collapsible size="md"` with Manage in `actions` — a peer of Layers/Categories |
| Custom drawer shape | `FeatureToggle` pill directly below Half-grid mode's checkbox | Shared `ToggleRow` checkbox; actions are full-width buttons |
| Fit unit to stack | Inline `text-[11px]` heading, always expanded, brighter labels | Nested `size="sm"` section, collapsed by default, `SettingsRow` rows |

Headings now resolve to the two tiers the design system already defines: `size="md"` for sections, `size="sm"` for nested ones. No new heading styles, no new i18n keys.

## Notable details

- **`ToggleRow` is new** (`src/shared/components/ToggleRow`). Half-grid mode previously hand-rolled a `role="checkbox"` div with its own keyboard handling; rather than copy that into the drawer-shape section, both now share one implementation. It takes `variant` so the mobile settings sheet keeps `lg` hit areas.
- **Label brightness was a real bug.** `HeightUnitSolver` set its own `labelClass = 'text-content-secondary'`, so "Target height (mm)" rendered a step brighter than "Grid unit" 40px above it. `SettingsRow` (already used by its neighbours) fixes it.
- **The solver's inputs stay raw**, not `DeferredNumberInput`: the solve previews live as you type and `target` starts empty, neither of which a commit-on-blur numeric input can express. That hand-rolling was justified; the row structure wasn't.
- **Corner cuts are now always reachable.** Previously they were a `pl-6` ghost link aligned to nothing when no outline existed. They build an outline from the plain rectangle, so they belong at both states.
- Drops the only intra-section divider in the sidebar, and reclaims ~120px by collapsing the solver.

## Verification

- `pnpm run typecheck`, `pnpm run quality` (0 errors, knip clean), 197 tests across the affected trees.
- Playwright spot check at 1400×1000 desktop and 390×844 mobile. Caught one real bug in the process: `Select`'s wrapper is `inline-flex`, so it needs `fullWidth` rather than a `w-full` class — otherwise it collapses to content width once it's no longer inside a `flex-1` parent.

---
## Summary by cubic
Aligns the Baseplate, Drawer Shape, and Stack Solver with the sidebar’s patterns to improve consistency, accessibility, and save space. Introduces a shared checkbox row (`ToggleRow`) and standardizes headers and row styles.

- **Refactors**
  - Baseplate: `Collapsible size="md"` with Manage in `actions`; Select uses `fullWidth`.
  - Drawer Shape: replace `FeatureToggle` with `ToggleRow`; actions are full-width buttons; corner cuts always available; supports `variant="mobile"` with 44px touch targets; removes the extra divider.
  - Stack Solver: rows use `SettingsRow`; moved into a nested `Collapsible size="sm"` collapsed by default; keeps raw inputs for live previews.
  - Sidebar: Half-bin mode now uses `ToggleRow`.
  - Added `src/shared/components/ToggleRow` with tests; updated Drawer Shape tests to use `role="checkbox"`.

- **Bug Fixes**
  - Fixed label brightness in the solver by using `SettingsRow`.
  - Prevented Select width collapse by switching to `fullWidth`.

<sup>Written for commit 763e2d59fc0257521a5340772713d5eb078cf857. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/baseplate/components/ActiveBaseplatePanel/ActiveBaseplatePanel.tsx
+++ b/src/features/baseplate/components/ActiveBaseplatePanel/ActiveBaseplatePanel.tsx
@@ -1,14 +1,16 @@
 /**
- * Compact "Active baseplate" switcher for the planner sidebar.
+ * Compact "Baseplate" switcher for the planner sidebar: a Select bound to the
+ * global baseplate library plus a "Manage…" affordance that opens the library
+ * modal.
  *
- * Mirrors the ActiveLayerPanel slot: a Select bound to the global baseplate
- * library plus a "Manage…" affordance that opens the library modal.
+ * Built on `Collapsible size="md"` with the manage button in `actions` so this
+ * slot reads as a peer of Layers and Categories rather than a lone label.
  */
 
 import { useCallback } from 'react';
 import { useViewStore } from '@/core/store/view';
 import { useTranslation } from '@/i18n';
-import { Button, Select } from '@/design-system';
+import { Button, Collapsible, Select } from '@/design-system';
 import { baseplateDesignId } from '@/core/types';
 import { useBaseplateLibrary } from '@/features/baseplate/hooks/useBaseplateLibrary';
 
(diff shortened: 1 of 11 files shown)


andymai:
Addressed the mobile touch-target finding in 763e2d59f. Both action buttons now derive their height from `variant` (`h-11` / 44px on mobile, matching the convention across `src/shell/Mobile/`; `h-8` on desktop) instead of a hardcoded `h-8`. Added tests pinning both heights so it can't silently regress.

PR #2593, opened 2026-07-15
feat(supporters): keep the supporters page up to date automatically

andymai:
New Ko-fi supporters now appear on `/supporters` within seconds — no hand-edit of `supporters.json`, no deploy.

## The constraint that shapes everything

Ko-fi has **no read API** and **no webhook replay**. There is nothing to poll and nothing to backfill, so this is a push-only, one-shot feed: whatever isn't recorded on delivery is gone for good. Three consequences:

- **Fail loudly, never silently 200.** Missing `KOFI_VERIFICATION_TOKEN` → 503; Redis down → 503. Ko-fi retries on a failure response, so a 503 is recoverable while a cheerful 200 would lose that supporter permanently.
- **Verify before touching Redis.** The token is compared constant-time before any I/O, so a forged request costs nothing. Rate limiting sits *behind* that as a backstop for a leaked token — it is not the front door.
- **Dedupe twice.** `message_id` guards Ko-fi's retries; a donor id guards repeat donations from one person. Without the second, a monthly subscriber earns a fresh bin every renewal.

## Flow

```
Ko-fi ──POST──> api/kofi-webhook.ts ──> Redis (supporters:donors)
                  verify → dedupe → filter          │
                                                    │ GET, s-maxage=60
                     SupportersPage <── useSupportersData
                     (bundled JSON = fallback)
```

`supporters.json` is now a **fallback**, not the source of truth. An outage shows a stale baseplate rather than an empty one, and an empty API response is read as "not seeded yet" so it can't wipe the scene.

## Privacy

**Stored:** a salted donor-id hash, and a display name. **Never stored:** the email, the amount, or the message the supporter left.

The salt is load-bearing, not decoration. An *unsalted* SHA-256 of an email is brute-forcible against a wordlist, so it would effectively still be an email. `deriveDonorId` salts with `TOKEN_SALT` and **refuses to derive an id without it** rather than emit a reversible digest.

⚠️ **This is the one judgement call to review.** It bends `supporters.json`'s "never store emails" rule to a hashed derivative. It's the only way to recognise a repeat donor, since `email` is Ko-fi's sole stable identifier. Happy to drop it and accept duplicate bins per renewal if you'd rather.

## Untrusted input

`from_name` is arbitrary text typed by a stranger that renders on a public page. `normalizeDisplayName` honours Ko-fi's `is_public`, runs the name through the existing `contentFilter`, and length-caps it. A rejected name makes the supporter **anonymous rather than dropped** — the name is untrusted, the person isn't.

## Deploy steps (not automatic)

1. Set `KOFI_VERIFICATION_TOKEN` in Vercel (Ko-fi → More → API). Until then the endpoint 503s and no supporter is lost — Ko-fi retries.
2. Point the Ko-fi webhook at `https://gridfinitylayouttool.com/api/kofi-webhook`.
3. Run `pnpm seed-supporters` once against prod Redis (`--dry-run` to preview). The 45 pre-webhook supporters have no email, so they get synthetic `seed:*` ids no delivery can collide with. Until this runs, the API returns only new supporters and the page keeps showing the bundled 45.

## Verification

Unit: 792 tests across `src/features/supporters` and `api/lib`.

Drove the **real handlers against a real Redis** (`redis-server`), not mocks — 15/15:

| Check | Result |
| --- | --- |
| Normal donation lands, name served | pass |
| Forged token → 401, name never stored | pass |
| Ko-fi retry (same `message_id`) deduped, no second bin | pass |
| Same donor again (new `message_id`) → still one bin | pass |
| `is_public: false` → anonymous bin, still counted | pass |
| Filtered name → anonymous, supporter kept | pass |
| Subscription renewal → ignored | pass |
| Stored hash has no email / amount / message | pass |
| Donor ids opaque, edge cache header set | pass |
| Seed idempotent; seeded + live entries coexist (45 → 46) | pass |

Two bugs my own tests caught while writing them, both fixed:
- `'a'.repeat(200)` as a name trips `contentFilter`'s repeated-char spam pattern, so it anonymises *before* reaching the length cap. The ordering is correct; the test fixture was unrealistic. Both behaviours are now pinned.
- `seed-supporters --dry-run` demanded `REDIS_URL` despite never connecting.

---
## Summary by cubic
Automatically keeps the `/supporters` page up to date via a Ko‑fi webhook and Redis, with safe fallbacks and tight validation. Adds docs in `api/README.md` explaining the sanitize order and regex ReDoS constraints.

- **New Features**
  - `api/kofi-webhook.ts`: Verifies the Ko‑fi token (constant‑time) before any I/O; 503s on missing token or Redis; distinguishes outage (503) from genuine rate limit (429) with a Redis liveness ping; dedupes using a hashed `message_id` and a salted donor hash; ignores subscription renewals after the first; stores only a donor‑id hash and display name.
  - `api/supporters.ts`: Serves the public supporter list; edge‑cached and rate‑limited.
  - Frontend: `useSupportersData` fetches the live list with a 1.5s timeout and falls back to bundled `supporters.json`; the page waits until data settles to avoid count/camera jumps.
  - Tooling/Tests/Docs: `pnpm seed-supporters` backfills the 45 pre‑webhook supporters; added unit tests for parsing, privacy, dedupe, fetch fallback, UI hold, and outage vs rate‑limit behavior; `api/README.md` documents ingest rules, sanitize order, and ReDoS constraints.

- **Bug Fixes**
  - Closed a polynomial ReDoS in display‑name filtering: cap names before filtering and bound the inline‑handler regex (`on\w{1,20}\s{0,10}=`).
  - Strip control, zero‑width, and bidi characters from display names to prevent spoofing and rendering issues; sanitize before capping, non‑ASCII preserved.
  - Webhook now validates content type, accepting form‑encoded or JSON and 415ing others.

<sup>Written for commit 64f74cc54c783061d648ea5219df5e67c22cc35e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2593?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thanks — one fixed, one I'd push back on.

**Null bytes in `message_id` — fixed** in 1192e6367. `message_id` is now hashed into its key (`supporters:msg:{sha256(id)[0:32]}`) rather than embedded raw.

Worth noting *why* it's a hash and not validation: the Ko-fi feed has no replay, so rejecting an unexpected id shape would lose that supporter permanently. A hash can't false-negative. Verified against a real Redis — an id carrying null bytes, newlines, `share:hash:`-style colons and 500 chars of padding still dedupes to exactly one bin, with the key a bounded 32-char hex.

**Rate-limit ordering — intentional, and I think the reasoning inverts here.** The rule "rate limit first" holds elsewhere in this codebase because those endpoints have no cheaper gate. This one does:

| | forged request cost |
| --- | --- |
| verify → rate limit (current) | env-var string compare. **Zero I/O.** |
| rate limit → verify | one Redis round trip (`zcount` + `zadd` + `expire`), *then* rejected |

Rate-limiting first would make the cheapest attack — flooding garbage at the endpoint — *more* expensive for us to absorb, charging a Redis op per forged request. Verifying first makes those rejections free. The rate limit is still there, positioned where it can actually do work: as a blast-radius cap on a **leaked** token, which is the only way past the first gate.

The one thing an unauthenticated caller can force is the `JSON.parse` — the token lives inside the body, so Ko-fi's payload format requires parsing before verification. Vercel's body cap bounds it.

I've documented the rationale in the handler and `api/README.md` so it doesn't read as an oversight to the next person. Happy to reorder if you see an attack this misses.

andymai:
CodeQL caught a **real high-severity bug I introduced** — fixed in bace46fce. Worth reading, because it's a good one.

Exporting `filterDisplayName` created a path from unbounded attacker text straight into the content filter's backtracking regexes. `normalizeDisplayName` filtered the **full-length** `from_name` and only capped it afterwards, so `/on\w+\s*=/` saw the entire payload. That pattern backtracks quadratically on `"ononon…"`: every `"on"` is a candidate start, and the greedy `\w+` walks the rest of the string before failing to find `=`.

Measured on the old pattern:

| input length | old | new |
| --- | --- | --- |
| 2,000 | 1.3 ms | 0.1 ms |
| 10,000 | 33 ms | 0.3 ms |
| 40,000 | 524 ms | 1.1 ms |
| 120,000 | **4,678 ms** | 3.5 ms |

60× the length for 3,600× the time — textbook O(n²). At Ko-fi payload sizes that's minutes of CPU per request. It needed a leaked token to reach, but the token gate isn't the fix.

Fixed at both ends:

1. **Cap before filtering.** This is also the more *honest* check: we render at most `MAX_DISPLAY_NAME_LENGTH` characters, so that's exactly the text that should have to survive the filter. The old order could reject a name for content past the cut that would never be displayed.
2. **Bound the pattern's quantifiers** (`on\w{1,20}\s{0,10}=`). Fixes it at source for every caller including the share API, where the same regex was reachable (bounded there by field limits, so latent rather than exploitable).

Detection is unchanged — verified `onclick=`, `onerror =`, `ONMOUSEOVER=`, `` all still caught, and benign strings (`one = two`, `onwards`, `Jo Example`) match the old pattern's behaviour exactly. Regression tests pin both the timing and the ordering.

andymai:
Good catch, and a fair one — fixed in 3944cb47f. This was a real gap, and I'd actually seen the 429 in a local run and misread it as the fail-closed path working.

Two things combined:

1. **`getRedis()` returns null only when `REDIS_URL` is *unset*.** It constructs a client happily when the server is merely unreachable, so the `if (!redis) → 503` guard never fired for an actual outage — precisely the case it was written for.
2. **`checkRateLimit` swallows Redis errors** and returns `allowed: false`, indistinguishable from a genuine rejection. So an outage answered 429.

You're right that this matters: 429 has murkier retry semantics than 503, and with no replay on this feed, guessing wrong loses the supporter permanently.

A rejection now pings Redis to tell the two apart. The ping only runs on the rejection path, so the happy path is untouched.

Verified against a real Redis:

| scenario | before | after |
| --- | --- | --- |
| Redis unreachable | `429 RATE_LIMITED` | `503 SERVICE_UNAVAILABLE` |
| genuinely over limit | `429` | `429` (unchanged) |
| healthy | `200` | `200` |
| forged token | `401` | `401` |

Also added `api/kofi-webhook.test.ts` (12 tests) covering the contract end-to-end — forged token never reaches the store *and* never spends a rate-limit check, unconfigured token 503s rather than accepting unverified, dedupe, renewals, private supporters stored as anonymous, and no email/amount/message written. The outage test was confirmed to fail against the previous behaviour before the fix went in.

PR #2594, opened 2026-07-15
fix(baseplate): stop discarding a saved baseplate name silently

### LINKED ISSUE #2591: Saving a baseplate draft silently does nothing on a fresh layout
## Symptom

On `/baseplate` with a fresh layout (no prior baseplate edits), clicking **Save** on an unsaved draft prompts for a name, accepts it, and then silently returns to "Unsaved draft". No design is created, no error, no toast. The name is discarded.

## Repro

1. Open `/baseplate` in a clean browser profile (no `gridfinity-library-index-v1` in localStorage).
2. Click **Save** — the naming input appears, prefilled "Baseplate 1".
3. Click **Save** again to confirm.
4. The header returns to "Unsaved draft" with the `New / Save / Manage…` cluster. The dropdown still has only "Unsaved draft"; nothing was persisted.

Confirmed via Playwright against `pnpm run dev`:

```
initial:        ["…","Export","New","Save","Manage…", …]
after 1st Save: ["…","Export","Save","Cancel", …]        <- naming mode, input present
after 2nd Save: ["…","Export","New","Save","Manage…", …] <- back to draft
select options: ['Unsaved draft']                        <- nothing saved
```

## Cause

`BaseplateSelector` guards `confirmNaming` on the raw store value:

```ts
const baseplateParams = useLayoutStore((s) => s.layout.baseplateParams);

const confirmNaming = useCallback(async () => {
  const name = draftName.trim();
  if (!name || !baseplateParams) {
    cancelNaming();   // <- silent no-op, indistinguishable from cancelling
    return;
  }
  ...
```

But `baseplateParams` is optional (`Layout.baseplateParams?: StoredBaseplateParams`, `src/core/types.ts:69`) and stays `undefined` until something writes it. `BaseplatePage` renders correctly regardless because it supplies a fallback:

```ts
baseplateParams: state.layout.baseplateParams ?? DEFAULT_BASEPLATE_PARAMS,  // BaseplatePage.tsx:83
```

The selector has no such fallback, so the page *looks* like it has params while the save path sees `undefined` and bails. `nextBaseplateName()` always returns a non-empty name, so `!baseplateParams` is the branch that fires.

## Suggested fix

Mirror the page's fallback in the selector — `DEFAULT_BASEPLATE_PARAMS` is already imported there for `handleNew`:

```ts
const baseplateParams =
  useLayoutStore((s) => s.layout.baseplateParams) ?? DEFAULT_BASEPLATE_PARAMS;
```

Worth also splitting the two guard conditions so a genuine failure surfaces instead of silently cancelling — `saveCurrentAsNew` returning `err` is swallowed the same way today.

Found while verifying #2590 (header styling only; unrelated to the cause).

andymai:
Closes #2591. **Stacked on #2590** — based on `fix/baseplate-header-consistency` because both touch `BaseplateSelector`. Retarget to `main` once #2590 lands.

## The bug

On a fresh layout, `Save` on an unsaved draft prompted for a name, accepted it, then silently returned to "Unsaved draft". No design, no error, no toast. The name was thrown away.

## Cause

`layout.baseplateParams` is optional (`core/types.ts:69`) and stays `undefined` until something writes it. `BaseplatePage` renders from a fallback:

```ts
baseplateParams: state.layout.baseplateParams ?? DEFAULT_BASEPLATE_PARAMS,  // BaseplatePage.tsx:83
```

…so the page looks fully configured. But `BaseplateSelector` read the raw store value, saw `undefined`, and took the cancel branch:

```ts
const baseplateParams = useLayoutStore((s) => s.layout.baseplateParams);
if (!name || !baseplateParams) { cancelNaming(); return; }   // silent
```

It now falls back to the same defaults the page is already drawing.

## Also: the guard conflated two different outcomes

An empty name is a **cancel** — silence is correct. A save that **fails** is not, and it took the identical branch, so a `saveCurrentAsNew` error was indistinguishable from backing out. Split, with an error toast on failure (`toast.baseplateSaveFailed`, mirroring the existing `toast.layout*Failed` family across all 10 locales).

## Why the tests missed it

The `useLayoutStore` mock hardcoded `baseplateParams: mocks.fakeParams`, so the `undefined` branch was unreachable from these tests. It's settable now.

**Both new tests were confirmed to fail against the original code before the fix went in** (2 failed / 6 passed), then pass after — so they pin the defect rather than my implementation.

## Verification

Browser, the exact repro from the issue:

```
before  : ["Unsaved draft"]
after   : ["Unsaved draft","Baseplate 1"]
buttons : [... "New","Save As","Manage…" ...]   <- flipped from Save, so a design is really active
```

---
## Summary by cubic
Fixes #2591: Saving a baseplate on a fresh layout no longer discards the entered name. `BaseplateSelector` now falls back to `DEFAULT_BASEPLATE_PARAMS` and shows an error toast on save failure.

- **Bug Fixes**
  - Use defaults when `layout.baseplateParams` is undefined so Save uses the same params the page renders.
  - Split cancel vs failure: empty name cancels; failures show `toast.baseplateSaveFailed` (added in all locales).
  - Tests: make `useLayoutStore` params settable; add cases for fresh-layout save, failure toast, and cancel-without-toast.

<sup>Written for commit a9dadeea5bc67273ca6abd5d58f6069f3758c16b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Superseded by the PR above — GitHub auto-closed this one when #2590 merged and its base branch was deleted, and a closed PR can't be reopened or retargeted once its base is gone. Same commits, rebuilt on `main`, plus the review point about keeping the name form open after a failed save.

PR #2601, opened 2026-07-15
refactor(baseplate): manage baseplates the way the bin designer does

andymai:
> "why does the baseplate generator page use a completely different ux/ui pattern for naming/saving/managing multiple designs? it should be closer to the bin designer"

It did, and #2590 only flattened the styling. That was makeup — the real divergence was the **save model** underneath.

## Before / after

```
before  /designer   Untitled Bin   [Designs]     [Export]  ✓ Saved
        /baseplate  [Baseplate 1 ▾] New  Save/Save As  Manage…

after   /designer   Untitled Bin   [Designs]     [Export]  ✓ Saved
        /baseplate  Baseplate 1    [Baseplates]  [Export]  ✓ Saved
```

## Why they diverged

`useDesignerInit`'s contract is literally *"Ensures the designer always has an active design"* — create one if absent, *"so auto-save works immediately"*. Because a design always exists and always saves, **there is nothing for a Save button to do**. Hence: a name, a list, a status.

The baseplate invented a draft — *"drafts (activeBaseplateId === null) stay unsaved until an explicit Save"*. That single decision conjured `New / Save / Save As / Manage…` and the dropdown. The buttons were a symptom, not the disease.

Partly an accident of timing: the baseplate library shipped **#2522 four days ago**; the designer's has existed since ~February.

## What changed

The baseplate has no draft state either. `useBaseplateInit` guarantees an active design, `useBaseplateAutoSave` keeps it current and reports status. New moves into the library modal, where the designer already keeps it (`DesignListDialog`).

## The one deliberate divergence

`useBaseplateInit` does **not** copy `useDesignerInit`'s "restore the last active design" step, and this matters:

> A `SavedDesign` is standalone, but `baseplateParams` live on the **`Layout`** (`core/types.ts:69`). Adopting another design would **silently overwrite baseplate settings the layout already has**.

So it creates a design *from the layout's own params* — exactly what the old explicit Save did, minus the click. There's a test pinning this ("preserves the layout's existing params instead of adopting another design").

It runs only on `/baseplate`, so an entry appears when someone opens the tool, not for every layout they own.

## Incidental

- **`SaveStatusIndicator` + `SaveStatus` move to `shared/`** — both pages render it and features can't import across each other. `binDesigner.saveFailed` → `header.saveFailed`, joining the already-shared `header.saving`/`header.saved` and carrying its translations.
- **The resting "saved" state is derived**, not set in the effect — ESLint correctly flagged the cascading render. The per-design reset adjusts state during render, per the React docs and `DeferredNumberInput`'s existing precedent.
- **Drops 4 dead i18n keys** and fixes `emptyHint`, which told users to *"Save a baseplate from the editor"* — a button that no longer exists.

## Verification

807 tests pass; typecheck clean; lint back to its 24-warning baseline with 0 errors.

Drove both pages side by side — the headers now read part-for-part identical, in the same order:

```
/baseplate  : Layout · Bins · Baseplate · Baseplate 1  · Baseplates · Export · Saved
/designer   : Layout · Bins · Baseplate · Untitled Bin · Designs    · Export · Saved

no Save/SaveAs/New : true
no dropdown        : true
click name → editor: true
renamed in place   : true
```

The side-by-side also caught two things unit tests couldn't: the parts were in the **wrong order** (`Export · name · Baseplates`), and `✓ Saved` never appeared because the status sat at `idle` and the indicator returns null on idle.

## Note

This supersedes the `baseplateParams` fallback from #2599 at the header level — `confirmNaming` is gone entirely, so the silent-discard bug can't recur. The `?? DEFAULT_BASEPLATE_PARAMS` guard lives on in `useBaseplateInit`.

---
## Summary by cubic
Brings the baseplate page to parity with the bin designer by removing drafts and autosaving the active design. The header now matches the designer: inline rename, “Baseplates” list button, Export, and a save status indicator.

- **Refactors**
  - Replaced `useBaseplateInit` with `useBaseplateLibraryInit({ autoCreate: true })` to resolve the pointer and create a default design on `/baseplate` only; planner mount stays inert.
  - Updated `useBaseplateAutoSave` to return shared `SaveStatus`, derive “saved” when idle, and guard against reporting after a design switch; header renders shared `SaveStatusIndicator`.
  - Simplified `BaseplateSelector`: shows the active name with inline rename and a “Baseplates” button; removed dropdown and Save/Save As/New.
  - Added “New Baseplate” in the library modal that creates from defaults via `nextBaseplateName(list)` and activates it.
  - Moved `SaveStatus` and `SaveStatusIndicator` to `shared/`; designer imports updated. Added tests for `nextBaseplateName` and `useBaseplateLibraryInit` auto-create/StrictMode.

- **Bug Fixes**
  - Fixed a double-create race that could orphan designs by consolidating creation into one hook and gating it to `/baseplate`.
  - Prevented save-status flashes from previous designs after switching.
  - “New Baseplate” now surfaces failures with a toast and keeps the modal open for retry.
  - Updated i18n: added `header.saveFailed` and new baseplate labels; removed dead keys and corrected the empty library hint.

<sup>Written for commit c1c5f78f5bb50a1a2bdf2e4843381fdaee81f17d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2601?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Both fixed — and the first finding was pointing at something worse than the race itself.

**`useBaseplateInit` shouldn't have existed.** I wrote it without reading `useBaseplateLibraryInit`, which already auto-seeds exactly that case, and does it better:

- guards against the active layout switching mid-await
- resolves once per layout
- writes via `setActiveBaseplateLocal` because *"load-time resolution must not create undo entries"*

Mine used the CQRS command, so **every page load would have pushed an undo entry** — `Ctrl+Z` after opening `/baseplate` would have "undone" opening it. That's worse than the orphan and nothing flagged it. It also carried a third copy of `nextBaseplateName`.

So it's deleted. `useBaseplateLibraryInit` gains an `autoCreate` option covering the one case it didn't handle: a layout with no params at all.

**`autoCreate` is opt-in, not unconditional** — and that distinction matters. This hook also mounts on the planner (`BaseplateLibraryInitMount`). Creating there would mint a library entry for *every layout on load*, whether or not its owner ever opened the baseplate tool — precisely the library spam this refactor exists to avoid. Your existing test `does not stamp the seed onto a layout switched to mid-resolve` failed the instant I made it unconditional, which is how I caught it.

Verified in a browser:

```
entries after planner load : 0 (expect 0)
entries after /baseplate   : 1 ["Baseplate 1"]
```

**Second finding fixed** as well: `performSave` now checks the design is still active before reporting, so a save landing after a switch can't flash a status belonging to the previous design.

New tests (separate file — vitest isolates per file, and these assert exact library contents while a neighbouring test's open connection blocks `deleteDatabase`): auto-create, the inert planner mount, StrictMode double-invoke yielding exactly one design, and opening the page staying non-undoable.

One thing I deliberately did **not** make robust: two *simultaneous* mounts still create two designs, since `inProgress` is a per-instance ref. `BaseplateLibraryInitMount`'s own docstring says the two mounts are "mutually exclusive by route", so that's architecturally prevented; StrictMode's double-invoke is the realistic case and is covered. Noted in the test rather than papered over with module-level state.

812 tests pass, typecheck clean, lint at its 24-warning baseline.

andymai:
Fixed in the latest commit — and a pointed catch. **This session opened by fixing exactly this bug class** (#2591: Save silently discarding a baseplate name), and I wrote a fresh one three files away.

`handleNew` swallowed the error: `setCreating(false)` ran, the button re-enabled, and the modal sat there with no design and no explanation. It now toasts and stays open so the click can be retried, matching the naming form.

One correction to the rationale, for the record: *"Every parallel path shows a toast on failure; this one doesn't"* isn't quite right about this file — **`BaseplateLibraryModal` has no toasts at all**. `handleRename` and `handleDuplicate` don't even check their results:

```ts
const handleRename = useCallback(async (id, name) => {
  await renameDesign(id, name);   // result ignored
  await refresh();
```

So the silence was the file's existing convention, not a deviation from it. The finding still stands on its own merit and I've fixed the path I added. The other three are the same class and pre-existing — I've left them out rather than quietly widen this diff; happy to file a follow-up if you want them.

The new test was confirmed to fail against the silent version before the fix went in. 776 baseplate tests pass, typecheck clean, lint at its 24-warning baseline.

PR #2626, opened 2026-07-18
fix(worker): recover the brepkit kernel after a borrow-flag poison

andymai:
## Problem

The brepkit WASM kernel is a per-worker singleton whose wasm-bindgen borrow flag can be **permanently stranded** (brepkit task #14). Two distinct bugs do it:
1. a `raw_vec` **"capacity overflow"** Rust panic — wasm is `panic=abort`, so the trap never releases the `&mut self` borrow;
2. a **non-panic stranding** — a JS exception unwinding through a `&mut self` method leaves the `BorrowMut` guard undropped (this one recurs as the arena regrows).

Once stranded, every later kernel call throws `recursive use of an object detected ... unsafe aliasing`, and no Rust code can reset the flag — the worker stays bricked until the page reloads. In the export-integrity matrix, one victim cascaded across ~180 scenarios, masking the true results.

## Fix

- **`wasmInstantiator.recoverBrepkitKernel()`** — recreates the kernel; a fresh `BrepKernel` re-registered as the default reroutes all later brepjs ops (they resolve `getKernel()` per call). `getLastBrepkitPanic()` exposes brepkit's panic-hook text to detect a panic-abort poison that left no catchable JS error on the poisoning request.
- **`workerContext`** — the central `runGeneration`/`runExport` catch paths detect the poison signature (the "recursive use" message, or a recorded trap), drop the now-stale cache handles that index the dead arena (bin/baseplate/mesh-imprint), and recreate the kernel so the **next request runs healthy**. Guarded to the brepkit kernel; occt-wasm/manifold untouched.
- **export-integrity harness** — the same recovery in `beforeEach`, which unmasks the true matrix.

## Impact

Unmasks the export-integrity matrix: **241 pass / 137 fail** (was a masked **165 / 213**) — **+76 scenarios recovered**, revealing the real geometry backlog that the cascade had hidden. The ~137 real failures are the known deferred parity families (scoop, screw base, solid cutouts, …), now visible and workable.

Verified: tsc clean, eslint clean on all three files, export-integrity smoke green. brepkit-side foils are unaffected (this is entirely tool-side).

Note: root-fixing the underlying `capacity overflow` in brepkit is tracked separately — recovery is still required regardless because bug #2 (the non-panic stranding) recurs.

---
## Summary by cubic
Recover the `brepkit` WASM kernel after a borrow-flag poison to stop worker bricking and cascading "recursive use" failures. The kernel is recreated and stale caches are cleared so the next request runs clean; export-integrity now shows 241 pass / 137 fail (+76 recovered). Hardened recovery and added tests.

- **Bug Fixes**
  - Detect poison via "recursive use"/"unsafe aliasing" errors or `getLastBrepkitPanic()`.
  - Add `recoverBrepkitKernel()` and `getLastBrepkitPanic()` in `wasmInstantiator`; cache `brepkit-wasm` and clear its panic-hook on load; recreate and re-register via `brepjs`.
  - Clear stale caches on recovery: `clearAllCaches`, `clearBaseplateCaches`, `clearMeshImprintCache`.
  - Apply recovery in `workerContext` catch paths for generation/export, and in export-integrity `beforeEach`.
  - Scoped to the `brepkit` kernel; `occt-wasm` and `manifold` unchanged.
  - Hardened recovery: guard cache eviction/recovery so disposers can’t block kernel recreation; add unit tests for `recoverBrepkitKernel` and `getLastBrepkitPanic`; ensure `loadBrepkit` clears stale panic; update `brepkit-wasm` mock.

<sup>Written for commit 1627d7ba8ac3875c51dc48c0c250d95f521ac65f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/generation/worker/generators/binGenerator.scenario.export-integrity.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.export-integrity.test.ts
@@ -59,7 +59,14 @@ let poisoned = false;
  * all later brepjs ops (they resolve getKernel() per call).
  */
 async function recoverBrepkitKernel(): Promise<void> {
-  clearAllCaches();
+  // Best-effort: dispose calls shape.delete() on the dead arena (brepkit's
+  // dispose is a no-op, safe while poisoned), but never let a throwing disposer
+  // block the fresh kernel below — that recreation is what actually recovers.
+  try {
+    clearAllCaches();
+  } catch {
+    /* fresh kernel below restores health regardless */
+  }
   const { registerKernel, BrepkitAdapter } = await import('brepjs');
   const brepkitWasm = await import('brepkit-wasm');
   const kernel = new brepkitWasm.BrepKernel();
--- a/src/features/generation/worker/generators/binGenerator.scenario.export-integrity.test.ts
+++ b/src/features/generation/worker/generators/binGenerator.scenario.export-integrity.test.ts
@@ -24,16 +24,64 @@ import { isOk } from '@/core/result';
 import { parseSTLBinary } from '@/shared/generation/stlParser';
 import { buildParams } from './__kernel-tests__/scenarioTypes';
 import { ALL_SCENARIOS } from './scenarios';
-import { setLastSolid } from './shapeCache';
+import { setLastSolid, clearAllCaches } from './shapeCache';
 import type * as BinExporterModule from './binExporter';
 
 let exportBin: typeof BinExporterModule.exportBin;
+
+// ── brepkit kernel-poison recovery ─────────────────────────────────────────
+// The brepkit WASM kernel is a per-worker singleton whose arena grows across
+// every scenario. Two distinct bugs strand its wasm-bindgen borrow flag once
+// enough state accumulates (see brepkit task #14):
+//   1. a `raw_vec` "capacity overflow" Rust panic (panic=abort → the trap never
+//      releases the &mut self borrow), first observed on the magnet+halfSockets
+//      scenario;
+//   2. a NON-panic stranding (a JS exception unwinding through a &mut self wasm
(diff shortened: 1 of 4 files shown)


andymai:
Addressed the review feedback in 1627d7ba8:

- **Unguarded cache disposal on the recovery path** (workerContext.ts:80 + test harness): wrapped cache eviction and `recoverBrepkitKernel()` in try/catch so a throwing `shape.delete()` disposer can never block the essential step — recreating the kernel. brepkit's dispose is a no-op today, but defense-in-depth matters most on a last-line-of-recovery path.
- **No test coverage for the new functions** (wasmInstantiator.ts:167): added unit tests for `recoverBrepkitKernel` (no-op/false before load; recreate + re-register after) and `getLastBrepkitPanic` (undefined before load; surfaces the hook value after), plus `loadBrepkit` clearing a stale panic on load. Also fixed the `brepkit-wasm` mock, which lacked the panic-hook exports `loadBrepkit` now accesses (that was the earlier core-2/3 CI failure).
- **`kernel as any` without eslint suppression** (export-integrity.test.ts:66): held — the Quality/lint gate is green and eslint is clean locally; test files permit `no-explicit-any`. Adding a suppression here would reintroduce the `unused-eslint-disable` error that was removed earlier.

Issue #2631, opened 2026-07-19
Bin generator - allow for removable dividers in both vertical and horizontal directions concurrently


kcslb92:
Firstly, what an incredible tool - thank you so much for pouring your time and energy into making this, it's super neat!

I'm designing Gridfinity bins to hold an assortment of fasteners. Each of the fasteners come in different lengths. Some of them are long enough that it requires going beyond 1 unit but not long enough to consume 2 whole unitss (eg those that are 45mm, 50mm, and longer) whilst others are short enough that they'd fit within a single unit (some of which are quite short at ~6mm, 8mm, 10mm, etc); in some cases it makes sense to "pair" them up. In the below example, you could use a given "column" to hold (various examples below) different combinations to consume the 2 units in length:

- 10mm + 15mm + 20mm + 35mm
- 20mm + 25mm + 35mm
- 50mm + 30mm
- 55mm + 25mm

I will try and explain below with a diagram.
In order to make effective use of the space, it would be helpful to be able to use divider inserts at various intervals in _both directions_. The tool currently caters to either vertical or horizontal on its own, but not both at the same time.
In order for this to work, you'd need to introduce the ability to have divider "slots" on the dividers themselves.
A picture speaks a thousand words:

Imagine a 6 wide, 2 high Gridfinity bin like this:

<img width="1209" height="466" alt="Image" src="https://github.com/user-attachments/assets/f124cdce-c159-4704-9fb9-c7d4113d7818" />

Let's say we wanted to split it so that each unit has the ability to have 2 dividers (12 total), it would look something like this (divider slots in red):

<img width="1578" height="602" alt="Image" src="https://github.com/user-attachments/assets/90b50b1f-a578-48d9-8b5b-27e0d6321a55" />

Ordinarily, each divider would be splitting an entire 2 units in length, see the below blue dividers):

<img width="1575" height="597" alt="Image" src="https://github.com/user-attachments/assets/a848e567-7d32-49dd-b592-3e5a3a843bcc" />

What would be nice, is if you could also specify a number to split in the other direction, in this case - "rows", of which there are 4 potential compartments for this example:

<img width="1602" height="565" alt="Image" src="https://github.com/user-attachments/assets/f36faef8-028f-4e3c-8c90-20dd9c92c159" />

Now, you could print dividers that are designed to be split in either direction, there would be a myriad of different combinations to demonstrate, for this example I'll show a couple of examples.

[...]

PR #2635, opened 2026-07-19
feat(designer): per-compartment cross dividers with face receptacles

### LINKED ISSUE #2634: Bin generator - short per-compartment dividers that slot into the long dividers
Follow-up to #2631 (shipped in #2632).

#2632 added bidirectional removable dividers: both directions can be enabled at once, and the full-length pieces interlock at crossings via egg-crate cross-lap notches. That covers the core ask, but not the second part of the original request.

## What's missing

In the #2631 fastener example, each column gets *different* split positions (e.g. one column split 10+15+20+35mm, the next 50+30mm). With full-length pieces, an inserted cross divider spans every column at that position, so split positions are shared across the whole bin.

Independent per-column subdivision needs **short, compartment-width dividers**: receptacle slots cut into the faces of the long dividers (mirroring the wall slot profile), so a square divider can drop into any single compartment at any populated position.

## Sketch of the work

- Cut wall-slot-style receptacles into the long divider faces at the cross-axis positions (same `calculateSlotPositions` + slot width as the wall slots, so everything stays aligned)
- New short divider piece type: spans one compartment plus tab engagement on each end
- Reconcile with the cross-lap notches from #2632 — a face receptacle and a cross-lap notch at the same position need to coexist or be mutually exclusive per position
- Original issue also floated dovetail-joined divider segments for odd widths; probably a separate decision

## Open questions

- Fit: the short divider engages plastic-on-plastic (divider face receptacles) rather than bin walls — may want its own tolerance default
- Export: short dividers add a third unique piece per direction; combined export labeling needs extending

andymai:
Closes #2634

Follow-up to #2632. That PR made both divider directions usable at once via full-length egg-crate pieces, but split positions were shared across the whole bin. This adds the second engagement style from the original request (#2631): **per-compartment short dividers**, so each column can be subdivided independently.

## What's new

- **Cross divider style** control (both-directions mode only): *Interlocking* (existing egg-crate lap joints, still the default) or *Per compartment*
- In per-compartment mode, the chosen **full-length direction** keeps long dividers that now carry **vertical receptacle grooves on both faces** at every cross position — mirroring the wall slot profile (same opening width, same positions, slide-in from the top on either side)
- The other direction becomes **short pieces**: an interior piece (divider-to-divider, receptacle tabs both ends) and an edge piece (wall-to-divider, wall-slot tab on one end)
- Groove depth is 30% of divider thickness per face, leaving a 40% web; dividers thinner than **1.2mm** degrade gracefully to interlocking mode — geometry and UI share one resolver so they can never disagree
- Divider pieces now carry their export labels from the builder (`divider-*-compartment`, `divider-*-compartment-edge` join the existing axis labels)
- 3D preview segments the short-direction ghosts per compartment and shows the short reference piece
- New optional `slotConfig.crossStyle` / `longAxis` fields — absent fields read as the old behavior, so persisted layouts are unaffected

**The bin body is untouched**: wall slots and lip cutouts are identical in both modes, so no mesh baselines change.

## Verification

- Real-kernel test extended: grooved long piece loses exactly `2n × grooveWidth × height × grooveDepth` (divergence-theorem mesh volume vs analytic), short pieces match their analytic lengths, all vertices finite
- New pure-math tests for spans/lengths/receptacle depth/mode resolution; insert-mode unit tests for piece labels, groove cutter counts, and both fallback paths
- Slotted scenario suites pass unchanged; i18n adds 7 keys across all 12 locales (`check:i18n` clean)

---
## Summary by cubic
Adds a per-compartment cross divider option so each column can be subdivided independently; long dividers get face receptacles and short pieces slide in. Defaults and the bin body are unchanged. Closes #2634.

- **New Features**
  - Cross divider style in both-axes mode: Interlocking (default) or Per compartment
  - Per compartment: the chosen long axis keeps full-length dividers with vertical face grooves; the other axis uses short interior and edge pieces that slide into grooves and wall slots
  - Long-axis picker, piece-length readout (shows short-piece lengths); preview shows per-compartment ghosts and both short reference pieces; warns if divider thickness < 1.2 mm
  - Divider count reflects per-compartment topology; short pieces aren’t generated when the short axis has no rows
  - Edge short pieces use symmetric min-tab ends so they fit either way round
  - Divider exports include labels from the builder (adds `*-compartment` and `*-compartment-edge`)
  - Optional `slotConfig.crossStyle` and `slotConfig.longAxis`; shared resolver degrades thin dividers to Interlocking; tests added

- **Refactors**
  - Preview segmentation/merge math deduplicated; lap-mode builder collapsed into one loop
  - STEP assembly uses per-axis grid pitch (matches STL path)
  - Hardened insert-mode math and config safety: interior short pieces size to the minimum inter-divider gap; persisted cross-mode values are clamped to safe defaults

<sup>Written for commit 988b1f5a0fe40d0b519c011086a9b082a2b9f289. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2635?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Re: Greptile review of 3084e1ae0 — all three issues are addressed as of 988b1f5a0:

1. **Missing groovePositions guard** — fixed in eb03fd6b8 (the quality-pass commit): the insert branch now emits short pieces only when `groovePositions.length > 0`, with the same guard mirrored in the SlotConfigurator readout and the preview, plus a unit test for the asymmetric case (`emits only the long piece when the short axis has no rows`).

2. **applyCuts double-delete concern** — no change: `fuseAll` does not take ownership of its inputs in brepjs. The pre-existing `slotBuilder.ts` pattern registers both the input cutters and the fused result for disposal (see `buildSlotCutsInScope`), and this exact lifecycle is exercised by the disposal regression kernel tests and the `dividerCrossLap` kernel test (real OCCT, repeated cuts, no use-after-free).

3. **Interior span / uniform-spacing assumption** — fixed in 988b1f5a0: `calculateShortDividerSpans` now takes the minimum inter-divider gap, so a single interior piece is safe in every compartment regardless of spacing; covered by a non-uniform-positions test.

Copilot's `longAxis` clamping comment is also fixed in 988b1f5a0 (`resolveCrossDividerMode` clamps unvalidated persisted values instead of indexing with them).

PR #2642, opened 2026-07-20
fix(baseplate): replace flat connector key with locking puzzle-lobe dogbone

### LINKED ISSUE #2637: Dovetail Key doesn't print usable dovetails even at 0.2 mm
I was excited to discover the Dovetail Key system.  However the Print Fit Sample simply doesn't work.

I can download the 3MF and slice it for a Prusa MK3S using PrusaSlicer.  I can see the dovetails and key in the slicer and they look correct.

But when they actually print, the dovetails are nearly rectangles and so naturally the key doesn't stay put in the dovetail, in ANY of the size variants.

Image

I am printing using PETG at 0.2mm layer height.

andymai:
## Problem

The `dovetailKey` connector ships a 3mm bowtie key whose undercut is 0.3mm/side (`TONGUE_BASE_HALF` 1.0 → `TONGUE_TIP_HALF` 1.3). That is below what FDM can reproduce: nozzle-radius corner rounding (~0.2mm), first-layer squish (~0.2mm), and the press-fit clearance (0.075mm) swallow the taper entirely, so printed grooves come out near-rectangular and the key pulls straight out. This is the same failure mode the codebase already documented for the integral dovetail when the `puzzle` style was introduced (#2241) — it was just never applied to the key.

Reported in #2637: PrusaSlicer shows correct dovetails, but the print holds nothing at any fit-sample offset (PETG, 0.2mm, MK3S).

## Fix

- The key is now two **puzzle lobes mirrored across the waist** into a dogbone (1.0mm/side undercut — the profile the integral `puzzle` style already prints reliably). Reach stays `TONGUE_PROTRUSION`, so bed-budget/bbox math is untouched.
- Key-mode seams cut **puzzle grooves** (`makePuzzleGroove`) with the existing `DOVETAIL_KEY_CLEARANCE` press fit, so cavity constraints (cell severing, inter-cell wall width) are exactly what the shipped puzzle style already proved in the field.
- The wider head reaches into the open socket mouths near the slab top, so the key is **relieved against the four junction bin feet** — the same socket-relief the snap clip uses (`relieveClipForSockets` generalized into `relieveForNeighborSockets`, shared by both).
- UI label: "Dovetail key" → **"Puzzle key"** (+ icon, print guide, all 10 locales). The style id stays `dovetailKey` so saved designs keep working.

## Compatibility

Plates printed with the old trapezoid grooves need a reprint to accept the new key. Those keys demonstrably never held, so there is no working fit to preserve — the alternative (a 5th connector style) would freeze a known-broken profile in the picker forever.

## Tests

- **New `connectorKeyFit.test.ts`** (real OCCT): the key (1) seats in the two-groove cavity with a clearance fit, (2) captures >1mm³ of bearing volume when the plates are pulled apart, (3) clears all four nominal junction bin feet post-relief while retaining its locking material.
- **New constants guard**: pins the key's undercut above the FDM swallow budget (corner rounding + squish + clearance) that the bowtie fell below — a future retune can't silently regress the lock into decoration.
- Scenario tests updated to the dogbone footprint (2×`PUZZLE_HEAD_HALF` cross-axis); watertight-STL assertions unchanged and passing.

Fixes #2637

---
## Summary by cubic
Replaced the failing bowtie “dovetail key” with a locking dogbone made from two mirrored puzzle lobes, and switched key-mode grooves to puzzle grooves for a reliable FDM lock. Keeps existing designs working, fixes pull-out on printed keys, and updates UI wording to “Puzzle key”. Fixes #2637.

- **Bug Fixes**
  - New key: dogbone profile (two puzzle lobes), same reach, press-fit via `DOVETAIL_KEY_CLEARANCE`.
  - Key mode now cuts puzzle grooves (`makePuzzleGroove`); seams remain symmetric.
  - Added socket-foot relief to the key using shared `relieveForNeighborSockets` (also used by the snap clip).
  - Preview/icon/UI: dogbone icon, label “Puzzle key”, part name `puzzle_key`, print guide text updated; 10 locales updated.
  - Tests: new fit/capture test (`connectorKeyFit.test.ts`), constants guard for printable undercut, updated scenarios and sample tray; preview mesh matches worker geometry.

- **Migration**
  - Plates printed with the old trapezoid grooves must be reprinted to accept the new key. The style id stays `dovetailKey`, so saved designs continue to load.

<sup>Written for commit e54a87a51d14426f619eb1aefceb5defb5929ac1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2642?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kevinp2:
Thanks for implementing the new puzzle key.  What is the best place to obtain the key itself?  From the fit sample?

I printed the fit sample today and the key fit best with the larger tolerances, i.e. +0.05 and +0.1 mm.  It looks like still larger tolerances are available on the web page going up to 0.3 mm.

kevinp2:
A question:  could the **detached margin connector** feature include the **puzzle key** option?  Since it already allows the puzzle option?  See example below for a 440 mm x 70 mm drawer.  Let me know if you would prefer that I submit a new enhancement request.

image

PR #2648, opened 2026-07-20
feat(supporters): recency signals, supporter messages, and find-your-bin

andymai:
## What this adds

The /supporters page gains a few touches that make it feel alive and let supporters find themselves:

- **"Joined this month"** plus a soft **arrival glow** on the most recent supporter's bin, so the wall visibly changes between visits.
- A muted **support-over-time sparkline** tucked in a corner.
- **Supporter messages**: the public thank-you note someone leaves now shows on their bin's tape card, plus a slow-rotating quote near the CTA so the warmth is visible without hunting for a bin.
- A **purpose line** by the Ko-fi button (the "why give").
- **Find your bin**: type a name and the camera flies to that supporter's bin.

Every bin stays equal (no tiers, no amounts shown); the newest glow is the only distinction.

## Data and privacy

- Each supporter record now carries a join date and, only for publicly-named supporters, a short message. Ko-fi's public opt-in gates the whole shout-out, so an opted-out supporter surfaces neither name nor message.
- The payment amount is folded into a collect-only per-currency total that is never tied to a person and never served to the page.
- Historical dates, messages, and totals are backfilled once from a manual Ko-fi transaction export via `pnpm seed-supporters --csv <export>`; messages and emails are written to Redis only, never to disk. Recency features stay hidden until real dated data loads.

## Notes

- New i18n keys are translated across all 10 locales.
- Camera fly-to lives in its own `CameraFocus` component; the 3D behaviour is covered by a Playwright visual check, the rest by unit tests.
- Needs a one-time `pnpm seed-supporters --csv ...` run against production Redis for existing supporters to gain dates and messages.

---
## Summary by cubic
Adds recency signals, public supporter messages, and “find your bin” to the /supporters page, with API/webhook/storage updates to store join dates/messages and track per‑currency totals. Improves accuracy and stability of names/messages and the fly-to flow.

- New Features
  - “Joined this month,” a soft arrival glow on the newest bin, and a corner sparkline of new supporters.
  - Rotating public supporter message near the CTA plus a short purpose line.
  - “Find your bin” search that flies the camera via `CameraFocus`.
  - Hardening: messages only with public names; exact‑match‑first search with a stable fallback; duplicate‑name notice only for exact duplicates; fly‑to waits for controls; clears stale find status when the search is submitted empty; minor i18n fix.

- Migration
  - API response changes from `{ named[], anonymousCount }` to `{ supporters: SupporterRecord[] }` (each may include `joinedAt` and a public `message`).
  - Redis donor values change to compact JSON records `{n,t,m}`; legacy bare names are still parsed. New collect‑only totals hash: `supporters:totals` (minor units, never served).
  - Webhook now stores join timestamp and a filtered message only for public supporters; emails and amounts are never tied to a person. Each payment increments `supporters:totals`; renewals create no new bin.
  - One‑time backfill: run `pnpm seed-supporters --csv <Transaction_All.csv>` to enrich `seed:*` records with dates/messages and write the totals baseline; the baseline only seeds when `supporters:totals` is empty (re‑runs never clobber live counters).

<sup>Written for commit 33e03b4c077e704098304e6316127d426e480a8a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2648?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed the review feedback in f9461a53b:

- **`isSupportersData` message/name invariant** — the guard now rejects any record carrying a `message` without a non-empty `name`, and `buildSupporterBins` never carries a message onto an anonymous bin. So an opted-out supporter's words can't surface even if a bad payload slips through. Added tests for both.
- **Find-your-bin matching** — prefers an exact (case-insensitive) name match; only falls back to a substring, and sorts that fallback (shortest name, then alphabetical) so the pick is stable across the shuffle instead of random.
- **Multiple-match status** — the "1 of N named X" message now only fires for genuine exact duplicate names, using the actual (canonical) name; partial matches just focus the best candidate with no misleading count.
- **Swedish "Just joined"** — fixed the fragment to "Precis gått med", matching the "gick med" phrasing.
- **`CameraFocus` nonce timing** — the request is only marked handled once it's actually applied; if OrbitControls isn't registered yet in animated mode, the effect leaves it unconsumed and re-runs when controls mount.

andymai:
Fixed the Sparkline accessibility comment in b1df1d5f7 — the SVG stays aria-hidden while the labeled `<figure>` conveys the chart to assistive tech, so the comment now matches behavior.

andymai:
Both addressed in 33e03b4c0: `handleFind` now clears the status on an empty submit, and the seed script only writes the `supporters:totals` baseline when that hash is empty (skips with a note otherwise) so a re-run can't clobber webhook-incremented counters.

PR #2696, opened 2026-07-21
feat(designer): magnetic-retention lids and tray tops (#2694)

andymai:
Closes #2694.

## What

Two new lid options in the bin designer.

### Magnetic lids
A new **Attachment** mode selector (Friction / Click rails / Magnetic) replaces the implicit rails-only retention. In **Magnetic** mode, four corner magnets bond the lid onto the bin: the bin grows inward corner posts and the lid grows matching bosses, each housing a press-fit magnet. The two magnets meet across the seated stacking-lip plane.

- Dedicated magnet diameter/depth fields, independent of the bin's base magnets.
- Because magnets hold independent of the lip, magnetic lids no longer trip the rail-grip compatibility blockers (e.g. wall cutouts on all four sides) — they get their own checks instead (custom-shape unsupported, magnet too deep for the bin).
- A press-fit / glue / alternating-polarity print hint surfaces in the panel.

### Tray tops
When a lid has no stackable baseplate, an optional recessed tray with a raised rim can be shelled into the top face so small items rest on the closed lid without sliding off. Depth and rim wall are configurable. Mutually exclusive with the stackable top.

## Geometry notes
- Corner posts grow **inward**, so the 42mm footprint stays drawer-safe (verified in a scenario test).
- The bin posts and lid holes derive from one shared corner-position helper, guaranteeing they line up.
- Export orientation is auto-selected: friction/magnetic tray lids export tray-up (recess prints clean); everything else keeps the floor-down flip for a support-free mating cavity.

## Rendered result

Magnetic lid (exploded) — four corner magnet pockets in both the lid and the bin's corner posts:

Magnetic + tray — corner magnets plus the recessed tray with its rim.

_(Screenshots captured locally via the new `lid-magnetic-tray-visual` e2e spec; attaching to the PR thread.)_

## Tests
- New unit test for the shared corner-position helper (XY-coincidence contract) and `usesMagneticLid` predicate.
- New real-WASM scenario tests: magnetic lid, bin posts, tray, magnetic+tray — all structurally valid/watertight; footprint unchanged.
- Attachment-aware compatibility tests (blocker downgrade + new magnetic checks).
- Migration + server-validation tests for the new fields.
- New `lid-magnetic-tray-visual` Playwright spec drives the panel and asserts each option changes the preview.

## Compatibility
- Migration derives the attachment mode from legacy rails (any rail on → click rails, else friction) and backfills magnet/tray defaults, so existing designs are unchanged.
[...]

Issue #2694, opened 2026-07-21
Magnetic lids


ScuttleSE:
**Problem**
Makes the lid...stick to bins

**Proposed Solution**
Toggle to allow magnets on both the lid and the bin, akin to this: https://makerworld.com/en/models/985321-box-with-lid-and-magnets-3x5x10u-gridfinity

**Alternatives Considered**
None really...

PR #2698, opened 2026-07-21
fix(designer): correct magnetic lid geometry and panel layout (#2694)

andymai:
Follow-up to #2696, fixing three issues reported on the first magnetic-lid cut.

### 1. Bin holders were full-height columns → corner gusset pads
The bin's magnet holders rose as solid cylinders from the interior floor to the rim, eating the whole corner. They're now **gusset pads**: a solid anchored to the two interior corner walls at the top that cantilevers inward (a printable overhang) to house the magnet, leaving the interior clear.

### 2. Lid magnet bosses bumped through the top surface → magnets tucked under the floor
The lid bosses protruded above the top face as visible bumps. The magnet now sits **directly under the floor plate** (top surface is smooth, no bumps or crease circles), with the boss hanging into the mating cavity below.

The bin pad's **top** now mates with the lid boss's **bottom** across one seat gap. Both are derived from the *same* resolved lid inputs (`resolveLidInputs`), so the bin and lid planes can't drift.

### 3. Magnet controls were buried below the fold → regrouped under the selector
The magnet diameter/depth steppers rendered after the stackable/tray toggles, far from the attachment selector. They now render **directly beneath** the Friction/Click rails/Magnetic selector, next to the mode they belong to.

## Verification
Rendered before/after in the designer (screenshots below): bin now has flush corner gusset pads at the top instead of columns; lid top is smooth; pads are aligned to mate. Friction lids correctly have no pads.

- New regression test: the bin pads are **height-independent** (a returning full-column bug would scale the added geometry with bin height).
- All existing magnetic/tray scenario, compat, migration, validation, and panel tests pass.

---
## Summary by cubic
Fixes magnetic lid geometry so the top stays smooth and the bin interior stays clear, and moves magnet controls next to the attachment selector. Adds a blocker for bins too small to fit corner magnets and clamps pad pockets to avoid floor punch‑through.

- **Bug Fixes**
  - Lid magnets sit under the floor; bosses hang into the cavity. Top face is flush.
  - Bin holders changed from floor-to-rim columns to top corner gusset pads with upward pockets. Interior and footprint stay clean; added geometry is height‑independent.
  - Shared placement/inset logic keeps magnets coaxial and clear of the stacking lip; bin pad top mates with the lid boss bottom from the same resolved lid inputs.
  - Compatibility: blocks bins that can’t fit corner magnets (too small for the required inset).
  - Safety: clamps pocket depth to the recessed pad capacity and pads never cut through the floor.

<sup>Written for commit 3699698ee3d9917710e8411a4de6ced10e468935. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/bin-designer/components/panel/LidSection/LidSection.tsx
+++ b/src/features/bin-designer/components/panel/LidSection/LidSection.tsx
@@ -194,8 +194,121 @@ export function LidSection() {
       </div>
 
       {/* Attachment method (#2694) — how the lid retains onto the bin. One of
-          friction / click rails / magnetic; the sub-controls below key off it. */}
+          friction / click rails / magnetic. The mode's own controls render
+          directly below so they stay next to the selector. */}
       <AttachmentSelector value={state.attachment} onChange={handlers.setAttachment} t={t} />
+
+      {/* Magnetic retention (#2694) — dedicated corner magnets bond the lid to
+          the bin. Independent of the bin's base magnets. */}
+      {state.attachment === 'magnetic' && (
+        <div className="space-y-2">
+          <StepperField
+            label={t('binDesigner.lid.retentionMagnetDiameter')}
+            unit="mm"
+            value={state.retentionMagnetDiameter}
+            onChange={handlers.setRetentionMagnetDiameter}
+            onStep={(delta) =>
+              handlers.setRetentionMagnetDiameter(
+                state.retentionMagnetDiameter + delta * state.retentionMagnetStep
+              )
+            }
+            min={state.retentionMagnetDiameterMin}
+            max={state.retentionMagnetDiameterMax}
+            step={state.retentionMagnetStep}
+            size="md"
+            aria-label={t('binDesigner.lid.retentionMagnetDiameterAria')}
+            commitMode="deferred"
+          />
+          <StepperField
+            label={t('binDesigner.lid.retentionMagnetDepth')}
+            unit="mm"
+            value={state.retentionMagnetDepth}
+            onChange={handlers.setRetentionMagnetDepth}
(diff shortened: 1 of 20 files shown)


andymai:
Addressed all review findings in 3699698ee:

- **Greptile — small-bin magnet bounds:** added a `magnetBinTooSmall` compatibility blocker. When a bin's half-extent is smaller than the magnet's corner-pad inset (lip clearance + boss wall + diameter), the four magnets would overlap through the centre, so the lid is now blocked with a clear message instead of producing crossed-over placements. Covered by two new compat tests.
- **Copilot — pocket depth could dig through the floor:** the pocket depth is now clamped to what the recessed pad can hold (no forced 0.4mm), and the pad bottom is additionally clamped to the interior floor, so a marginal magnet can never punch a pocket through it.
- **Copilot — stale header comment:** the lid-magnet header now describes the floor-hidden magnet (top at `-topThickness`), not the old `-(depth + ceiling)` interface.

Note: `pnpm run lint` may show two `import()`-type errors in `booleanStage.ts` locally — that's uncommitted CAPCUT debug instrumentation in my working tree, not part of this branch (HEAD and main both lack it), so it isn't in the PR.

Issue #2690, opened 2026-07-21
Swappable labels: physical Cullenect fit verification + nozzle-aware socket clearance


andymai:
Spin-off of the last two follow-ups from #2666, which closed with everything implementable shipped. Both of these are blocked on things outside the codebase:

## 1. Physical click-fit verification against a printed Cullenect/gflabel plate

Every socket dimension is pinned against the standard's parametric source (`Cullenect.scad`, v2.0.0: plate `42·U − 6` × 11 × 1.2mm; socket = plate + 0.3mm total clearance; 0.2×0.4mm ribs 0.2mm above the floor) and merge-gated on geometric verification — scenario tests measure the pocket from the generated mesh. What has **never** happened is a physical print test: click a real Cullenect or gflabel plate into a generated socket (and a generated plate into a third-party socket).

The fit-calibration card from #2666 (bin designer → socket mode → "Print fit test") is the tool for this: print it plus one ecosystem plate and record which offset clicks best on a calibrated 0.4mm printer. If nominal (+0.00) is not the winner, consider shifting `LABEL_SOCKET_CLEARANCE_MM` rather than leaving every user to dial `plateFitOffset`.

## 2. Automatic nozzle scaling for socket clearance

`effectiveLabelSocketClearance` already accepts a nozzle parameter, but the bin GENERATE path deliberately passes `undefined`: params flow persisted-store → worker → mesh-persistence hash, so injecting printer state would either pollute synced designs or leave stale mesh caches (decision recorded on #2666). Revisit only when the bin generation path gains a settings seam; until then the per-design `plateFitOffset` (−0.2..+0.5mm) covers wide-nozzle calibration manually.

andymai:
Closing per request.

**Part 2 (nozzle-aware socket clearance)** shipped in #2727 (merged): sockets auto-scale their pocket clearance to the active print nozzle, applied across preview, layout 3D, all export paths, and the plate-width/fit math, plus the fit-calibration card.

**Part 1 (physical Cullenect/gflabel click-fit verification)** was not performed — it needs a print test on real hardware. If nominal (+0.00) doesn't click best on a calibrated 0.4mm printer, the follow-up is shifting the baseline `LABEL_SOCKET_CLEARANCE_MM` rather than leaving every user to dial `plateFitOffset`. Reopen if/when that print test is worth tracking.

Issue #2666, opened 2026-07-20
Bin designer: Cullenect-compatible click-in swappable labels + batch label plate export


andymai:
**Problem**

Label tab text is part of the bin geometry: relabeling a bin means regenerating and reprinting the entire bin. When contents change or a bin is swapped for a different size, the label can't move with the contents.

A user requested a dovetail-style slot in the label tab so separately printed labels could slide in and transfer between bins, with batch label generation in the tool — referencing this whole-row approach: https://makerworld.com/de/models/2695839-gridfinity-fastener-labeling-bar (per-bin would be better, since bins get rearranged independently).

**Proposed Solution**

Adopt the **[Cullenect](https://github.com/CullenJWebb/Cullenect-Labels) swappable-label standard** (click-in, not dovetail — see "Why not a dovetail" below) and generate both sides of it:

_1. Socket mode on the label tab_

A new label tab mode alongside engraved/embossed text: the tab face gets a Cullenect-compatible socket instead of printed-in text.

- Plate spec: width `42·U − 6` mm (1U = 36, 2U = 78, 3U = 120), 11 mm tall, 1.2 mm thick; perimeter latch clicks past 0.2 × 0.4 mm ribs in a socket sized plate + 0.3 mm clearance, 1.2 mm deep.
- Placement: per-compartment (existing tab placement machinery), quantized to the **largest standard plate width that fits the compartment group**; columns too narrow for a 1U plate fall back to a single bin-spanning socket. Standard widths keep every plate transferable between bins.
- Socket profile is an enum (`socketStyle`) so future profiles (e.g. a slide channel) can be added without rework.

_2. Label plate generator_

Generate the matching plates in the tool: text from per-compartment texts / bin labels, embossed or debossed 0.2–0.4 mm (snapped to whole layer heights so a single filament swap produces clean two-color text), multi-color via the existing `paint_color` 3MF pipeline. Plates printed flat, no supports.

_3. Batch export_

- Whole-layout ZIP gains a `labels/` folder: enumerate placed bins → linked designs → compartment texts + bin labels, dedupe, pack plates onto bed-sized sheets, plus a guide entry in the manifest.
- Bin designer gains an "Export label plates" action for a single design.

Because the socket is Cullenect-compatible, users can also print plates from existing generators (e.g. [gflabel](https://github.com/ndevenish/gflabel), with its hardware icon library) or use plates they already own.

**Why not a dovetail slide?**

[...]

PR #2727, opened 2026-07-22
feat(labels): scale label sockets to the print nozzle (#2690)

andymai:
## What & why

Swappable-label sockets (#2666) were tuned for a 0.4mm nozzle. On a wider nozzle the pocket clearance stayed the same, so plates fit too tight to click in. This scales a socket's pocket clearance to the active print nozzle automatically, so plates keep clicking in on 0.6/0.8mm nozzles.

Addresses the nozzle-scaling follow-up in #2690.

## How it works

The nozzle (`settings.printSettings.nozzleSizeMm`) is merged into generation **transiently** (`withSocketNozzle`) and folded into the mesh cache key, but is **never persisted** into the design — so a printer's nozzle doesn't sync to your other devices or ride along in a shared/exported design. `withSocketNozzle` returns the input unchanged for non-socket bins and at/below the 0.4mm baseline, so nothing that doesn't need new geometry is affected and no existing cached mesh is invalidated.

Applied consistently across every boundary that shows or cuts a socket, so the width picker can never offer a plate the widened socket rejects:

- Live designer preview (+ draft, cache key, export-warm, and a regen on nozzle change)
- Layout 3D linked-design preview
- Bin export (single + split) and whole-layout ZIP export
- Plate-width / plate-count / fit-warning math
- The fit-calibration card scales to the nozzle too, so its winning offset transfers to bins printed on that same nozzle

The fit-test dialog copy was refined to explain the nozzle/material dependency and that you can test against a real Cullenect/gflabel plate you already own.

## Verification

- New mesh-measuring scenario: a 3×1 socket on a 0.6mm nozzle produces a measurably wider pocket (the assert derives the expected clearance from the nozzle, so it fails if generation ignores it). No existing geometry snapshot changed.
- New unit tests for the settings seam.
- typecheck, lint, knip, module boundaries, and all i18n checks pass.

## Still open in #2690 (not in this PR)

Physical click-fit verification against a printed Cullenect/gflabel plate — a print test, not a code change. If nominal (+0.00) isn't the best fit, the next step is shifting the baseline `LABEL_SOCKET_CLEARANCE_MM`. #2690 stays open to track it.

---
## Summary by cubic
Scales swappable‑label socket pocket clearance to the active print nozzle so plates keep clicking in on 0.6/0.8mm, without persisting the nozzle in designs. Regenerates only when the nozzle changes geometry and quantizes nozzle values in cache keys to avoid misses. Addresses Linear #2690.

- **New Features**
  - Introduced `withSocketNozzle` to inject `settings.printSettings.nozzleSizeMm` transiently into generation and the bin mesh cache key; never persisted in `BinParams`.
  - Applied nozzle scaling across live preview/draft/cache key/export warm, linked-design previews, single/split/ZIP exports, plate width/count/fit-warning math, and layout plate export; socket bins auto‑regenerate on nozzle changes.
  - Fit test card now takes the nozzle and scales nominal clearance; updated help copy; added unit tests and a 0.6mm scenario that measures a wider pocket.

- **Bug Fixes**
  - Worker feature cache now keys sockets by nozzle in socket mode to avoid stale narrow sockets after a nozzle change; text mode remains nozzle‑invariant.
  - Hardened `withSocketNozzle`: ignores disabled labels, strips a stale injected nozzle when reverting to the 0.4mm baseline, and is idempotent; added regression tests.
  - Nozzle-change regeneration now gates on `withSocketNozzle`, so bins only re-generate when the nozzle affects geometry (labels disabled, non-socket, or ≤0.4mm baseline skip).
  - Quantized nozzle folded into linked-design and plate-count cache keys to avoid float-noise fragmentation and stale/missed entries.

<sup>Written for commit 8882d9f49a846114cb20ff32c0cbed0bf1f0332a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2727?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/bin-designer/components/panel/LabelTabsSection/useLabelTabsSection.ts
+++ b/src/features/bin-designer/components/panel/LabelTabsSection/useLabelTabsSection.ts
@@ -1,6 +1,7 @@
 import { useCallback, useMemo } from 'react';
 import { useShallow } from 'zustand/react/shallow';
 import { useDesignerStore } from '@/features/bin-designer/store';
+import { useSettingsStore } from '@/core/store';
 import { DESIGNER_CONSTRAINTS } from '../../../constants';
 import { binDimensions } from '@/features/bin-designer/utils/binDimensions';
 import { useTranslation } from '@/i18n';
@@ -50,6 +51,7 @@ export function useLabelTabsSection() {
     }))
   );
   const t = useTranslation();
+  const nozzleSizeMm = useSettingsStore((s) => s.settings.printSettings.nozzleSizeMm);
 
   const labelStatus = getFeatureStatus(params, 'label');
 
@@ -419,9 +421,12 @@ export function useLabelTabsSection() {
 
   const socketPlan = useMemo(() => {
     const { innerW } = binDimensions(params);
-    const clearanceMm = effectiveLabelSocketClearance(undefined, label.plateFitOffset);
+    // Nozzle-scaled to match the worker's cut so pickers/warnings can never
+    // disagree with the geometry (a wider nozzle can drop a compartment from
+    // a 2U plate to 1U).
+    const clearanceMm = effectiveLabelSocketClearance(nozzleSizeMm, label.plateFitOffset);
     return planLabelSockets(compartments, innerW, clearanceMm);
-  }, [params, compartments, label.plateFitOffset]);
+  }, [params, compartments, label.plateFitOffset, nozzleSizeMm]);
 
   // Socket segment disabled when no standard plate fits anywhere, even
   // bin-spanning \u2014 sockets are never emitted at nonstandard widths.
--- a/src/features/bin-designer/hooks/useExport.ts
+++ b/src/features/bin-designer/hooks/useExport.ts
@@ -30,6 +30,7 @@ import { useToastStore } from '@/core/store/toast';
 import { calcMaxGridUnits } from '@/core/constants';
 import { DEFAULT_PATTERN_SCALE } from '@/features/bin-designer/types';
 import { getActiveBridge, bridgeManager } from '@/shared/generation/bridge';
+import { withSocketNozzle } from '@/shared/generation/socketNozzle';
 import { generateFileName } from '@/features/bin-designer/utils/fileNaming';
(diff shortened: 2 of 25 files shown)


andymai:
Addressed the review findings in d9d3914:

**Greptile (4/5) — nozzle missing from feature cache key.** Fixed: `labelTabsFeature.cacheKey` now folds `nozzleSizeMm` into its socket-mode key part, so two nozzles can't share a stale in-worker feature-cache entry. Text-mode keys are unchanged (no churn). Added a regression test.

**Copilot — `withSocketNozzle` stale nozzle on re-wrap.** Fixed: the helper now strips an already-injected nozzle when reverting to baseline, so re-wrapping an already-merged params object (e.g. a draft path) can't carry a wide-nozzle pocket into a 0.4mm generation. Regression test added.

**Copilot — injects even when `label.enabled` is false.** Fixed: the guard now also requires `label.enabled`, so a disabled socket-mode label no longer churns the cache key.

**Copilot — `planLabelPlateExport` 4-arg call sites.** The new nozzle param is intentionally required (the one production caller always passes it); updated the test call sites to pass the baseline nozzle rather than loosening the signature.

andymai:
Second re-review pass addressed in 8882d9f (the three earlier inline comments were re-anchored duplicates already fixed in d9d3914):

**Copilot — nozzle-change regen guard too broad.** The `useGeneration` effect now gates on `withSocketNozzle` itself (the single source of truth for whether the nozzle changes geometry), so it skips regeneration when labels are disabled, the bin isn't socket mode, or both the old and new nozzle are at/below the 0.4mm baseline — instead of regenerating for any socket-mode bin.

**Copilot — raw nozzle in cache-key strings.** Quantized the nozzle folded into the linked-design and plate-count cache keys (`toFixed(3)`) so float noise can't fragment the key into avoidable misses. (The worker feature cache key already quantizes.)

PR #2878, opened 2026-07-27
feat(bin-designer): add a spacer mode so bins of different heights finish flush

### LINKED ISSUE #2869: Bin Spacer Mode
**Problem**
Height compatibility between different bins.

**Proposed Solution**
Add a spacer mode to the Bin Designer. Essentially allows creating a 1U Frame to stack below existing bins.
See example here:
https://www.printables.com/model/1169027-gridfinity-base-risersspacers-1x1

andymai:
Closes #2869.

## Problem

There's no way to make a short bin's rim reach a tall one's. You can approximate it by designing a plain bin as a riser, but you then pay for a floor you don't want and a full interior you can't use.

## What this does

Adds a **Spacer** toggle to the bin designer's Base section: a floorless riser you stack under a bin. Feet, envelope, and stacking lip are untouched, so **the stacking arithmetic is inherited from a normal bin** — a 2u spacer under a 2u bin reaches the top of a 4u one, with no new height rules to get wrong. The scenario suite pins that: a spacer's bounding box matches a plain bin of the same size and height.

## How the floor comes out

Almost entirely reuse. A spacer derives `dimensions.lightweight = true` and asks `lightweightBaseBuilder` for a new `'through'` open direction, which is just **zShift 0**: the inner foot is already a uniform-`wallThickness` lateral offset of the outer foot at every depth, so without the ±wt shift the cut leaves the cup open at *both* ends. The floor-opening slug that the lite path already builds punches the body floor over each cup, so the whole cell becomes a clean through-hole.

**What holds a multi-cell spacer together is the webbing between the cups** (each keeps its own wall, and adjacent walls meet) plus the perimeter. That's the load-bearing invariant here — get it wrong and interior feet come loose as separate solids — so a 3×3 export case guards it.

## Deliberate exclusions

- **Magnets / screws.** A pad standing inside a through-hole would have nothing to attach to. `deriveDimensions` suppresses `withMagnet`/`withScrew` for a spacer (the same guard shape `isFlat` already uses, so a crafted share payload can't produce a disconnected solid), and the constraint engine greys out the toggles. A scenario asserts a spacer carrying `magnet_and_screw` generates identically to a plain one.
- **Floor-dependent features** — compartments, scoop, labels, inserts, cutouts, floor pattern, slotted/solid styles, flat base. Declared once in `CONSTRAINT_RULES`, so the UI explains why and enabling a spacer clears them in one resolve. Wall features (patterns, cutouts, text, handles) still apply.

Not a new `BinStyle`: it's a base property, which keeps it composable with the wall features and out of the ~20 `style` consumers and the exhaustiveness gate.

## Threading

Every place that special-cases the shelled base takes the spacer too. The one that matters most is **`canBinUseDirectMesh`** — the procedural draft models a solid floor, so without the gate the instant preview would show a floor that isn't in the exported mesh (and would stay on screen if BREP failed). Also: the split floor-scarf (no floor to bite into), the generation timeout budget, split alignment, and the floor-pattern fit + estimate. `floorPatternApplies` in the worker needed nothing, since it already keys off the derived `lightweight`.

## Tests

- `binGenerator.scenario.spacer.test.ts` — 9 real-kernel cases: 1u with and without lip, the 3×3 multi-cell attachment guard, the envelope invariant, fractional footprint, half sockets, attachment-style suppression, differs-from-lite (proves the floor was actually punched), and the export fuse.
- `context.test.ts` — implies the shelled path, preserves wall/total height and envelope, suppresses hardware, inert on a flat base, and never shares a shell cache entry with the lite bin it resembles.
- `engine.test.ts` — enabling a spacer clears the floor-dependent set in one resolve, keeps the lip and feet, is blocked by each interior feature, and can always be turned back off.
- `useBaseSection.test.ts` — write-through, the magnet it clears, and the greyed-out reason.

Registered the new scenario module in `scenarios/index.ts` — `binGenerator.scenarioCoverage.test.ts` enforces that.

---
## Summary by cubic
Adds a Spacer mode to the bin designer: a floorless riser you stack under a bin so different heights finish flush. Feet, envelope, and lip stay the same, so a 2u spacer under a 2u bin reaches a 4u bin. Closes #2869.

- **New Features**
  - UI: Spacer toggle in Base with a short hint; stays selectable and auto-clears conflicting features; disabled only on a flat base.
  - Generation: New 'through' open direction in the lightweight base so cups open both ends; spacer implies lightweight and removes the floor through all cells while keeping feet and lip.
  - Constraints: Enabling spacer clears floor-dependent features and attachment hardware; disables lightweight since it’s implied; clears slotted/solid and divider slots; separate reasons for interior vs style; only flat base blocks it.
  - Preview/Export: Excluded from direct-mesh; split alignment disabled like lightweight; floor-pattern fit/estimate skipped; generation timeout adjusted.
  - Caching: Distinct shell cache key for spacers.
  - Tests & i18n: Spacer scenarios and unit tests added; locales updated with spacer strings and separate style/interior reasons; README notes added.
  - Validation: Share payload validation requires `base.spacer` to be boolean; API docs list `spacer` among validated keys.

<sup>Written for commit 33d27ea978891c00c30178eaf23bfb5e92fe5c46. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2878?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Greptile flagged a corrupted string in `ja.json` — that one doesn't reproduce. The file parses clean, contains zero U+FFFD replacement characters, and the seven added Japanese strings render correctly (`スペーサー（床なし）` etc.). Likely its own rendering of the full-width parentheses. The other two findings were real and are fixed above.

PR #2887, opened 2026-07-27
feat(bin-designer): add 26 label plate icons across fasteners and tooling

andymai:
Stacked on #2886 — review that first; this PR's diff is just the catalog.

Delivers the icons [discussion #2877](https://github.com/andymai/gridfinity-layout-tool/discussions/2877) asked for, using the SVG pipeline from #2886. The allowlist goes from **6 to 32**.

**Fasteners (18)** — the original six plus socket cap screw, set screw, self-tapping screw, threaded rod, split pin, lock washer, wing nut, square nut, threaded insert, eye bolt, thumb screw, standoff.

**Tooling & consumables (14)** — drill bit, hex key, tap, countersink, utility blade, spring, o-ring, bearing, magnet, zip tie, saw blade, file, end mill, circlip.

## Every silhouette was rendered at print size before being wired in

The readable band is 7.8 mm and the relief is 0.4 mm. At that size a lot of hardware looks like other hardware, so the set was reviewed as rendered geometry, not as a list of names, and revised:

- A plain capsule reads as a **battery**, so the set screw carries thread down both edges and a hex socket.
- A bare cone reads as an **arrowhead**, so the countersink gained a blunt pilot tip and its cross-flute.
- A horseshoe without pole faces reads as an **arch**, so the magnet gained banded poles.
- `bearing` and `sawBlade` needed balls and teeth respectively, or they were indistinguishable from `washer`.
- `dowelPin` and `rivet` were cut — each was a near-twin of a neighbour and earned nothing at 7.8 mm.

## Two kernel traps, now guarded and documented

**Zero-length edges.** Repeating the current path point is invisible in a browser and fatal in the kernel (`makeLineEdge: construction failed`). The ridge helpers take a `continues` flag for callers already sitting on that point.

**Tangent arcs can collapse.** The horseshoe's outer arc runs into its leg edges at a tangent. Built as a true arc it encloses **23.51** instead of **45.60** — while the bounding box stays correct. Same silent-failure shape as the washer bug in #2886, and again only an area assertion sees it. That icon uses a polyline arch, locked by a ±1% area check.

Both are written up in `docs/label-icons.md` so the next contributor doesn't rediscover them.

## Other notes

- Hexagon vertices round to 6dp rather than 4, so the ported nut still matches its pre-port rendered width to 5dp.
- "Magnet" is identical to English in cs/de/nb/sv; allowlisted as a loanword, matching the existing convention.
- The existing tests iterate `LABEL_PLATE_ICONS`, so all 32 are covered automatically. The plate sweep now runs 64 builds in ~14s — comfortably inside its budget, and it is already split into its own file so CI can shard it.

## Known interim rough edge

The picker is still the flat text `<Select>`, which is poor at 32 entries — you scroll a word list with no preview. That is deliberate: the grid picker with SVG previews is PR 3, so this PR stays reviewable as a pure catalog change.

---
## Summary by cubic
Adds 26 new label plate icons for fasteners and tooling, expanding the catalog from 6 to 32 and improving reliability in the SVG-to-geometry pipeline. Updates icons, tests, docs, and i18n.

- New Features
  - Expanded icon set to 32: 18 fasteners and 14 tooling/consumables.
  - Added new IDs to `LABEL_PLATE_ICONS` with silhouettes tuned for 7.8 mm x 0.4 mm print readability.
  - Updated i18n strings across locales and allowlisted `binDesigner.plateIcon.magnet`.

- Bug Fixes
  - Guarded kernel pitfalls: zero-length segments and tangent arcs collapsing; added a ±1% area check for the horseshoe magnet.
  - Documented pitfalls in `docs/label-icons.md`.

<sup>Written for commit 02e449c372f2e33e8c56e581285f5c066a62b3ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Superseded by #2890 — GitHub auto-closed this when #2886 merged and its base branch was deleted, and won't reopen a PR whose base is gone. Same commit, rebased onto `main`, with the review feedback from here already applied.

PR #2903, opened 2026-07-27
fix(bin-designer): persist pending design edits when leaving the designer

### LINKED ISSUE #2895: Export bins from within layout not exporting correctly
When in the layout screen, after having built all the bin designs and having them linked to the layout, I wish to export all bins as STLs or 3MF.
From the bin list on the right, when clicking the export layout (down arrow icon) I get the correct dialogue and export works fine, but although the bins are all exported, they are not the current version. In most instances I get multiple copies of a standard 3x4 bin which is my default, irrespective of what is actually in the layout. They are named correctly but the actual bin in the stl/3mf does not correspond to the designed bin.

The same happens when exporting individual bins from the layout page by clicking on the bin and selecting export button from the right (beside edit design button).

The only way to get the correct bin exported is by entering each design in the bin module and exporting from there.

Running on latest chrome version on a mac.

andymai:
Fixes #2895.

## The bug

Exporting bins from the layout — either the bulk export from the bin list or the single-bin export in the inspector — produced stale geometry, typically the design's creation-time defaults. Filenames were right, the meshes were not. Exporting the same design from inside the bin designer worked.

## Root cause

Layout export resolves designs from IndexedDB via `loadDesign`; the designer exports from its live store. That asymmetry only matters if the two disagree — and they did.

`useAutoSave` is mounted inside `DesignerPage`, so leaving the designer unmounts it. Its effect cleanup aborted the pending save:

```js
return () => { abortToken.current = true; clearTimeout(timerRef.current); };
```

Between the 1s debounce and `waitForGenerationComplete` (up to 5s, plus a 150ms render settle), there was a window of roughly six seconds in which navigating back to the layout silently discarded the write. Cutout-heavy designs sat at the wide end of that window because generation is slow.

The designer store is module-level, so it survives the route change and kept rendering the user's edits. Nothing surfaced the loss until an export read the actual stored record.

A second, quieter half: even when the write landed, `if (abortToken.current) return;` sat *after* `updateDesignParams`, so a superseded save could persist params while skipping the registry upsert and the `design-saved` event. The planner palette and the linked-bin mesh cache key off `registry.updatedAt`, so the layout could keep a stale mesh for a design that had in fact been rewritten.

## The fix

Split `persist` (the write plus the bookkeeping every downstream consumer depends on) out of the abortable thumbnail phase, and flush unsaved params on unmount.

The flush passes `undefined` for the thumbnail rather than capturing one — the canvas is already gone, and `updateDesignParams` treats `undefined` as "leave the stored thumbnail alone", so `useBackgroundThumbnailRegen` refreshes it later.

## Tests

Five regression tests in `useAutoSave.test.ts`, all of which fail on `main`:

- persists a pending edit when unmounted inside the debounce window
- persists an edit whose save was still waiting on generation
- updates the registry and notifies linked bins when flushing
- does not write when nothing changed since the last save
- does not write for an unsaved design

The 8 pre-existing tests still pass, as do `src/features/bin-designer/hooks` and `src/features/design-linking` (429 tests). Typecheck and lint clean.

## Not covered

Closing the tab mid-window still loses the edit — `beforeunload` cannot await an IndexedDB write. That is a separate, narrower gap than route navigation, which is what #2895 describes.

---
## Summary by cubic
Persist pending design edits when leaving the designer to fix stale geometry in layout exports, and dedupe unmount saves to prevent duplicate writes and events.

- **Bug Fixes**
  - Split persistence from the abortable thumbnail step so registry updates and `design-saved` events are never dropped.
  - Flush unsaved params on unmount; pass `undefined` thumbnail and let background regen refresh it later.
  - Dedupe saves: track last-saved and in-flight params/config to skip identical writes and events; still flush if the in-flight save has older params.

- **Tests**
  - Added regression tests covering flush during debounce/generation, registry updates and event emission, in-flight save dedupe, and no-op cases.

<sup>Written for commit 344c4fee28783791289a3a9763a7d545732f908e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2903?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/bin-designer/hooks/useAutoSave.test.ts
+++ b/src/features/bin-designer/hooks/useAutoSave.test.ts
@@ -349,6 +349,71 @@ describe('useAutoSave', () => {
       unsubscribe();
     });
 
+    it('does not duplicate a write for the save already in flight', async () => {
+      let resolveWrite: ((value: unknown) => void) | null = null;
+      vi.mocked(DesignerStorage.updateDesignParams).mockReturnValue(
+        new Promise((resolve) => {
+          resolveWrite = resolve;
+        })
+      );
+      useDesignerStore.setState({ currentDesignId: 'existing-id' });
+
+      const { unmount } = renderHook(() => useAutoSave());
+
+      act(() => {
+        useDesignerStore.getState().setParam('width', 4);
+      });
+
+      // Let the save reach updateDesignParams, where it now blocks.
+      await advanceThroughSave();
+      expect(DesignerStorage.updateDesignParams).toHaveBeenCalledOnce();
+
+      await act(async () => {
+        unmount();
+      });
+
+      expect(DesignerStorage.updateDesignParams).toHaveBeenCalledOnce();
+      expect(resolveWrite).not.toBeNull();
+    });
+
+    // The in-flight save carries older params, so the newer edit still needs
+    // flushing rather than being swallowed by the dedupe above.
+    it('still flushes when the in-flight save is for older params', async () => {
+      let resolveWrite: ((value: unknown) => void) | null = null;
+      vi.mocked(DesignerStorage.updateDesignParams).mockReturnValueOnce(
+        new Promise((resolve) => {
+          resolveWrite = resolve;
+        })
+      );
+      useDesignerStore.setState({ currentDesignId: 'existing-id' });
+
+      const { unmount } = renderHook(() => useAutoSave());
+
+      act(() => {
+        useDesignerStore.getState().setParam('width', 4);
+      });
+      await advanceThroughSave();
+      expect(DesignerStorage.updateDesignParams).toHaveBeenCalledOnce();
+
+      mockUpdateDesignParams();
+      act(() => {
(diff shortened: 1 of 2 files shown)


andymai:
Good catch on the duplicate write — fixed in 344c4fe.

`persist` now records what it's writing in an `inFlight` ref, and the unmount flush skips only an *exact* match on params + config. A save carrying older params still gets flushed, which is the case that matters: that's the stale-export scenario this PR exists to fix.

Two tests cover both sides — `does not duplicate a write for the save already in flight` and `still flushes when the in-flight save is for older params`.

PR #2972, opened 2026-07-30
chore(quality): type-check the test suite behind a baseline ratchet

andymai:
## Summary

**Test files have never been type-checked.** `tsconfig.test.json` correctly `include`s every `*.test.ts(x)`, `src/test/**`, and `__kernel-tests__` file — but it is not listed in the root `tsconfig.json` `references`, and `tsgo -b` only walks the reference graph. Running it directly surfaces **9,622 errors across 382 files**.

The gap stayed invisible because the config *is* wired into `eslint.config.js` as a parser project. Typed lint rules use the type-checker for their own queries but never report TypeScript's own diagnostics, so `pnpm lint` and `pnpm typecheck` both pass on a test that asserts against a property which does not exist.

## Why it matters

A worked example from `src/core/storage/ShareService.test.ts:439`, in a test named "verify bin data preserved":

```ts
expect(decodedBin!.position).toEqual(originalBin!.position);
expect(decodedBin!.size).toEqual(originalBin!.size);
```

`Bin` has no `position` or `size`. Both are `expect(undefined).toEqual(undefined)` — they pass, and the share encode/decode round-trip is verified for nothing about bin geometry.

The errors split into two very different populations:

| Class | Count | Files |
|---|---|---|
| Branded-type (`number` → `GridUnits` / `HeightUnits` / `Mm`) | 8,355 | 383 (107 exclusively) |
| Everything else — stale tests, vacuous assertions, wrong mocks | **1,269** | **276** |

Other confirmed members of the second group: `api/lib/validation.test.ts` calls a function with a missing required argument at 18 sites; `designerValidation.test.ts` exercises a `spacer` field that no longer exists; `core/labs/features.test.ts` asserts on a removed `comingSoon` property.

## Changes

Wiring the reference in directly would fail on the whole backlog, so this PR installs the gate first and leaves the burn-down to follow-ups.

- **`scripts/check-test-types.ts`** — runs `tsgo -p tsconfig.test.json --noEmit`, aggregates diagnostics per file, and compares against a checked-in baseline. Keyed on **per-file counts**, which survive line moves and message rewording but still catch the case that matters (a new required field on `Bin` silently breaking 40 test files).
- **Exact match, not a ceiling.** Decreases fail too, with `pnpm run check:test-types -- --update` as the one-command remedy. A ceiling would let a fixed error be silently replaced by a new one; exact matching forces the number to ratchet monotonically down.
- **`scripts/test-type-baseline.json`** — 9,622 errors, 382 files.
- **`tsconfig.test.json`** — adds `scripts/**/*.test.ts`. Those script tests and the sources they import type-check clean today (baseline total unchanged), so it costs nothing and it covers this PR's own gate script.
- Wired into `pnpm run quality`, the CI Quality job, and pre-commit. Runs in ~3.4s.

### Retirement path

When the baseline reaches zero: add `{ "path": "./tsconfig.test.json" }` to the root references, delete the script and the baseline. Documented in `CLAUDE.md` and in the script header.

### Deliberately not in scope

`scripts/` is also globally ignored by ESLint, and `tsconfig.scripts.json` only includes `scripts/sync-admin/**`. Extending it to all of `scripts/` surfaces 273 errors, but ~90% are app files pulled in transitively and failing on a DOM-less `lib` — a config-shape problem needing its own design, not a burn-down.

## Test plan

- [x] `scripts/check-test-types.test.ts` — 16 tests covering diagnostic parsing (including that indented elaboration lines aren't double-counted), baseline validation, and all four drift directions
- [x] Gate verified end-to-end in both directions: a new test file with one error fails with `✗ Test files that newly fail to type-check`; a stale-high baseline fails with `→ Progress! Lock it in with…`
- [x] Coverage verified real — a deliberate error introduced into `scripts/check-test-types.ts` itself is caught via its sibling test
- [x] Corrupt-baseline path returns a clear message instead of an unhandled `JSON.parse` throw
- [x] Guard against a compiler crash being read as "zero errors" and passing vacuously
- [x] `pnpm run quality` — 0 errors (61 pre-existing `max-lines` warnings unchanged)
- [x] Full pre-commit hook chain passes

---
## Summary by cubic
Adds a ratcheted type-check gate for the test suite to block new type errors and burn down the backlog; initial baseline is 9,621 errors across 381 files. The check builds the English locale on demand and requires exit 0 before accepting a zero-error result to avoid false passes.

- **New Features**
  - Added `scripts/check-test-types.ts`: runs `tsgo -p tsconfig.test.json --noEmit`, tallies per-file errors, and compares against a checked-in baseline.
  - Baseline at `scripts/test-type-baseline.json` (9,621 total, 381 files); exact match required. `--update` refuses increases or newly failing files unless `--allow-increase`.
  - Integrated into CI “quality” and `.husky/pre-commit` via `pnpm run check:test-types` (reported as “test types”).
  - The check builds `src/i18n/locales/en.json` when missing to avoid TS2307 in fresh clones and CI.
  - Included `scripts/**/*.test.ts` in `tsconfig.test.json` so the gate script and its tests type-check.
  - Safer gate: requires exit 0 before trusting a zero-error result; detects when the compiler never ran; rejects unreadable/malformed baselines (NaN/negative/fractional counts, arrays); treats baseline `note` as optional; normalizes `total`; and parses diagnostics robustly (ignores indented elaborations, normalizes Windows paths, handles parentheses).

- **Migration**
  - Run `pnpm run check:test-types` locally; inspect with `pnpm exec tsgo -p tsconfig.test.json --noEmit`.
  - After fixes, update the baseline with: `pnpm run check:test-types -- --update` (use `--allow-increase` only if increases are intended).
  - When the baseline reaches zero, add `{ "path": "./tsconfig.test.json" }` to the root `tsconfig.json` and remove the script + baseline.

<sup>Written for commit ab15aa0c6563e4b43bf6dc7fb58c986b3f74fbc2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2972?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
### Review round-up

Five issues were found after the initial push — four by review, one by CI. All are the same family the gate exists to prevent: **a path where the check passes without having actually verified anything.** Worth listing because they shaped the final design.

1. **Spawn failure read as a clean compile.** `runCompiler` returned only stdout+stderr, so `ENOENT` yielded `"\n"` and the "no parseable diagnostics" guard never fired. Harmless while the baseline is non-empty; a silent pass once it reaches zero, which is the goal. Split on `execFileSync`'s `status`.
2. **Non-zero exit with unparseable output read as zero errors.** A crash or a changed diagnostic format exits non-zero and parses to nothing. Under `--update` that would have written an empty baseline and switched the ratchet off. Now requires exit 0 before trusting a zero result.
3. **`--update` could paper over a regression.** The header promised "must only go down" while the code overwrote unconditionally. It now refuses to raise any file's count; `--allow-increase` is the deliberate override.
4. **`NaN` counts silently disabled the ratchet.** `typeof NaN === 'number'` passed validation, and every comparison against `NaN` is false — so `to > from` and `to < from` are both false and the file read as *matching*. Counts must now be non-negative integers.
5. **The generated locale.** CI failed twice on `src/i18n/context.tsx` TS2307. `src/i18n/locales/en.json` is generated and gitignored; a CI step fixed it, then I removed that step on a review suggestion that `prepare` made it redundant — **which was wrong**. `pnpm install` skips `prepare` when it has nothing to do, verified locally ("Already up to date", file not regenerated) and in the CI log, which contains zero occurrences of the generator's output. The check now generates the file itself, covering CI, fresh clones, and pre-commit with one mechanism.

One review suggestion was **not** taken: the claim that the diagnostic regex mis-parses paths containing parentheses. The non-greedy group expands until the rest of the pattern matches, so `src/foo (copy).test.ts(10,5): error TS…` already parses correctly. Added a regression test rather than changing working code.

Test count went 11 → 29. The narrowing bug in `classifyRun` was caught by this gate running against its own source, which is a reasonable smoke test of the whole idea.

PR #2987, opened 2026-07-30
test(bin-designer): fix all test-type errors in src/features/bin-designer

andymai:
## Summary

Drives the `src/features/bin-designer/**` slice of the test-type-baseline ratchet (`scripts/test-type-baseline.json`, #2972) from **344 errors across 63 files to 0**. `tsconfig.test.json` was never referenced from the root `tsconfig.json`, so these test files were never type-checked until now.

## What changed

**Mechanical (bulk of the diff):**
- Wrap raw ID/unit literals with branded constructors (`designId()`, `binId()`, `layerId()`, `categoryId()`, `gridUnits()`, `heightUnits()`, `mm()`) from `@/core/types`.
- Spread `DEFAULT_GENERATION_STATE` / `DEFAULT_UI_STATE` into store fixtures that predated later-added required fields (`isDraft`, `perfHistory`, a dozen `DesignerUIState` fields) instead of hand-listing every property.
- Spread `DEFAULT_BIN_PARAMS.<field>` (wall side configs, handles, slotConfig, label) instead of hand-rolled partial `WallCutout`/`HandleSide`/`SlotConfig`/`LabelTabConfig` literals.
- `screen.getByRole(...)` results needing `.value`/`.checked`/`.disabled` now use the generic form (`getByRole<HTMLInputElement>(...)`) instead of assuming `HTMLElement`.
- `cellMask.cells` fixtures switched from `boolean[]`/`number[]` to the real `(0 | 1)[]` type, at the correct half-bin resolution (`MASK_CELLS_PER_UNIT = 2` → a 2×2-unit bin needs a 4×4 mask, not 2×2).

**Real bugs/gaps this revealed** (production behavior unchanged except where noted):

- **`paramMigration.ts`** (production, type-only change): `MigrateParamsInput` only exposed the *current* `WallPatternConfig`/`FeatureColorConfig` shape for `wallPattern`/`featureColors`, but `migrateParams` is explicitly a legacy-data coercion boundary (mirrors the existing `WallConfig | LegacyWallConfig` pattern already used for `walls`) and needs to accept malformed/legacy shapes (e.g. `dividers: 'yes'`). Added `LegacyWallPatternConfig` and folded the already-internal `LegacyFeatureColorInput` into the re-declared input type via `Omit` + union — a plain intersection of two differently-shaped optional properties does **not** loosen the type (verified empirically with a standalone repro). Split the wallPattern merge into a `wallPatternRaw` staging variable so it stays in the legacy shape until the existing coercion block rebuilds a real `WallPatternConfig`. No behavior change: confirmed by the full 104-test `paramMigration` suite, and two now-provably-unnecessary `as never` casts in the test file were removed as a direct consequence.
- **`DesignerStorage.test.ts`**: `StorageError` union access to `.key` needed narrowing by `code` (a soft `expect(error.code).toBe(...)` doesn't narrow the type). Added a small `expectStorageCorrupted` helper. Mutation-tested by reverting the `depth`/`filletRadius` legacy-field strip in `migrateParams`' handle migration and confirming the rewritten `'depth' in result.handles` assertion (previously an illegal `as Record<string, unknown>` cast) now fails as expected.
- **`binDownloadHelpers.test.ts`**: `FaceGroupData.feature` was a stale field name — the real field is `tag`.
- **`useAutoSave.test.ts` / `useBackgroundThumbnailRegen.test.ts`**: `let resolveX: ((v: unknown) => void) | null` captured from `new Promise((resolve) => { resolveX = resolve })` doesn't type-check against the resolver's real `Result<...>` type — fixed by typing the holder to the real resolved type instead of `unknown`.
- Various `SessionUser` test fixtures were missing the now-required `provider` field.

No forbidden escape hatches (`any`, `@ts-expect-error`, `as unknown as X`, `Record<string, unknown>` widening) were introduced anywhere in this diff.

## Test plan

- [x] `pnpm exec tsgo -p tsconfig.test.json --noEmit` — 0 errors under `src/features/bin-designer/`
- [x] `pnpm exec vitest run src/features/bin-designer` — 333 files / 3590 tests, all passing
- [x] `pnpm run check:test-types` — baseline updated (9612 → 9268), passes
- [x] Mutation-tested the rewritten `DesignerStorage.test.ts` handle-strip assertion (confirmed it fails when the production strip is reverted)

---
## Summary by cubic
Fixes all test type errors in `src/features/bin-designer/**`, cutting this slice from 344 errors to 0 and lowering `scripts/test-type-baseline.json` total from 4231 to 3887. Improves test fixtures and types; no runtime behavior changes.

- **Refactors**
  - Use branded ID/unit constructors from `@/core/types` (`designId`, `binId`, `layerId`, `categoryId`, `gridUnits`, `heightUnits`, `mm`).
  - Spread `DEFAULT_GENERATION_STATE`, `DEFAULT_UI_STATE`, and `DEFAULT_BIN_PARAMS.*` (handles, walls, slotConfig, label) in fixtures.
  - Use `getByRole<...>` generics when reading `.value`/`.checked`/`.disabled`.
  - Align tests with current types/props (e.g., `exportFileNameConfig`, preview controls `xray`/`projection`, split-view mesh entries, `aria-label` names) and correct cell masks to `(0 | 1)[]` at half-bin resolution.

- **Bug Fixes**
  - `migrateParams` input accepts legacy/malformed `wallPattern` and `featureColors`; stage `wallPatternRaw` before coercion. No behavior change; tests verify.
  - Narrow error unions by `code` and add small helpers in storage and STL parsing tests; remove unsafe casts.
  - Update export tests to use `faceGroups.tag` (was `feature`).
  - Fix promise resolver types in autosave/thumbnail tests to concrete `Result<...>`.
  - Add missing fixture fields (e.g., `provider` on `SessionUser`).

<sup>Written for commit bccfa2da740484919168c4f8fe93abbf6a00ea46. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2987?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Reviewed the one production change (`paramMigration.ts`) — it holds up, with one nuance worth recording.

**The type widening is right.** `migrateParams` is explicitly a legacy-coercion boundary, and `MigrateParamsInput` already modelled that for `walls` (`WallConfig | LegacyWallConfig`) but not for `wallPattern`/`featureColors`. Typing the input as `Partial<BinParams>` demands the caller already match the *current* strict shape, which defeats the purpose of a migration entry point. Using `Omit` + union rather than a plain intersection is also correct — an intersection would keep the strict member and narrow to it.

**The runtime rebuild is at least as strict as before, not looser:**

- `pattern` gains an explicit `typeof rawPattern === 'string'` guard ahead of the `VALID_WALL_PATTERNS.has()` check. Same outcome as before for a non-string (falls back to `'honeycomb'`), now stated rather than relying on a `Set.has()` miss.
- `scale` clamp and `dividers: … === true` are unchanged.

**One small behaviour change, in the right direction** — worth calling out since the description says "no behaviour change":

`enabled` is now `wallPatternRaw.enabled ?? false`. Previously it came through the spread, so a legacy save carrying `enabled: undefined` produced `{ ...{ enabled: false }, ...{ enabled: undefined } }` → `enabled: undefined`, i.e. a `WallPatternConfig` whose `enabled` was not a boolean. It is now `false`. That is a correctness improvement for a field the type says is boolean, and the 104 `paramMigration.test.ts` cases cover it — just not literally "no change".

PR #2994, opened 2026-07-30
test(analytics): fix all test-type errors in src/shared/analytics

andymai:
## Summary

- Zeroes out the `src/shared/analytics/**` slice of the `pnpm run check:test-types` baseline ratchet: **1353 → 0 errors** across 10 files (`scripts/test-type-baseline.json` total 3887 → 2534). No other file's count changed.
- ~94% of the errors were branded-type mismatches (`GridUnits`/`HeightUnits`/`Mm`/`BinId`/`LayerId`/`CategoryId` from `@gridfinity/branded-types`, re-exported via `@/core/types`) where test fixtures used raw number/string literals instead of the brand constructors (`gridUnits()`, `heightUnits()`, `binId()`, `layerId()`, `categoryId()`, `mm()`). Fixed by branding fixture values at construction, mostly via small local factory helpers (`createBin`/`makeBin`/`makeDrawer`/`makeLayer`/`makeCategory`) instead of repeating constructors at hundreds of call sites.
- No event names, payload keys, or `expect(...)` assertions were changed — verified via diff that every assertion line is byte-identical to `HEAD`.

## Two real (non-branding) issues found and fixed

- **`posthog.test.ts`**: 6 places wrote `preferences: { enabledFeatures, dismissedFeatures: [] }` into `useLabsStore`. `dismissedFeatures` does not exist on `LabsPreferences` (`{ enabledFeatures, lastModified, version }`), and the object was missing both required fields — the test was installing a structurally invalid store value at runtime. Replaced with `{ ...createDefaultLabsPreferences(), enabledFeatures }`, which `computeLabsMetrics` (only reads `enabledFeatures`) is unaffected by. `src/shared/hooks/useCrossTabSync.test.ts:260` has the same latent `dismissedFeatures` field — left alone, it's outside this PR's slice.
- **`purposeInference.test.ts`**: its local `createTestLayout` helper was missing the `Layout`-required `version`/`name` fields, and mapped `label: b.label` / `notes: b.notes` (`string | undefined`) directly onto `Bin`'s required `string` fields. Added the missing fields and `?? ''` fallbacks; verified every consumer (`purposeInference.ts`, `shared/utils/bins.ts`) already treats an empty-string and an undefined label identically.

## Left alone (reported, not fixed)

- `gapAnalysis.ts`'s internal `createOccupiedGrid` takes `layerId: string` while its only caller passes a branded `LayerId` — a private-helper typing inconsistency with no behavioral effect.

## Test plan

- [x] `pnpm exec tsgo -p tsconfig.test.json --noEmit | grep -c "^src/shared/analytics/"` → `0`
- [x] `pnpm exec vitest run src/shared/analytics` → 19 files / 486 tests passed
- [x] `pnpm run typecheck` → clean
- [x] `pnpm run check:test-types -- --update && pnpm run check:test-types` → baseline updated (3887 → 2534), then matches
- [x] Mutation-tested the one materially-rewritten assertion (`errorFilters.test.ts` passthrough fixture): broke the production early-return, confirmed the test failed, restored it
- [x] `prettier --check` and `eslint` clean on all changed files

---
## Summary by cubic
Eliminates all `check:test-types` errors in `src/shared/analytics/**` by branding test fixtures and tightening test helpers. The slice drops from 1353 → 0 errors, reducing the baseline to 1847 (was 3200). No assertions or event names changed; analytics tests all pass.

- **Refactors**
  - Replaced raw literals with branded constructors from `@/core/types`: `gridUnits()`, `heightUnits()`, `mm()`, `binId()`, `layerId()`, `categoryId()`.
  - Added small test factories (`makeBin`, `createBin`, `makeDrawer`, `makeLayer`, `makeCategory`) to centralize branding and cut repetition.
  - Minor test tidy-ups: stricter `vitest` mock types, `vi.stubEnv('DEV', false)`, safer `Reflect.deleteProperty`, and explicit `STAGING_ID` usage.

- **Bug Fixes**
  - `posthog.test.ts`: Set labs store with `{ ...createDefaultLabsPreferences(), enabledFeatures }` instead of the invalid `{ enabledFeatures, dismissedFeatures: [] }`.
  - `purposeInference.test.ts`: Local `createTestLayout` now includes `Layout.version`/`Layout.name` and defaults `label`/`notes` to `''` to satisfy `Bin` shape.

<sup>Written for commit 59cd9ea36a83b04f67436a964370f9ed98361466. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/2994?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
One more pre-existing item worth flagging (found by the mlTelemetry sub-agent, left untouched — it produces zero type errors so was outside this PR's error budget):

`mlTelemetry/init.test.ts`'s `mockLayout` is built as `{...} as unknown as Layout` (line 48, pre-existing before this PR). The literal's shape is stale — nested `position: {x,y}` / `size: {width,depth,height}` instead of `Bin`'s flat `x,y,width,depth,height`, `category: null` instead of a `CategoryId`, and layers/drawer missing their `height`. It passes today only because `./trackers` is fully `vi.mock`ed in this file, so nothing ever reads the layout's interior — the assertions just check `toHaveBeenCalledWith(mockLayout, ...)` identity. Converting it to a real `Layout` would touch fixture values outside this PR's branding-only scope, so it's left as a follow-up for whoever next touches this file.

PR #3003, opened 2026-07-30
fix(generation): stop the compartments path leaking WASM handles (#2985)

### LINKED ISSUE #2985: WASM handle leak in the compartments generation path (delta 280 vs budget 10)
## Summary

The compartments path leaks WASM handles. Reproduced independently of the PR that surfaced it:

```
[diag] cache churn (40 unique sizes): warmed=0 afterChurn=0 delta=0
  wall cutouts         delta=0
  handles              delta=0
  scoops               delta=0
  label tabs           delta=0
  compartments 2x2     delta=280
  wall pattern         delta=0
  wall cutouts + compartments delta=280

AssertionError: compartments 2x2: delta 280 exceeds regression budget 10
```

Every other feature is **exactly** zero. `wall cutouts + compartments` is also 280, i.e. the same 280 — the leak tracks compartments alone, and cutouts add nothing. That points at the compartment/divider construction or disposal path specifically, not at a shared boolean/fuse stage.

## Reproduce

```
NODE_OPTIONS=--expose-gc pnpm exec vitest run --config vitest.profile.config.ts \
  src/features/generation/worker/generators/__kernel-tests__/disposalRegression.test.ts
```

Takes ~5 minutes.

## Why it was invisible

`disposalRegression.test.ts` called `buildParams({ gridW, gridD })` in a loop commented "Generate with varying grid sizes to force cache evictions (LRU capacity = 5)". `gridW`/`gridD` are **not** `BinParams` fields, and `buildParams()` is a plain `{...DEFAULT_BIN_PARAMS, ...overrides}` spread — so the extra keys were carried into an object `generateBin` ignores. All 20 iterations generated the identical default 2×2 bin.

The test's whole purpose is to force LRU eviction across many distinct shapes. It was hammering a single cache key, so it could not have detected a leak. Renaming to `width`/`depth` (#2983) made it do its job, and it immediately went red.

Note this test is **not** CI-gated — `__kernel-tests__/**` is excluded from the default vitest projects and the assertions are `it.skipIf(!globalThis.gc)`. So this has been latent for however long the typo has existed, and nothing would have caught it.

## Impact

Unbounded handle growth in the generation worker for any bin with compartments — one of the most commonly used features. Likely surfaces as gradual memory growth or an eventually wedged worker during a long editing session with repeated regeneration, rather than as an immediate failure.

## Status

`#2983` leaves the diagnostic correctly failing rather than re-hiding it. Since it is neither CI-gated nor part of the default test run, it blocks nothing — it now accurately reports a real defect. The leak itself is unfixed and wants its own investigation into the compartments/divider disposal path.

andymai:
## Problem

The compartments generation path leaks OCCT/WASM handles — the `disposalRegression` profiling test measured `delta 280` (budget 10), while every other feature was exactly `0`. Left unfixed it surfaces as unbounded generation-worker memory growth (and eventually a wedged worker) over a long editing session with repeated regeneration.

## Root cause

`cavityDrawing()` rounded each exterior compartment corner with brepjs's `customCorner()`, which **leaks 2 OCCT handles per call** when a later `lineTo()` consumes its deferred fillet. The compartments path is the only default generation path that calls it: 4 bin corners × 2 handles × churn iterations = the reported 280. (It is a general brepjs defect, not specific to compartments — label tabs would hit it too in a narrower-width config; compartments is just the path the test exercises.)

## Fix

brepjs's `closeWithCustomCorner()` — which rounds the corner between the last-drawn segment and the closing edge — routes through a **different, leak-free path**. Rotate the corner ring so the last rounded corner is drawn last and closed that way:

- A cavity with a **single** perimeter corner (the common case — each interior cell touches at most one bin corner) then makes **zero** `customCorner()` calls → leak goes to **0**.
- An **edge-spanning** compartment with two adjacent perimeter corners still routes the earlier one through `customCorner()` — a residual that halves rather than eliminates for that shape, and needs an upstream fix to brepjs's own `customCorner()` disposal.

The fillet geometry is unchanged (same underlying `filletCurves()`); only tessellation differs.

## Verification

- **Disposal regression:** `compartments 2x2` delta **280 → 0** (and `wall cutouts + compartments` 280 → 0).
- **Geometry integrity:** manifold/watertight (18), export-invariant + tilted-divider (8), and both-kernel scenario tests all pass.
- **Snapshots:** triangle counts shift by a constant −56/bin on the default kernel (a benign tessellation change); both `.snap` and `.brepkit.snap` updated. Typecheck + lint clean.

## Follow-ups (out of scope)

- `disposalRegression.test.ts` runs in **no** CI job today (it's under `__kernel-tests__`, `skipIf(!globalThis.gc)`), which is why this went unnoticed — worth wiring into a scheduled `--expose-gc` job.
- The residual edge-spanning-compartment leak wants an upstream brepjs fix to `customCorner()`'s disposal.

Closes #2985

---
## Summary by cubic
Fixes a WASM handle leak in the compartments generation path by reordering corner drawing and using `closeWithCustomCorner()`. Prevents unbounded generation-worker memory growth; disposal regression drops from 280 to 0.

- **Bug Fixes**
  - Cause: `brepjs` `customCorner()` leaked 2 handles per call when a later `lineTo()` consumed its deferred fillet.
  - Change: Rotate the corner ring so the last rounded corner closes via `closeWithCustomCorner()`, avoiding `customCorner()` in the common case. Geometry is unchanged; only tessellation shifts.
  - Verification: disposal regression 280 → 0 (also for wall cutouts + compartments), geometry/invariance tests pass, snapshots updated (triangle counts −56/bin on default kernel).

<sup>Written for commit cb78faa7c0d392c1e9925b1fa4217c9016c3abf2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3003?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/generation/worker/generators/__snapshots__/binGenerator.scenario.compartments.test.ts.brepkit.snap
+++ b/src/features/generation/worker/generators/__snapshots__/binGenerator.scenario.compartments.test.ts.brepkit.snap
@@ -2,8 +2,8 @@
 
 exports[`compartments > 2×2 bin with 1×1 (none) compartments 1`] = `4172`;
 
-exports[`compartments > 2×2 bin with 2×2 compartments 1`] = `4344`;
+exports[`compartments > 2×2 bin with 2×2 compartments 1`] = `4023`;
 
-exports[`compartments > 2×2 bin with 3×2 merged top-left compartments 1`] = `4030`;
+exports[`compartments > 2×2 bin with 3×2 merged top-left compartments 1`] = `3875`;
 
-exports[`compartments > 2×2 bin with 4×4 dense grid compartments 1`] = `4792`;
+exports[`compartments > 2×2 bin with 4×4 dense grid compartments 1`] = `4216`;
--- a/src/features/generation/worker/generators/__snapshots__/binGenerator.scenario.compartments.test.ts.snap
+++ b/src/features/generation/worker/generators/__snapshots__/binGenerator.scenario.compartments.test.ts.snap
@@ -2,8 +2,8 @@
 
 exports[`compartments > 2×2 bin with 1×1 (none) compartments 1`] = `2796`;
 
-exports[`compartments > 2×2 bin with 2×2 compartments 1`] = `2908`;
+exports[`compartments > 2×2 bin with 2×2 compartments 1`] = `2852`;
 
-exports[`compartments > 2×2 bin with 3×2 merged top-left compartments 1`] = `2936`;
+exports[`compartments > 2×2 bin with 3×2 merged top-left compartments 1`] = `2880`;
 
-exports[`compartments > 2×2 bin with 4×4 dense grid compartments 1`] = `3196`;
+exports[`compartments > 2×2 bin with 4×4 dense grid compartments 1`] = `3140`;
--- a/src/features/generation/worker/generators/compartmentBuilder.ts
+++ b/src/features/generation/worker/generators/compartmentBuilder.ts
@@ -82,30 +82,50 @@ function cavityDrawing(corners: CavityCorners, cornerRadius: number): Drawing {
   // Cap the radius at half the smaller span so opposing fillets can't overrun.
   const { cavW, cavD } = cavitySpan(corners);
   const r = Math.max(0, Math.min(cornerRadius, cavW / 2 - 0.05, cavD / 2 - 0.05));
-  const rBL = exterior.bl ? r : 0;
-  const rBR = exterior.br ? r : 0;
-  const rTR = exterior.tr ? r : 0;
-  const rTL = exterior.tl ? r : 0;
-
   if (r <= 0.1 || !hasExteriorCorner(exterior)) {
     return draw(bl).lineTo(br).lineTo(tr).lineTo(tl).close();
   }
 
-  // Start at the midpoint of BL→BR so close() forms a real edge through BL,
-  // letting customCorner(rBL) apply to it (mirrors buildSlabProfile in
-  // baseplateSlab.ts). Starting at a corner would make close() degenerate.
-  // Use the true midpoint (not [mid_x, bl[1]]): an override can tilt the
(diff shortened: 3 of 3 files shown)


andymai:
Closing without merging — neither viable fix for #2985 is safe to ship, so the leak is deferred to an upstream brepjs fix.

**What was tried:**

1. **App-level `closeWithCustomCorner()` (this PR's original approach) regresses #1968.** It routes through brepjs's `_customCornerLastWithFirst`, which fillets the *first↔last* curve junction — a geometrically different corner than `customCorner`'s interior fillet. CI caught it: `binGenerator.scenario.issue-1968.test.ts` fails (corner-material-missing 0.09–0.19 vs the <0.05 budget), i.e. it re-opens the thin-wall corner-gap bug. Compartment snapshots also shift −56/bin. Not a geometry-equivalent swap.

2. **Patching brepjs `saveCurve` to dispose the two fillet-consumed curves is ineffective.** Keeps geometry byte-identical (~1070 tests pass across every `customCorner` user, no snapshot changes) but `disposalRegression` `delta` stays **280** — so the leaked handles are not the two `Curve2D` inputs; they're intermediate OCCT objects created *inside* brepjs's `filletCurves`, unreachable from `saveCurve`.

**Conclusion:** the real fix belongs in brepjs's `filletCurves` disposal (upstream). The leak is gradual and non-correctness; #1968 is a correctness bug — so shipping the workaround would be a net regression. #2985 stays open.

Separately worth doing: wire `disposalRegression.test.ts` into a scheduled `--expose-gc` CI job — it runs nowhere today, which is why this went unnoticed.

PR #3021, opened 2026-07-30
feat: support the wall taper on multi-compartment bins (#3017)

### LINKED ISSUE #3017: Wall taper is unavailable on multi-compartment bins
Split out of #2933. The curved-bin wall taper/flare is gated to hollow, single-compartment bins, so a drawer-fit bin with dividers can't reach into a drawer's curved sides.

## The gate is not what the comment says

`pipeline/context.ts` strips the taper for `solid || hasMultipleCompartments(params)`, and the comment gives two reasons — a silent no-op, *or* dividers staying at nominal inner dims and protruding through the tapered base.

Measured, only the first happens. Lifting the multi-compartment strip and generating a 2×2u bin with a 12mm chamfer band, 10mm overhang per side:

| Bin | plain vol | tapered vol | delta | tris |
|---|---|---|---|---|
| 1×1 | 52897.3 | 48144.7 | 4752.6 | 5276 → 10404 |
| 2 cols | 54686.1 | 54686.1 | **0.0** | 5776 → 5776 |

Identical volume and triangle count, and the mesh stays watertight (`boundaryEdges=0`, `nonManifoldEdges=0`). `buildBinBox`'s multi-cavity branch returns before the `ov?.taper` check at `boxBuilder.ts:390`, so `buildTaperedBox` is never reached and the dividers never get the chance to conflict.

Not verified: whether the additive-divider path behaves the same. The probe used the default compartment config.

## What a fix actually needs

Not divider-span arithmetic — the multi-cavity box branch needs a lofted outer. That means composing `buildTaperedBox`'s `outer_loft − cavity_loft` with the per-compartment cavity drawings instead of the current shelled prism, and only then worrying about whether dividers sized to the rim's inner width breach a wall that narrows below them.

The ruled-loft invariant from #2948 applies: paired lofts must share z-nodes, not merely sample the same profile, and `assertWatertight` cannot see a breached wall — use the mesh Euler characteristic.

## Meanwhile

The strip should stay. It currently prevents a control that silently does nothing, which is the right behaviour until the multi-cavity path can actually taper.

andymai:
Closes #3017. Completes the curved-bin work from discussion #2933 — a drawer-fit bin with dividers can now reach into a drawer's curved sides.

## The gate was hiding a no-op, not a conflict

`pipeline/context.ts` stripped the taper for multi-compartment bins, with a comment offering two reasons: a silent no-op, *or* dividers protruding through the tapered base. Measured in #3017, only the first happens — `buildBinBox`'s multi-cavity branch returns before the `ov?.taper` check, so `buildTaperedBox` was never reached and dividers never got the chance to conflict. Resizing divider spans, the obvious fix, would have changed nothing.

## What it does now

The multi-cavity path starts from `buildTaperedLofts().outer` instead of the prismatic extrusion, and clips the compartments that can reach the wall against `.cavity`.

**The clip is load-bearing.** Below the band the tapered outer is `flare` mm narrower, so a rim-sized compartment prism doesn't merely thin the wall — it spans the entire wall thickness and the cut opens a slot straight through it. And that leaves the surface closed, so `assertWatertight` cannot see it. Both new scenario tests compare Euler characteristics against the untapered bin, the same guard #2948 established for the ruled-loft breach.

`buildTaperedBox` is now a thin wrapper over `buildTaperedLofts` — single-cavity behaviour is byte-identical, and its six existing scenario tests are untouched.

## Performance

The first working version ran a 12x12 grid in 29.7s against a 30s timeout. Two changes fixed it:

**`cutAll` instead of a cut per compartment.** The loop cut sequentially, so every cut re-traversed the whole growing solid. One multi-tool boolean instead.

**Batched clipping.** Compartments that can reach the tapered wall are fused and clipped in a single intersect rather than one each — on a 12x12 grid, 1 intersect instead of 44.

12x12 grid, 4x4u bin:

| | before | after |
|---|---|---|
| flat | 12740ms | **1906ms** |
| chamfer taper | 31477ms | **5128ms** |
| fillet taper (25mm band) | — | **10927ms** |

Most of this is not taper-specific: the **flat** path is 6.6x faster, benefiting every multi-compartment bin whether or not it tapers. Meshes are unchanged — the generators suite's triangle-count snapshots all pass untouched.

Worst case is now fillet at 10.9s against a 30s base, so the timeout bonus is headroom for larger footprints rather than a necessity.

**A third approach measured worse and is recorded in the code so it is not retried:** taking the complement twice (`cavity - compartments` = dividers, then `cavity - dividers`) needs only two booleans and no per-compartment work, but ran 19.2s — the intermediate divider solid is far more complex than the compartments it came from.

## Scope

Solid bins keep the strip: no cavity for the inner loft to cut, so the control would still silently no-op. The UI gate relaxes from "hollow, single-compartment" to just "hollow", and the hint key is renamed to match across all 14 locales.

## Verification

- `binGenerator.scenario.taper` — 9 passed, 2 new (chamfer 2-col, fillet 2x2), both asserting closed + manifold + unchanged Euler characteristic
- generators 2438, unit + dom 15342 — all three projects run in full, not scoped
- typecheck, lint (0 errors), four i18n gates, module boundaries

Review fixes in this PR: the timeout gate now resolves through `resolveOverhang` (the same helper the worker uses) so it cannot drift on legacy `enabled`-absent configs or polygon masks; the hollow-only hint no longer names the chamfer profile in es/nl/pt-BR; and a redundant `scope.register` on the multi-cavity base is gone.


andymai:
### Performance addressed

The original numbers were not shippable — a 12x12 grid ran 29.7s against a 30s timeout. Two changes:

**`cutAll` instead of a cut per compartment.** The loop cut each compartment sequentially, so every cut re-traversed the whole growing solid. One multi-tool boolean instead.

**Batched clipping.** Compartments that can reach the tapered wall are fused and clipped in a single intersect rather than one each — on a 12x12 grid, 1 intersect instead of 44.

12x12 grid, 4x4u bin:

| | before | after |
|---|---|---|
| flat | 12740ms | **1906ms** |
| chamfer taper | 31477ms | **5128ms** |
| fillet taper (25mm band) | — | **10927ms** |

The flat path is 6.6x faster as a side effect, which benefits every multi-compartment bin whether or not it uses a taper. Meshes are unchanged — the generators suite's triangle-count snapshots all pass untouched.

Worst case is now fillet at 10.9s against a 30s base, so the timeout bonus is retuned down (15s → 10s flat, 250ms → 100ms per compartment) and is headroom rather than a necessity.

**A third approach measured worse and is recorded in the code so it is not retried:** taking the complement twice (`cavity - compartments` = dividers, then `cavity - dividers`) needs only two booleans and no per-compartment work, but ran 19.2s — the intermediate divider solid is far more complex than the compartments it came from.

PR #3083, opened 2026-08-01
feat(drawer-shape): import a drawer perimeter from SVG or DXF

### LINKED ISSUE #3054: Baseplate & Layout Editor: Create Custom Perimeter that grid is fitted to
**Problem**
Custom drawer shape solves half of the issue. Internally removed grid elements have no finished edge 

**Proposed Solution**
Allow user to draw or import a custom outline. Editor should have dimensions, allow edge fillets, and be editable after created to fine tune..

only whole grid elements will be fitted inside the perimeter.. padding will fill the rest. 

**Alternatives Considered**
Use a different tool like: https://www.extrabold.tools/gridfinity-baseplate which allows this.

andymai:
Drawing a perimeter by hand is fine for a shape you are inventing and useless for one you have already measured. This reads the file the measurement came in: SVG or ASCII DXF, at true scale, straight into the pen editor.

Closes #3054. Stacked on #3082 — review that first; this PR's diff is only the import work.

## DXF is the faithful format here, not the awkward one

Group code 42 on a polyline vertex **is** `OutlineVertex.bulge` — the same `tan(sweep/4)` convention the type's own docstring cites. So an arc drawn in CAD arrives as an arc:

```
DXF LWPOLYLINE + bulge -> M 20 18 L 400 18 L 400 158 A 98.995 98.995 0 0 0 260 158 L 260 318 L 20 318 Z
```

That is a real 99mm arc, not a polyline approximating one. SVG gets the same treatment for circular `A` commands (splitting sweeps over 180° so `|bulge| <= 1` holds); only beziers are flattened, at the 0.05mm tolerance the rest of the outline pipeline already uses.

## Loose entities are chained

2D CAD exports a profile as separate `LINE` and `ARC` entities far more often than as one closed polyline. Without chaining, an ordinary drawer imports as nothing at all. The same fixture written both ways now produces a byte-identical path.

An open chain is dropped rather than force-closed — silently bridging a gap invents an edge the user never drew.

## Nothing is silently rescaled

True scale is the entire point of importing from CAD, so a perimeter larger than the drawer asks instead of guessing:

> The imported shape measures 900 × 600 mm, which is larger than this drawer.
> [Grow drawer to 21.5 × 14.5 u] [Scale to fit] [Cancel]

Growing fits the loop against the drawer it is **about** to have rather than re-reading the store after the resize, so the outline and the resize land in one commit instead of racing. Growing is only offered when the drawer could actually get big enough. Loops dropped (only the largest becomes the perimeter) and points thinned to reach `OUTLINE_MAX_VERTICES` are always toasted, never silent.

The prompt is inline rather than a nested modal: the pen dialog already owns a focus trap, and keeping the shape on screen is what makes the choice answerable.

## Two things worth flagging for review

**`svgLength` and `svgTransform` moved to `src/shared/utils/svg/`.** The cutout editor and the drawer perimeter read the same files and must agree on units, transforms and where the origin sits. A cross-feature import is a hard `check-module-boundaries.sh` violation and a second copy would have been worse. The move is mechanical — no logic changes — so bin-designer's existing tests are the regression net, and both helpers gained the direct tests they never had.

**The parsers load through a dynamic `import()`.** `DrawerShapeSection` is pulled in eagerly by both the sidebar and the mobile settings sheet, so a static import would put an SVG path parser and a DXF reader in the eager bundle for every visitor who never imports a file.

## Verification

81 tests across the parsers, the chainer, fitting and simplification, plus 12 on the import hook covering the whole oversize decision. Driven end to end in the app against real fixtures: a DXF polyline with a concave arc bite, the same profile as loose LINE/ARC entities, an SVG in physical millimetres with a circular arc corner and an inner circle that correctly loses to the perimeter, and an oversized file taken down the grow path (drawer became 21.5 × 14.5 u and the outline applied).

Full coverage run green (statements 80.84%, branches 74.76%, functions 78.81%, lines 82.20% — all four up slightly on the base branch). Typecheck, all four i18n checks (13 new keys across 15 locales), module-boundary, design-system, exhaustiveness and missing-tests gates all pass.

**Note on CI:** `ci.yml` triggers only on `pull_request: branches: [main]`, so a PR stacked on another branch gets no Build, Unit Tests or Quality run. Those jobs will run once this is retargeted to `main` after #3082 merges.

---
## Summary by cubic
Import a drawer perimeter from SVG or ASCII DXF at true scale into the pen editor, with preserved arcs, robust chaining of loose CAD entities, and an inline oversize prompt. Closes #3054.

- **New Features**
  - Import SVG/DXF into the pen editor; preserves circular arcs (DXF bulge code 42, SVG `A` commands). Only beziers are flattened.
  - Chains loose `LINE`/`ARC` segments into closed loops; drops open chains.
  - True-scale placement with an inline prompt when larger than the drawer: grow to the required units, scale to fit, or cancel.
  - Picks the largest closed loop, normalizes winding, and simplifies only when exceeding `OUTLINE_MAX_VERTICES`.

- **Refactors**
  - Moved `svgLength` and transform helpers to `src/shared/utils/svg` so all importers share units, transforms, and viewBox handling.
  - Load parsers via dynamic `import()` to keep the eager bundle small.

<sup>Written for commit 35ed46ae4948576779cd81177106e4c211a64330. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3083?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Superseded by #3085. GitHub auto-closed this when its stacked base branch was deleted on the #3082 merge, and would not let it reopen after the rebase onto `main`. Same branch, same commits, now targeting `main` so full CI runs.

Issue #3169, opened 2026-08-03
Custom Perimeter Grid Shift Appears to Truncate the Perimeter In Layout and Baseplate after Shifting


gismofx:
**Describe the bug**
When centering grid within custom perimeter, the perimeter appears to be truncated from the model in both X and Y.

**To Reproduce**
Steps to reproduce:

1. Refer to #3149 for Layout and custom perimeter points
2. Apply Grid shift of 4.5mm

<img width="753" height="610" alt="Image" src="https://github.com/user-attachments/assets/12a18d77-e8df-4fd1-b359-478ff3ec8f88" />

3. Observe that grid perimeter shifted. Also, it appears to be cut off (rendering issue that should be fixed)
4. Switch to baseplate, ensure sync with layout is checked
5. Observe that left side of baseplate is truncated similar to how layout view looks but in 3d

<img width="969" height="698" alt="Image" src="https://github.com/user-attachments/assets/61f0aaa0-39ec-4a14-b57a-2773f7d91ef4" />

6. You can repeat this with a shift in Y as well and perimeter will be truncated. See below, you can see the bottom edge is truncated. Obvious as the radius is no longer tangent to bottom edge

<img width="667" height="361" alt="Image" src="https://github.com/user-attachments/assets/60c19b67-c751-4d18-8ba7-c59b83c95818" />

**Expected behavior**
1. Layout should render entire custom drawn perimeter inclusive of whatever grid shifts
2. Baseplate should never be truncated, entire perimeter should always be generated. Regardless of X or Y shift values.

**Screenshots**
See above.

**Environment:**

- Browser: Chrome 150
- Device: Desktop

PR #3227, opened 2026-08-04
feat(baseplate): let users draw custom split lines

### LINKED ISSUE #3115: Baseplate - Custom Split Lines to subdivide a large baseplate.
**Problem**
The automated split might not be ideal for me. I want to define where the grid plates are split. There are also some other methods to split like "T-Junctions" which eliminate 4-way intersections which can be weaker.

**Proposed Solution**
Create a simple editor in baseplate mode that lets me draw horizonal and vertical lines along grid where the grid plates will be split
Would also be neat if the tool warns the user that any custom sub-division violates the build plate size that is defined.

**Alternatives Considered**
Use a different grid baseplate generator

andymai:
Closes #3115.

## What

The split planner's search is the only way to decide where a large baseplate is cut, so a user who wants a seam somewhere else has no recourse. This adds an optional user-drawn plan that replaces the search while leaving everything downstream unchanged.

`splitOverride` (chunk sizes per axis, in grid units) is stored per layout and edited by clicking the seam ticks in the panel's mini-map, which is now proportional to the real plate. The plan is seeded from whatever the planner last produced, so hand-editing starts from the automatic answer rather than a blank plate. One tick is one command, so one undo step. A "Reset to automatic" link clears it.

## Why it drops into the existing pipeline

Three things keep a hand-drawn plan compatible with the planner's invariants:

- it is **not** run through `reorderForDisplay`, which prettifies search output (largest-first, fractional pinning, palindromic) and would slide hand-placed seams elsewhere
- the **padding-reduction hint is suppressed**, being advice about the automatic plan
- cuts may only land on **cell boundaries**, which keeps a fractional plate's half unit on its `fractionalEdge` — the invariant a piece's `fractionalEdgeX/Y` tag already assumes

Margins, outline shaping, fingerprint dedup and stack grouping recompute from the resulting pieces exactly as for an automatic plan, so shaped plates, detached margins and stack printing compose untouched.

## Normalization

A plan is meaningful only while its sums still match the plate, so `buildFullParams` drops an orphaned one (grid resize, fractional-edge flip, malformed synced payload) and the plate falls back to the automatic tiling — it is never rendered from a plan that disagrees with the geometry. `normalizeSplitOverride` validates shape as well as sums, since the key is allowlisted server-side without a type check.

It is deliberately **absent from `meshCacheKey`**: it selects which pieces exist, and each piece's own params already carry everything its mesh depends on. It **is** a `selectGenerationTriggers` input, since the piece set changes.

## Bed fit

`tiling.bedOverages` lists pieces exceeding the bed using `axisChunkMm` — the same budget the planner's own feasibility check runs, including exterior padding, `outlineOverhang` (#3169/#3212) and male tongue protrusion (#1498). Like that check it does not credit a 90° rotation, so a custom plan can't pass a test the automatic one would fail. Offending pieces flag red and `canExport` goes false; the preview keeps rendering so mid-edit feedback still works.

## UI note

Seam ticks sit in rulers flanking the map rather than as lanes drawn across it. Two crossing sets of full-span hit targets contend at every intersection — whichever renders last wins — which left half of one axis's controls unclickable. Caught by the e2e spec against the first implementation. The map itself carries a `pointer-events-none` overlay showing where the seams land.

## Out of scope

Automatic staggering of seams to avoid four-way junctions (raised by @typytypytypy in the issue, citing GRIPS) is **not** included. It breaks the planner's separable per-axis model — a piece's edge could face two neighbours — and touches seam connector placement, `splitReorder` and `pieceFingerprint` dedup. Worth filing separately. Both commenters asked for automatic-first with manual adjustment on top, which is what this delivers for the manual half.

## Testing

- `splitOverride.test.ts` (24) — seam/chunk conversion, fractional-edge placement, validity gate incl. hostile payloads
- `splitPlanner.test.ts` (+11) — override tiling, order preservation, hint suppression, per-axis bed overage incl. padding/tongue budgets
- `SplitViewStrip.test.tsx` (19) — tick rendering, add/remove, depth-axis orientation, custom caption, reset, over-bed labelling
- `buildFullParams.test.ts` (+3), `useBaseplateExport.test.ts` (+2), `api/lib/baseplateValidation.test.ts` (new, 9)
- `e2e/baseplate-custom-split.spec.ts` (4) — drives the real app
- Full suite: 20,363 passed. Typecheck, lint, all 4 i18n gates, boundaries and design-system checks clean. 7 new strings translated across all 14 locales; `baseplate.splitInfo` split into `.one`/`.other` since a custom plan can now merge down to a single piece.

Verified in the running app: a hand-drawn 3x4 plan on a 504x420mm plate generates 12 real pieces, and merging every seam produces the over-bed warning with export disabled.

---
## Summary by cubic
Adds user‑drawn split lines for baseplates so you can place seams exactly where you want. Export now blocks over‑bed pieces across all paths, including whole‑layout ZIP, with clear errors. (#3115)

- **New Features**
  - Draw seams via ticks on side rulers in a proportional mini‑map; plan is saved per layout as `splitOverride`, seeded from the last automatic plan, with “Reset to automatic.”
  - Planner tiles from the custom plan; margins, outline shaping, fingerprint dedup, and stack grouping behave as before.
  - Over‑bed pieces are labeled and still preview; export is disabled when `tiling.bedOverages` has entries.

- **Implementation Notes**
  - `normalizeSplitOverride` validates shape/sums and drops orphaned plans after grid or fractional‑edge changes; manual plans skip `reorderForDisplay`, suppress the padding‑reduction hint, and constrain cuts to cell boundaries respecting the active fractional edge.
  - `splitOverride` is excluded from `meshCacheKey` and included in `selectGenerationTriggers`.
  - Validation allowlists `splitOverride` and bounds chunk lengths/values (API caps chunks per axis to 100; command/schema add matching bounds).
  - Export builder enforces the bed‑fit gate and throws `BaseplateBedOverageError`; whole‑layout export degrades to bins‑only with a targeted message.
  - Mini‑map hover uses a label `Set` for O(1) piece lookups on large grids.

<sup>Written for commit 7aabf2ab8b51b6f1c7f59a66cdd44f777eebd213. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed all review feedback in 4bdb8db.

**Greptile — whole-layout export bypass.** Confirmed and fixed. `canExport` only greys out the baseplate page's own button; the whole-layout ZIP calls `buildBaseplateExportPieces` directly and would have shipped the same oversized pieces. The enforcing gate now lives in that builder, which is the single choke point every export path runs through. It throws `BaseplateBedOverageError` carrying the offending labels. The whole-layout path already degraded a baseplate failure to a bins-only archive, so this composes with existing behaviour — but it now names the pieces (`layoutExport.baseplateOverBed`) instead of the generic "the baseplate could not be generated", since an over-bed split is a fixable user mistake and the old message gave nothing to act on. `canExport` stays as the UI affordance.

**Copilot — unbounded `splitOverride` in the command payload.** Fixed. The `defineCommand` schema now uses the same bounded `splitChunkSizes` as `COMMAND_SCHEMAS` (length ≤ `GRID_MAX * 2`, values 0.5–`GRID_MAX`). The point about `selectGenerationTriggers` is the sharp one: an unbounded array would ask the worker for a piece per entry.

**Copilot — O(cols\*rows\*pieces) label lookup.** Fixed. Precomputed `Set` of piece labels, `.has()` per cell.

Also fixed a test-only issue this surfaced: `useLayoutExport.test.ts` mocked `@/features/baseplate` with a bare factory, so the new named import resolved to `undefined` and the `instanceof` check threw. It now spreads the real module.

New coverage: 4 tests on the builder gate (throws, carries labels, fits-normally, automatic-plan-untouched) and 1 on the whole-layout toast asserting the interpolated text. Full suite 20,368 passing; typecheck, lint, i18n gates and e2e green locally.

PR #3243, opened 2026-08-05
feat(community): graduate the showcase to on-by-default, marked experimental

andymai:
## What

`community_showcase` becomes `preview` with `defaultEnabled`, so the Community tool and the Publish action are there for everyone, and Labs keeps the switch that turns them off.

## `defaultEnabled` is new

It had to be threaded through **every read and every write**. Resolving it in `isFeatureEnabled` alone would leave `toggleFeature` flipping a default-on flag to *on*, and `disableFeature` reading "never stored" as "already off" — the one control that turns the feature off would silently do nothing. All four paths now go through `resolveEnabled()`.

`getEnabledCount` derives from the same resolution, so the Labs badge counts what the app actually runs rather than what was clicked.

## What "experimental" looks like

Different surfaces have different room:

| Surface | Marker |
| --- | --- |
| Publish button (designer header) | `Badge` beside the label at 2xl, dot when it collapses to an icon |
| Publish dialog | `Badge` beside the title, present at the moment of the irreversible action |
| Community nav segment | one dot, muted |

All three spell the word out in the accessible name, so the marker is never the only carrier.

## One dot, two signals

The Community segment has room for one marker, and **news wins the slot**. "There is something new in here" is actionable; "this is experimental" is a standing property already spelled out on the page the segment leads to. The experimental dot is muted, the news dot is accent, and both stay in the accessible name.

## Post-export nudge

The export dialog offers to publish **once, ever**. An export is the moment a design is finished enough to be worth sharing, which is why the offer rides it rather than sitting in the UI permanently.

It skips a design that is already published, and stands down entirely when `recordExportAndShouldPromptSupport()` claims the success view: two asks on one action is one ask too many. A browser with no usable storage reads as already-offered, so a lost dismissal cannot become a prompt on every export.

## Follow-up

This is 3 of 3, after #3239 and #3241.

## Test plan

- `src/core/store/labs.test.ts` — default-on counts as enabled, toggles off on the first press, disable/enable round-trips, count agrees with `isFeatureEnabled`
- `src/shared/components/ToolSwitcher` — news dot takes the slot from the experimental dot, both named in the label, only the community segment is marked
- `src/features/bin-designer/utils/publishNudge.test.ts` — one-shot, and fails closed without storage
- `src/features/bin-designer/components/ExportDialog` — offer fires after a successful export, opens the dialog from the toast action, fires once, skips a published design, yields to the support prompt
- `src/features/bin-designer/components/DesignerPage` — publish button names itself experimental on desktop and mobile
- Full `src/core`, `src/shared`, `src/features/community`, `src/features/bin-designer`: 10618 passing; typecheck and lint clean

Not verified in a browser: no Chrome extension was connected to this session.

---
## Summary by cubic
Graduates the Community Showcase to a preview feature that’s on by default but still switchable in Labs, and gates the “fits gap” gesture behind its own opt-in flag to avoid changing grid interactions by default. Adds a one-time publish nudge after export and fixes UI and flag resolution issues found in a browser pass.

- **New Features**
  - Community tool and Publish action are visible by default (`community_showcase` → preview with `defaultEnabled: true`). The UI marks it “Experimental” on the Publish button (badge at 2xl, dot on icon) and in the Publish dialog header.
  - After a successful export, show a one-time toast prompting “Publish it” (10s). Opens the Publish dialog, skips if the design is already published, and yields if the support prompt is active.

- **Bug Fixes**
  - Centralized feature flag resolution in `resolveFeatureEnabled`; `useFeatureFlag` now selects through it. Write paths honor `defaultEnabled`, and `getEnabledCount` counts active experimental/preview features by resolved state.
  - Split the “Find bins that fit” gesture/button into a separate flag, `community_fits_gap` (opt-in). `community_showcase` no longer affects the grid’s right-button behavior.
  - Publish preview is square and capped on mobile; stacked split fixes the rail border; footer stacks with a full-width primary on small screens.
  - Designer header collapses ToolSwitcher labels below 2xl; the experimental dot is muted; overflow menu aligns to the trigger’s end and uses “Star on GitHub” for clarity.

<sup>Written for commit 7968dbe7990ad406f01818e12af8eadb5ff41722. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
## Browser pass (Playwright, dev server, 390–1600px)

Driving the real app turned up six defects the unit tests could not see. All fixed in `48ba21d`.

### The flag flip reached nothing

`useFeatureFlag` was a **second copy** of the resolution rules reading `enabledFeatures[id] ?? false` directly, so `defaultEnabled` was honoured by the store and ignored by every component that gates on it. `/community` fell through to the layout planner and the Community segment never rendered, while `isFeatureEnabled` happily returned `true`.

There is now one resolver, `resolveFeatureEnabled`, and the hook selects through it. Regression test added at the hook level, which is where the coverage gap was.

### Publish preview pillarboxed the design

Captures are 384×384; the frame was `aspect-[4/3]` with `object-contain`, so every render sat between grey bars. Square now, and capped below `md` where the split stacks — a full-width square was the entire first mobile screen, putting the form below the fold.

### Stacked split kept the rail's side border and its fixed width

`Dialog.Sidebar` carries `border-r` and `w-64`. Once `Dialog.Split` stacks, that renders as a stray vertical line two thirds of the way across a phone screen.

### Mobile footer squeezed the primary into three lines

The licence paragraph took the leading slot and crushed "Sign in & publish". The footer stacks below `md` now, button full width.

### Designer header ran out of room

A fourth switcher segment cut the design name to `Untitl…` at 1280. The switcher drops its labels below 2xl on that surface, keeping the icons — it is persistent global nav, and the name of the thing being edited is worth more than its labels.

### Overflow menu grew rightward from its trigger

It landed over the inspector panel, and the GitHub row said only "Star" once the icon that carried the context was gone. `Menu.Root` takes `align` now; the row names its destination.

Also: the experimental dot is muted everywhere. In accent it sat beside the save indicator and read as unsaved work.

### Verified after

11203 tests, typecheck and lint clean. Screenshots at 390 / 900 / 1024 / 1280 / 1440 / 1600 across the planner, designer, community route and publish dialog.

andymai:
## Fits-gap gets its own flag

Turning `community_showcase` on by default turned on more than the showcase. Nine grid-editor tests failed, and they were failing **correctly**: they assert that a right-button drag on the layout grid does nothing and that "Find bins that fit" is absent, and neither was true any more.

Concretely, the graduation was about to change a core editor interaction for everyone: right-drag on the grid would select a gap and suppress the browser context menu, and a new button would appear in the grid toolbar. That is a layout-editor change, not a consequence of the gallery being available to browse.

`community_fits_gap` is now its own opt-in flag. `community_showcase` keeps the gallery, the publish action and the Community tool, and no longer reaches into the grid.

## One refactor reverted

I had moved the flag resolvers next to `getFeature` in `core/labs/features.ts`, which reads better but broke ten further tests: inside the registry module that `getFeature` call is module-internal, so `vi.mock` on the module's own export no longer intercepts it and every test that mocks the registry silently stopped reaching it. They are back in the store, calling the imported `getFeature`, with a comment recording why they live there.

## Verified

Typecheck clean, lint clean, full suite 20583 passing.

The one red job was `baseplateGenerator.scenario.overtile` timing out at 30s on the self-hosted runner. This branch touches nothing under `src/features/generation` or `src/features/baseplate`, and the file passes locally in full (5/5), so it is runner-side. Re-run.

PR #3244, opened 2026-08-06
fix(designer): give the wall-less tray a floor and a seatable foot

andymai:
The wall-less tray (#3234) had no floor and a foot that could not seat in a baseplate. Both trace to one premise in the original design note: *"the base profile flares as it rises, so the feet already merge into a continuous surface that serves as the floor."*

## The floor

`buildBaseSocket` sizes each foot `cell.widthUnits * unitX - CLEARANCE` (41.5mm) on a 42mm pitch and rounds its top, so adjacent feet stop **0.5mm apart and never touch**. An ordinary bin hides that under the `wallThickness` slab `shell()` leaves below its cavity — the tray built no box, so it had no slab.

Vertical rays through a 2×2 tray, export mesh:

| column | tray (before) | plain bin |
| --- | --- | --- |
| `(0, 0)` four-foot junction | `EMPTY` | `[5.00, 6.20]` |
| `(0, 21)` internal grid line | `EMPTY` | `[5.00, 6.20]` |
| `(21, 0)` perpendicular | `EMPTY` | `[5.00, 6.20]` |

Four disconnected islands joined only by the perimeter lip, with a through-slot along every internal grid line.

## The foot

`buildTopShapeLoft` extends the lip `LIP_TAPER_WIDTH` **below its own base plane** (`Z_ANGLE_BOTTOM`) for the 45° support that blends it into the wall. Dropped onto a zero-height wall, that support landed inside the Gridfinity taper and back-filled it to full width:

| z (foot spans 0–5) | tray (before) | plain bin |
| --- | --- | --- |
| 2.60 | 20.750 | 18.600 |
| 3.50 | 20.750 | 19.500 |
| 4.50 | 20.750 | 20.500 |

A straight 41.5mm prism for the foot's top 2.7mm — it would land ~2.4mm proud of a baseplate pocket.

## The lip toggle

An `IMPLICATION_RULE` forced `stackingLip: true` whenever `tile` was set, because the lip *was* the entire shell. Toggling it in the panel built the identical solid.

## Fix

`shellStage` builds the tray body as a `wallThickness` solid slab over the full footprint — the same slab an ordinary bin gets. With the lip on it takes `buildTopShape`'s support-free branch and seats on the slab, so the ring's underside is a flat annulus bearing on solid material (supported to print) and the foot profile is untouched.

That makes the lip a genuine choice, so the forcing rule goes.

| | before | after |
| --- | --- | --- |
| with lip | 9.30mm | **10.50mm** |
| without lip | 9.30mm (forced) | **6.20mm** |

It was only ever 9.3mm *because* it had no floor. `assembledHeight` (drawer clearance), `printEstimates` (it priced a `height`-tall wall out of a `height` that is inert on a tray), `splitBinBuilder`, and the hint copy in all 15 locales follow.

## Why the old suite passed a floorless tray

Every assertion was bounding-box Z, triangle-count inequality, or watertightness — and a ring of feet joined by a lip satisfies all three. "Watertight" answers *is the surface closed*, not *is the interior where I think it is*.

`meshAssertions` gains `verticalSolidSpans` / `isSolidThrough` (and `sectionHalfWidth`), so the scenarios now assert the floor at internal grid lines and the foot against a real plain bin. Re-run against the old `shellStage`, **7 of the 12 tray scenarios fail** with e.g. `no material between z=5 and z=6.2 at (0, 0)` and `tray foot deviates from a plain bin's at z=2.85mm: 20.750mm vs 18.850mm`.

Also adds a ring-footprint case: the support-free lip branch was dead code in production until now, and a mask with a hole is the one shape where it has a second loop to get right.

## Verification

- 17,984 unit + dom tests pass
- 261 files / 2,576 generator tests (real WASM) pass, exit 0 — including the triangleCount snapshots, so nothing moved outside the tray path
- `typecheck` and `check:i18n` clean; lint byte-identical to `main` (3 pre-existing warnings, 0 new)

Verified numerically (ray casts and plane sections against a plain bin), not visually in the app.

## Notes

- **"Wall thickness" now drives the tray's floor slab** and stays editable for a tray. The control is meaningful; the label reads oddly on a wall-less bin. Renaming is another 15-locale change, so it is left alone.
- **Pre-existing, untouched:** `boxWallHeight` goes *negative* for a 1u bin at `HEIGHT_UNIT_MM_MIN: 3` (`3 - SOCKET_HEIGHT = -2`) and reaches `extrude(-2)` unguarded — true before and after this change. Worth its own issue.

---
## Summary by cubic
Adds a real `wallThickness` floor slab to wall-less trays and preserves the Gridfinity foot, so trays print cleanly, seat in baseplates, and preview correctly. The stacking lip is now optional, and UI, height/readouts, print estimates, and the draft preview all reflect the slab.

- **Bug Fixes**
  - Tray geometry: feet + solid slab (+ optional lip). The lip uses the support-free branch and fuses to the slab; the foot matches a plain bin and seats in baseplates.
  - Correct readouts and costs: `assembledHeight` = `SOCKET_HEIGHT + wallThickness` (+ lip). With lip: 10.5mm; without lip: 6.2mm. Print estimates ignore inert `height`, drop wall cost, and respect the lip toggle; hint text updated in all locales. Removed the rule that forced `stackingLip`.
  - Editor preview: keep trays off the direct-mesh draft path to prevent flashing 7mm walls from an inert `height`.
  - Use resolved slab thickness end-to-end: pass `resolveTrayFloorThickness` into `buildBinBox` (as both wall and slab thickness) to avoid 0/NaN crafting corrupting cache keys or tapered outers.

- **Refactors**
  - `shellStage` builds the slab for trays and fuses the optional lip; `deriveDimensions` adds `trayFloorHeight`; `splitBinBuilder` matches the support-free lip; direct-mesh gate rejects trays.
  - Mesh assertions add `verticalSolidSpans` / `isSolidThrough` / `sectionHalfWidth`; scenarios include a ring footprint; probes move off axes.
  - New shared resolver `resolveTrayFloorThickness` keeps slab thickness consistent across generation and `assembledHeight`, clamping 0/NaN.

<sup>Written for commit dc033553e7fe10b72bc89f56a9b57995c1e087e9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3244?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed all three review findings.

**Greptile P2 — direct preview ignores tray mode.** Confirmed and fixed. `canBinUseDirectMesh` rejects `base.spacer` with the exact rationale ("the procedural draft models a solid floor, so drafting it would show a bin that isn't there") but had no gate for `base.tile` — the opposite miss. The draft extrudes a full-height hollow body from `params.height`, which is inert on a tray, so editing one flashed a walled 7mm bin over a plate whose body is a `wallThickness` slab. This is very likely part of what was originally reported as wrong geometry, since it is what you see while editing. The fallback table now covers both cases; the spacer had the gate but no test row.

**Copilot — `assembledHeight` didn't clamp `wallThickness`.** Valid, and my inconsistency. Rather than duplicating the guard I gave it one home: `resolveTrayFloorThickness` in `types/base.ts`, re-exported through the shared-types barrel and used by both `deriveDimensions` and `assembledHeight`. A crafted 0/NaN thickness would otherwise have had the readout describe a plate the generator never built, or report NaN.

**Copilot — probe points on the axes.** Taken. `verticalSolidSpans` collapses coincident hits precisely because a planar face is fan-triangulated from the footprint centre, and the tests passed on that collapse — but a probe that leans on it is weaker than one sampling clear of the shared edges. Columns moved off-axis while staying inside the 0.5mm inter-foot gap.

The two generator-shard failures on the previous run were both 30s timeouts (`baseplateGenerator.scenario.overtile` and `labelPlateBuilder`), neither on the tray path and both passing locally — a sibling in the same shard logged `tsumiishi-kikko carves a 1×1×6 bin 24247ms` against a 30s limit. This push supersedes that re-run.

PR #3255, opened 2026-08-06
fix(designer): keep a spacer's body above the socket it stands on

andymai:
Follow-up to #3244, which surfaced this while checking a branch predicate.

`wallHeight` is `height * heightUnitMm - SOCKET_HEIGHT`. `MIN_SPACER_HEIGHT` (1u) was written for the 7mm default, but Physical Units allows a height unit down to `HEIGHT_UNIT_MM_MIN: 3` — where a 1u spacer asks for a **−2mm wall**.

## Neither degenerate case fails loudly

Probed against the real generator, 1×1 spacer at 1u:

| `heightUnitMm` | `wallHeight` | before                                   |
| -------------- | ------------ | ---------------------------------------- |
| 3              | **−2.00**    | no error — 7.30mm mesh, silently mangled |
| 4              | **−1.00**    | no error — 8.30mm mesh, silently mangled |
| 5              | **0.00**     | throws `EXTRUDE_ZERO_VECTOR`             |
| 7              | 2.00         | ok — 11.30mm                             |

At exactly `SOCKET_HEIGHT` the box extrude is a zero vector and generation dies. **Below it, OCCT accepts the negative vector and extrudes downward** — the box is built inside the socket and the stacking lip fuses 2.1mm below the socket top, swallowing the foot taper. That case ships a broken bin instead of failing, and `assembledHeight` agrees with the bogus mesh (both say 7.3mm), so nothing surfaces it.

It's the same taper-swallowing failure #3244 fixed on the tray, reached by a different route.

Only a spacer can get here. Every other socketed base is held at `MIN_HEIGHT` (2u), which clears the socket at any permitted height unit (`2 × 3 − 5 = 1mm`).

## Fix

**Root cause.** `minHeightUnits` derives the spacer's floor from the height unit: `ceil((SOCKET_HEIGHT + MIN_BODY_WALL_MM) / unit)`. It bites only below a 6mm unit — **at the 7mm default the floor is still 1u, so no stored design moves**. Mirrored in `api/lib/designerValidation.ts`.

**Backstop.** `deriveDimensions` floors the socketed wall at `MIN_BODY_WALL_MM`. This carries more weight than it looks: `heightUnitMm` is allowlisted server-side but **never range-checked**, so the generator is the only thing between a crafted payload and an inverted box. It degrades to the shortest real wall rather than losing the bin.

After: every unit produces a valid solid, and 7mm is untouched.

| `heightUnitMm` | `wallHeight` | after                    |
| -------------- | ------------ | ------------------------ |
| 3 / 4 / 5      | 1.00         | ok — 10.30mm, watertight |
| 7              | 2.00         | ok — 11.30mm (unchanged) |

## The tray is deliberately excluded

A tray's `height` is **pinned** to 1 by `useBaseSection.commit` for fingerprint stability, on a path that returns early without consulting `minHeightUnits`. Raising the shared `relaxed` floor would have made the validator reject every tray at `heightUnitMm ≤ 5` — breaking what #3244 just fixed. A tray's wall is pinned to 0, so it never subtracts a socket and cannot reach this bug.

I verified that carve-out is load-bearing rather than incidental by inverting it: the tray test fails without it.

## Tests

- `minHeightUnits` across the whole reachable unit range (3–20), asserting the derived floor always leaves ≥ `MIN_BODY_WALL_MM` of wall, plus the nonsense-unit fallback and the tray carve-out.
- Server-mirror tests: 1u spacer rejected at 3/4/5mm units, accepted at 2u, still accepted at the default, tray unaffected.
- Two generator scenarios at 3mm and 5mm units. Both **fail against the pre-fix `deriveDimensions`** and pass after.

## Unrelated but blocking review

`api/lib/designerValidation.test.ts` carried a **literal NUL byte** in a pre-existing `sanitizeTags` fixture, which made the whole 1,600-line file binary to git — every change to it landed as an unreviewable `Bin 60855 -> 62961 bytes` diff, including the tests added here. Rewritten as a source escape (the six ASCII characters `\\u0000` inside the string literal): identical runtime value, file is text again, and the control-char test still passes. (The diff on that file still shows as binary in this PR because the _old_ blob is binary; it will be text from here on.)

## Verification

- 18,035 unit + dom tests pass
- typecheck clean; lint identical to `main` (1 pre-existing `max-lines` warning on `designerValidation.ts`, which was already 500+)

---
## Summary by cubic
Prevents spacers from generating inverted or zero-wall bodies when using small height units by raising the spacer’s minimum height and clamping the generator’s wall extrude. Trays are unaffected.

- **Bug Fixes**
  - Spacer floor now tracks `heightUnitMm` to ensure `wallHeight = height * heightUnitMm - SOCKET_HEIGHT` stays ≥ `MIN_BODY_WALL_MM`.
  - Generator floors socketed walls to `MIN_BODY_WALL_MM` to guard against crafted payloads.
  - Tray explicitly excluded: height pinned to 1 and wall is 0, so no socket subtraction or floor change.
  - UI and server validation mirror the new floor via `minHeightUnits(base, heightUnitMm)`.
  - Replaced a literal NUL in `sanitizeTags` test fixture with `\\u0000` to restore text diffs.

<sup>Written for commit a942cbba43dd23e53149cba0843aa53a5a5948a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3255?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Copilot's note on the hard-coded `5` in the height-floor test is valid — it duplicates `GRIDFINITY.SOCKET_HEIGHT` and would drift if the spec constant ever moved. It is test-readability only and the literal is correct today, so I'm merging on green and folding the constant reference into the follow-up rename PR rather than spending another full CI cycle on it.

Issue #3036, opened 2026-07-31
Cut Editor for Lid/Tray Top


typytypytypy:
**Problem**
I made a bin with a lid/tray top, the insides are segmented beautifully(Thanks to your tool), but when I shut the cabinet drawer the tray contents slide.

**Proposed Solution**
Implement the same cut editor for the lid/tray top.

andymai:
Thanks, this is a good ask and the use case is clear: a lid that holds its contents when the drawer moves.

Marking this as accepted in principle. Some notes on what it will take, so expectations are set.

The wall-cutout editor and the lid are currently separate systems. Cutouts are authored against the bin's interior compartment grid, and the lid is generated as its own plate with an attachment style (friction, click rails, magnetic). So this is not a matter of pointing the existing editor at a new surface; the lid needs its own notion of "here is a region, cut or emboss it", plus a preview and an export path.

The other thing worth deciding up front is what you actually want geometrically, because they are different features:

1. **Cuts through the lid** (holes/slots), which is what a literal port of the cutout editor gives you.
2. **A shaped underside** that presses down onto the contents and stops them sliding, which is what your description of the problem actually calls for.

Reading your report, I think you want (2) rather than (1). A hole in the lid does not stop a tray's contents from sliding; a matching negative on the lid's underside does. That is closer to mirroring the bin's compartment layout onto the lid as shallow ribs than it is to the cutout editor.

Could you confirm which you are after? If it is (2), a natural first version is "project the bin's divider layout onto the lid underside as ribs, with a depth setting" — that reuses the compartment model you have already set up and needs no new editor at all.

I would rather build the thing that solves the sliding problem than the thing the title asks for, if those turn out to be different.

typytypytypy:
Thanks — I think I can describe the request more clearly now.

First, I want to emphasize that this is an incredible piece of software. I have been building and designing things for years, and this may genuinely be one of the most useful tools I have come across. The amount of functionality already built into it is impressive.

What I am really looking for is the full bin editor, with all of its existing functionality, but with the option to use lid attachment geometry as the bottom instead of a standard Gridfinity base.

In other words, the upper tray could be treated as a normal shallow bin:

- It could use the existing compartment editor.
- It could have dividers, pockets, custom cutouts, and shaped recesses.
- Its walls and interior could be configured just like any other bin.
- Instead of having a normal Gridfinity bottom, its underside would use the selected lid connection geometry, such as friction fit, click rails, or magnets.

[...]

PR #3259, opened 2026-08-06
refactor(designer): rename the wall-less base to "Base only"

andymai:
"Tray" was already taken **twice** in the bin designer, so the wall-less base was a third meaning:

| existing                                          | what it means                                                                                       |
| ------------------------------------------------- | --------------------------------------------------------------------------------------------------- |
| `binDesigner.lid.topSurface.tray` → "Tray"        | the lid's recessed top, with "Tray floor", "Tray depth", "Tray rim"                                 |
| `binDesigner.lid.section.tray` → "Organiser tray" | a bin whose underside is a lid (#3036) — `base.trayBottom`, `isTrayBottom`, "Create matching tray…" |
| `binDesigner.tile` → "Tray (no walls)"            | **this one**                                                                                        |

It also didn't match its own code, which calls it `base.tile` / `isTile` throughout.

## Why "Base only"

It is the term **this repo already reached for on its own**, three separate times, for exactly this shape:

```ts
MIN_HEIGHT: 2,       // height units (1U = base only, 2U minimum for usable cavity)
MIN_BIN_HEIGHT: 2,   // Minimum bin height in units (1U = base only, no usable cavity)
MIN_LAYER_HEIGHT: 2, // Minimum layer height in units (1U = socket base only)
```

A wall-less bin _is_ a bin that is only its base. It is Gridfinity-native too: `kennetek/gridfinity-rebuilt-openscad` — the source of truth cited at the top of `gridfinity.ts` — calls the foot assembly `gridfinityBase`. That project has no name for this variant at all, which is presumably why one had to be invented in the first place.

It also lives in the **Base** section directly under "Spacer (no floor)", so it reads as that toggle's pair rather than importing a fourth meaning of "tray".

## What changed

- **Label, hint and the six availability reasons across all 15 locales.** Each locale keeps its own phrasing — only the subject noun is swapped (`Ein Tablett` → `Eine reine Basis`, `Un plateau` → `Une base seule`, `トレイ` → `ベースのみのビン`, …), so no translation is rewritten wholesale.
- **The internals added in #3244 move with it:** `trayFloorHeight` → `tileFloorHeight`, `resolveTrayFloorThickness` → `resolveTileFloorThickness`. These matched neither the param nor the feature, and collided with the lid's own "Tray floor". They now match `base.tile` / `isTile`.
- **Prose and scenario names** through `context.ts`, `types.ts`, `shellStage.ts`, `rules.ts`, `base.ts` and `scenarios/tile.ts`.

## What deliberately did not change

- **`base.tile` itself.** `params` is hashed wholesale by `communityParamsFingerprint`, so renaming the key would shift the fingerprint of every published design and need a migration.
- **Both other trays.** Including the bare "tray" in `isSocketlessBase`'s doc, which means the tray-_bottom_ bin, not this one — the renames are explicit pairs rather than a blanket sweep, precisely to avoid that. Verified after: `topSurface.tray`, `LID_TRAY_*`, `isTrayBottom`, `DEFAULT_TRAY_BOTTOM` and "Create matching tray…" are all intact.

Also takes the review note left on #3255: the height-floor test now reads `GRIDFINITY.SOCKET_HEIGHT` instead of a hard-coded `5`.

## Verification

- 18,065 unit + dom tests pass; the tray/spacer/direct-mesh generator suites pass (71 tests)
- `typecheck` and `check:i18n` clean (3,582 keys matching across all 14 locales)
- lint identical to `main` (2 pre-existing `max-lines` warnings)
- The diff is **271 insertions / 270 deletions** — a symmetric rename with no behaviour change

---
## Summary by cubic
Renamed the wall-less base mode to "Base only" to avoid conflicts with other tray meanings and align with `base.tile`/`isTile`. UI copy and internals updated with no behavior changes or fingerprint shifts; minor i18n and prose fixes included.

- **Refactors**
  - Updated labels, hints, and availability reasons across 15 locales; only the subject noun changed.
  - Renamed `resolveTrayFloorThickness` to `resolveTileFloorThickness`, and `trayFloorHeight` to `tileFloorHeight`; comments and tests adjusted.
  - Kept `base.tile` unchanged to preserve `communityParamsFingerprint`.
  - Left other tray features intact (`topSurface.tray`, tray-bottom).
  - Tests now use `GRIDFINITY.SOCKET_HEIGHT` instead of a hard-coded `5`.

- **Bug Fixes**
  - Corrected Korean topic particle in four strings; tightened English comments/messages after the rename (assembledHeight gate, context collar note, base resolver doc, tile scenario error).

<sup>Written for commit 915348879528cec24547c6c63f095aa7de878298. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
All five review findings were valid and all five were artifacts of the rename, so they're fixed.

**Korean particle** is the substantive one. The strings inherited 는 from 트레이, which ends in a vowel; 빈 ends in a consonant, so it takes 은. Four strings were wrong (`tileDisablesStyle`, `tileDisablesSpacer`, `flatFloorDisablesTile`, `lidBaseDisablesTile`) — `tileDisablesWalls`/`Interior` used 빈에는 and were already fine.

That prompted me to audit **every** string I touched across all 15 locales for agreement rather than only the one flagged. The other 13 read correctly: the swapped subject is nominative and sentence-initial in each (`Eine reine Basis`, `Une base seule`, `Sama podstawa`, `En ren bas`, …), so no verb or article agreement moved with it.

The remaining four are English slips from the mechanical swap: `assembledHeight.ts` still said "a tray keeps its lip" in the base-only gate, `context.ts` came out with a duplicated article ("ignores the / the field there"), the `base.ts` resolver doc lost its comma after the subordinate clause, and the scenario error read "A base-only wall is 0" as if truncated.

Worth noting for the record: this is the second time Korean particle agreement has been caught on this feature — the original tray commit also fixed three Korean typos. A subject-noun swap is exactly what breaks it.

PR #3315, opened 2026-08-07
fix(security): resolve 14 findings from the 2026-08-07 audit

andymai:
Fixes 14 findings from the 2026-08-07 security audit, grouped into four themed commits. Every fix carries a regression test that fails on `main`.

## Resource exhaustion

**Telemetry distribution maps had no entry cap** (`api/lib/mlTelemetry/validators.ts`). The validators checked each key's format and each value's range, but `VALID_BIN_SIZE_REGEX` admits an unbounded set of distinct valid keys (`1x1x1.1`, `1x1x1.11`, …) and every key fans out into Redis hash fields with a refreshed 90-day TTL. One unauthenticated request could introduce hundreds of thousands of persistent fields in a datastore shared with rate limiting, sessions and sync. Added per-field caps that reject on overflow, matching the array validators alongside them, and mirrored the caps in the client builder so an oversized honest payload truncates rather than being rejected whole.

**`ml:meta:validation:failed_by_type` grew permanently** (`api/ml-telemetry.ts`). The existing `replace(/[^a-z0-9_]/gi,'_')` prevented key injection but not cardinality, and the hash is in `ML_LIFETIME_KEYS` so nothing ever expires or prunes it. Unknown types now collapse into a single `other` bucket against a known-type set.

## Moderation state

**Moderation state was bound to a design id the owner can shed** (`api/community.ts`, `api/sync/account.ts`). `DELETE` returns 409 and `PUT` returns 403 for a hidden design, but re-publishing the identical payload minted a fresh live design with an empty reports set, because the idempotency and duplicate checks only match live designs. Deleting the account did the same, since the cascade purges the card, the reports and the reasons outright.

Every takedown path now writes a content-hash tombstone (`community:moderated:content`) and publish refuses a match. It is keyed by content and holds no user identifier, so it survives both deletions while account deletion stays a real erasure. An admin restore lifts it.

_Known limit:_ the tombstone is the exact content hash, so an attacker who edits the description can re-publish. Binding it to the params fingerprint per author would close that, but it means retaining a pseudonymous author id past account deletion — a privacy trade I did not want to make unilaterally. Happy to add it if you'd prefer that balance.

**Print delete-then-repost** (`api/community/prints.ts`). The upsert path refused to edit a hidden print back into visibility, but deleting it first left nothing for that check to find. Deleting a non-live print now 409s, and creation re-derives status from the persisted reporter set.

**`unlike` was a takedown oracle** (`api/community/[id].ts`). The Lua toggle reads the like count off the card hash whatever the status, so a hidden design answered `{likes: N}` where a deleted one answers `{likes: 0}` — both 200s, so it looked uniform. Non-live designs now answer exactly as missing ones. The unlike still clears the heart, preserving the A15 behaviour.

**Account deletion ignored all community-print state** (`api/sync/account.ts`). Print records, list membership, reporter entries and uploaded photo blobs survived indefinitely, carrying the deleted user's display name and photos on any still-live design. The cascade now purges them via the reverse indexes documented for exactly this, and resyncs the affected designs' print counts.

## Content moderation

**`collectDesignText` missed most engraved text** (`api/lib/contentFilter.ts`). It pushed a string only when its own key was `text`/`label`/`name`, and array elements hit the not-an-object return before any key was consulted. `compartments.compartmentTexts[]`, `label.rowTexts[]`, `surfaceText.lidText` and `surfaceText.walls.*` are all accepted and stored by the designer validators and engraved into the geometry recipients see — and none of them was moderated. The parent key now travels with the recursion, and the engraved-text keys are on the allowlist (kept as an allowlist so enum values, hex colours and font names stay out of the blocklist).

**Designer shares skipped the filter entirely** (`api/share.ts`). The `type:'designer'` branch ran only structural validation. Extracted `filterDesignParamsContent` so both share types run one implementation.

**Share `authorName` skipped sanitization and the blocklist** (`api/share.ts`) while every sibling recipient-facing field gets both. Now sanitized and filtered, rejecting with `CONTENT_BLOCKED`.

**Terminal escape injection into the moderation CLI** (`scripts/community-admin/`). A design's `name`/`authorName` reach a moderator's TTY, and neither `trim()` nor the content filter removes an interior ESC — so an attacker could clear the screen, rewrite adjacent rows so one design's id reads as another's, or erase the DIRECT WRITE banner. Control bytes are now stripped at publish validation and replaced with `�` at render (replaced, not deleted, so tampering stays visible). Table widths measure the sanitized value, or a crafted name shifts every row below it.

## Client and auth

**Collaborator-controlled CSS injection** (`src/shell/Collab/`). `presence.color` is written client-side and validated by nobody, then interpolated into the `background` shorthand — which accepts a comma-separated `url()` layer, so no `;` breakout was needed, and the CSP is report-only. Now normalized against a strict 6-digit hex at each consumption boundary. Note the sibling `border`/`boxShadow` sinks on the same elements were never exploitable: only the shorthand accepts `url()`.

**Mesh decompression bomb** (`src/shared/generation/meshAsset.ts`). `decodeMeshData` inflated attacker data through `Response.arrayBuffer()`, which accumulates the whole stream, so peak memory was the full decompressed size before any of its size checks ran. Now inflated chunk by chunk against a ceiling derived from the triangle budget, cancelling on the first chunk that crosses it. The per-asset size cap the server enforces on shares is also applied to the local-import path, which stored `meshAssets` verbatim.

**OAuth cookies lacked `__Host-`** (`api/lib/cookies.ts`). The callback is a cross-site top-level GET so `checkCsrfDefense` cannot apply, the state comparison has no per-victim binding, and GitHub uses no PKCE — cookie integrity is the only defense. Without the prefix a sibling subdomain can plant these scoped to the registrable domain. All three cookies now carry it, mirroring the session cookie's prod/dev pattern.

## Not in this PR

- **Unsalted `deriveUserId`** (finding #12). Salting changes every existing user's primary key: their sync blob path, Redis keys, published-design set, and membership in every likes/reports set. It needs a lazy login-time migration, which belongs in its own reviewed PR rather than buried here. Opening separately.
- **Enforcing CSP.** Finding #6 also recommends promoting `Content-Security-Policy-Report-Only`. That needs report data reviewed and a staged rollout — it can break Vercel/PostHog/Liveblocks/blob-CDN loads in ways unit tests won't catch. The colour validation closes the actual injection.
- **`ml:meta:vocab_versions` / `client_versions`** take caller-controlled values from _valid_ events and are not format-validated. They do carry a sliding TTL, so growth is bounded in time rather than permanent — a lesser version of finding #3, flagging it rather than expanding this diff.

## Testing

`pnpm run typecheck` and the full `pnpm run test:run` suite pass. Each fix has a regression test; I verified the CSS-injection test fails against the unfixed code rather than passing trivially.

`CLAUDE.md` gains one gotcha entry covering the three patterns behind most of these: moderation binds to content not to a deletable id, a key-format check is not a key-cardinality check, and a new `BinParams` text field is unmoderated until its key is listed.

---
## Summary by cubic
Fixes 14 findings from the 2026-08-07 security audit. Hardens moderation, telemetry, content filtering, client inputs, mesh decoding, and OAuth cookies to prevent reposts, resource exhaustion, injection, and OOMs.

- **Bug Fixes**
  - Moderation: Block re-publishing taken-down content via content-hash tombstones (auto-hide, admin hide, denylist); admin restore clears it.
  - Prints: Block delete-then-repost (409 on hidden; create re-derives status); clear cover if from a hidden print; on account deletion purge prints/photos and resync only if the card exists; require every indexed entry to resolve before dropping reverse indexes; skip deletes when the record is missing (e.g., rotated salt) to avoid orphaned photos.
  - Likes: `unlike` on non-live designs responds like missing (still clears the heart; no takedown oracle/leak).
  - Content filter: Moderate engraved text across params (`lidText`, wall text, `compartmentTexts`, `rowTexts`); run the filter for `type:'designer'` shares; sanitize and blocklist `authorName`; strip control bytes from `name`/`authorName`; CLI output neutralizes control sequences; fail closed if the text walk truncates past its cap.
  - Telemetry (`ml-telemetry`): Cap distribution map entries (size/category/domain) to bound Redis fields; bucket validation failures by known types with an `other` catch-all for `ml:meta:validation:failed_by_type`.
  - Client: Normalize `presence.color` to strict `#rrggbb` at use sites to prevent CSS `url()` injection (applied in collab cursor/ghosts/selection rings).
  - Mesh assets: Stream-decompress with a decoded-bytes ceiling; enforce per-asset encoded size cap on local imports; skip oversized embedded assets when restoring.
  - Auth: Prefix OAuth state and PKCE verifier cookies with `__Host-` in prod (mirrors the session cookie).

<sup>Written for commit d51500641ed6d85770e60c7302603f25a77f5f1a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3315?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thanks — two of the three are real and now fixed in c2b23c3. The third is a pre-existing condition, not something this PR introduces.

**Issue 3 (count sync recreates deleted cards) — valid, fixed.** Confirmed by reading the order: the community cascade `DEL`s `communityDesignKey(id)` for every `publishedId` at account.ts:170-178, well before the print purge. `syncCommunityPrintCount` then `HSET`s that key, and `HSET` creates it — so a user who printed their own published design got the card resurrected as a malformed, never-expiring hash holding only `prints`. Exactly the index/blob drift `pnpm sync-admin` exists to hunt. Now resyncs only a design whose card still exists.

**Issue 1 (salt rotation strands print records) — valid, fixed, and worse than described.** The persisted-`authorPublicId` fallback only fires when `deriveAuthorPublicId` returns `null`, which happens when `TOKEN_SALT` is *unset* — not when it is *rotated*, where it returns a new non-null id that addresses keys that never existed.

The part the review missed: dropping `communityPrintedKey` wasn't the only leak. `deleteCommunityPrint` unconditionally `SREM`s the reverse-index entry itself, so the handle was destroyed even without the final `DEL`. My first attempt at this fix still failed its own test for that reason. The fix is to skip the delete entirely for a record we cannot read, and to keep the reverse index unless the purge resolved at least one record.

**Issue 2 (Safari 15) — not introduced by this PR.** `safari15` is a declared Vite target, but `main` already calls `new DecompressionStream('deflate')` unconditionally in `decodeMeshData`:

```
$ git show main:src/shared/generation/meshAsset.ts | grep -n DecompressionStream
80:  stream: CompressionStream | DecompressionStream
175:    raw = await pipeThrough(base64ToBytes(data), new DecompressionStream('deflate'));
```

`encodeMeshData` uses `CompressionStream` the same way, so mesh assets could never be created or read on Safari 15 either. This change swaps one unconditional use for another — same reachability, same failure mode. It's a real gap in the declared target, worth its own issue, but it isn't a regression here and gating it on this PR would mean the deflate bomb stays open meanwhile.

Both fixes carry regression tests, and I verified the card-resurrection one fails against the unguarded code rather than passing trivially.

andymai:
Thanks @copilot — two valid, one already stale. Fixed in d515006.

**`filterDesignParamsContent` fails open at the cap — valid, and the best catch on this PR.** `collectDesignText` stops at `MAX_TEXT_FIELDS` and the function returned `passed` on whatever it had collected, so the filter reported success on text it never read. This PR made it materially easier to reach: extending collection into arrays means `compartmentTexts` alone can contribute many entries, where before only a handful of scalar keys counted.

Fixed by collecting one past the cap so truncation is *detectable* — at exactly `MAX_TEXT_FIELDS` you cannot distinguish "fully collected" from "truncated", since both yield the same length — then rejecting when the sentinel is present. Failing closed is free here: every engraved-text field is individually bounded (`MAX_COMPARTMENT_GRID` 12, `MAX_INSERTS` 20, `MAX_MESH_ASSETS` 8, five `surfaceText` fields), so a legitimate design lands around ~200. Reaching 500 means the params were built to reach it.

**`purgeCommunityPrints` returning true on partial resolution — valid.** Now requires every indexed entry to resolve. Worth noting why this is safe against the idempotency concern that usually argues for the looser check: `deleteCommunityPrint` already `SREM`s each resolved entry, so a retry sees the index narrowed to exactly the unresolved strays rather than re-walking the whole set.

**Doc comment ordering — already fixed.** You reviewed 2d286d9; the "attempted first" clause went away when I rewrote that docstring in c2b23c3. Current text makes no ordering claim.

Both fixes have regression tests, including one asserting that padding cannot hide a blocked string past the cutoff.

PR #3318, opened 2026-08-07
fix(auth): make the user id unreversible

andymai:
Finding #12 from the 2026-08-07 audit, deferred out of #3315 because it touches identity for existing accounts.

## The problem

`deriveUserId` was `sha256(provider:subject)` with **no salt**. GitHub's `providerSubject` is a public, sequential, bounded (~2^28) account id, so the entire output space is precomputable. Anyone who obtained a single Redis key holding these ids — a `community:reports:{id}` set from a backup or a partial read — could match every member against a rainbow table and name the GitHub account behind each pseudonymous reporter, liker and publisher.

The sibling derivations (`deriveDonorId`, `deriveAuthorPublicId`) both salt with `TOKEN_SALT` and both document why. `deriveUserId` was the outlier, while its own docstring promised the value "can't be reversed into a Google/GitHub account id by anyone with read access to a single Redis key."

## The fix

A **random** id resolved through a **salted identity map**:

```
identity:{sha256(`${TOKEN_SALT}:identity:${provider}:${providerSubject}`).slice(0,32)} -> userId
```

Random rather than a salted hash is strictly stronger: a salted digest is still reversible to anyone holding both the salt and the input space, while a random id has no input left to recover.

Both halves are load-bearing. The map key has to stay salted — an unsalted `sha256(github:N) -> userId` entry rebuilds precisely the join the random id removes. That's why sign-in **fails closed with a 503** when `TOKEN_SALT` is unset rather than falling back to an unsalted key.

## Why existing accounts are adopted, not rotated

This is the part worth reviewing carefully.

On a map miss, the resolver falls back to the legacy derivation and, if a profile exists under that id, **adopts it into the map**. No key moves, no blob moves, nothing breaks. The account keeps its layouts, designs and published community records.

The consequence is that pre-existing accounts keep reversible ids. That's deliberate. `authorPublicId = f(userId)`, and `authorPublicId` is baked into:

- print-photo Blob paths, via `printPathSecret(designId, authorPublicId)` — **already-public CDN URLs**
- the `authorPublicId` stored in public design record blobs and card hashes
- `community:author:{authorPublicId}`, `community:print:{designId}:{authorPublicId}`, and the `community:prints:{designId}` ZSET members

So rotating an existing user's id means rewriting content that has already been handed out. Closing that residue properly means first pinning `authorPublicId` to a stored value rather than re-deriving it, which is its own change with its own risk — not something to smuggle into this one.

Net effect: every account created from here on is unreversible, and the exposure is capped at today's user set instead of growing.

## Operational note

Sign-in now depends on `TOKEN_SALT`, joining community blob paths, print photo paths and `authorPublicId`. A salt rotation already orphans all of those, so the salt is effectively permanent per deployment — this adds one more reason. Documented in `api/auth/README.md`.

## Testing

`pnpm run typecheck` and the api suite pass. New coverage in `api/lib/userId.test.ts` and `api/auth/auth.test.ts` for: the minted id not matching the unsalted hash, adoption of a legacy account, stability across sign-ins, the `SET NX` race between two concurrent first sign-ins, refusal without `TOKEN_SALT`, and the map key not embedding the raw subject.

I verified the two load-bearing tests fail against the old derivation rather than passing trivially — and caught one that _was_ vacuous while writing it (it compared against a hash of a subject the fixture never uses), which is fixed.

---
## Summary by cubic
Replaced reversible `userId` hashing with random IDs resolved through a salted identity map. Adopts legacy IDs by checking durable state (profile, sync indexes, published set) so dormant users aren’t stranded; new users get unreversible IDs.

- **Bug Fixes**
  - Added `resolveUserId` to map `(provider, subject)` via a salted `identity:{sha256(...)} → userId` and mint a random 32-hex ID.
  - On first sign-in, adopt the legacy ID when any durable state exists (profile, indexes, or published set) in one `EXISTS` round trip; otherwise mint a new ID.
  - Sign-in fails closed (503) when `TOKEN_SALT` is unset to avoid unsalted keys.
  - Race-safe first sign-ins using `SET NX`; never splits one identity across two IDs.

- **Migration**
  - Ensure `TOKEN_SALT` is set and treated as permanent for this deployment.
  - No action needed for existing users; their IDs are kept to avoid breaking public artifacts.

<sup>Written for commit 2e778e0d94e7c3423fc563409ba7f6b3dc5ed89b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3318?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Valid, and a good catch from both of you — fixed in 2e778e0.

I asserted in the code comment that the profile is "the honest test for 'this account already exists'". It isn't. `PROFILE_TTL_SECONDS` is one year, refreshed on sign-in, while `users:{uid}:index:{layouts,designs,baseplates}` and `community:published:{uid}` carry **no TTL at all** — I confirmed there is no `expire`/`EX` on any of them. So a user dormant for over a year has an expired profile and completely intact data, and the profile-only probe would have signed them into an empty account with everything they owned stranded under the legacy id and nothing left pointing at it.

The probe now covers the profile plus the four durable keys. `EXISTS` accepts multiple keys and returns a count, so it stays one round trip.

Two notes on the fix:

- **The deleted-account case still behaves correctly.** The deletion cascade drops all five of these keys, so a re-login after account deletion still declines to adopt and mints a fresh id — it does not resurrect the legacy id and re-adopt whatever the cascade may have missed. There is a test pinning that.
- **New coverage** is parameterised over each durable key individually, so a future change that drops one from the probe fails loudly rather than silently narrowing it. I verified all five new tests fail against the profile-only version rather than passing trivially.

Docs in `api/auth/README.md` updated to say durable state rather than profile.

PR #3393, opened 2026-08-10
perf(baseplate): paint the instant draft before the Manifold one

andymai:
The baseplate's two draft paths were an `else if`, so when the Manifold draft graduated it replaced the procedural direct-mesh rather than layering on top of it. 90 days of `baseplate_preview_timing` measure what that cost, split on `preview_kind` (n=119,111):

| path | state | p50 | p90 |
| --- | --- | --- | --- |
| `direct` | warm | 11ms | 54ms |
| `direct` | cold | 18ms | 72ms |
| `manifold` | warm | **348ms** | **1,288ms** |
| `manifold` | cold | **1,193ms** | **4,951ms** |

Manifold carried 60% of previews, so the majority of edits traded an ~11ms first frame for a ~350ms one, and a cold kernel for a ~1.2s one. Nothing fills that window: the canvas keeps showing the previous, wrong-size plate until the draft resolves.

The two are complements, not alternatives. The procedural mesh is synchronous and approximates magnet holes and chamfers; the Manifold draft is a WASM round-trip that runs the real generator. Paint the cheap one first and let Manifold, then BREP, refine it — the instant-then-refine order the bin designer already uses. The Manifold draft's direct-mesh fallback goes with it: the procedural mesh is on screen before the draft is dispatched, so a failed draft has nothing to fall back to.

## Where the policy lives

The ladder moves to `planPreviewDrafts`, alongside the predicates it composes, because it is three interacting rules and this is the second time one of them has been changed by editing a branch condition. Both skip rules are unchanged and now carry tests:

- a **shaped plate** takes neither draft — the procedural mesh draws rectangles only, and a Manifold draft repeats the exact build's own coplanar outline clip;
- an edit whose **last exact build came in under the `draftPolicy` gate** takes neither — that gate exists to stop a draft that would be replaced immediately, and painting the procedural mesh unconditionally would reintroduce the jump it prevents.

## Dead telemetry scaffolding

`previewKindRef`, `directMeshDurationRef`, `directMeshStartRef`, `firstBrepDoneRef` and `shouldTrack` fed `baseplate_preview_timing`, removed in #3380. They are write-only now, and `previewKindRef` ("which path produced the first on-screen frame") stops being answerable once both paths run.

## `scrubPerfBreakdown`

The harness that sized the rest of the stack. `previewPerfBreakdown`'s "warm" case regenerates one identical config repeatedly, so every cache holds it by the second sample — it measures cache *replay*, and reports 104ms for a height edit whose cold cost is 1,111ms. A scrub visits a new config per tick, so nothing replays.

Per tick, median of 8 distinct configs on a 3×3 magnet+lip+scoop+3×3-compartment bin:

| scrub | total | base | features | boolean | merge |
| --- | --- | --- | --- | --- | --- |
| height | 778ms | 348 | 81 | 253 | 97 |
| wall thickness | 769ms | 361 | 78 | 245 | 89 |
| scoop radius | 427ms | **8** | 78 | 248 | 92 |
| width | 1,188ms | 664 | 80 | 252 | 199 |
| replay | 112ms | 8 | 3 | 10 | 90 |

That lands on the production p75 of 1,255ms rather than the old harness's 104ms. The scoop row is the `shellKey`-stable one: when the shell cache hits, `base` collapses from ~350-660ms to 8ms. Tessellation is the floor — ~90ms even when every cache hits.

`shellStageBreakdown` then splits `base` further. Median of a 6-step height sweep, idle machine, warm lip/socket caches:

| piece | 3×3 | 6×6 |
| --- | --- | --- |
| box | 13ms | 12ms |
| lip | 1ms | 1ms |
| **fuse** | **226ms** | **222ms** |
| socket | 6ms | 21ms |

The box↔lip fuse is 92%/87% of those four pieces and costs the same on a 3×3 as a 6×6 — a fixed OCCT boolean, not footprint-scaled. Those four are a **subset** of the stage, so ~245ms is a lower bound on the ~350ms `base` a height tick pays; the remainder is compartment cavity drawings, `collectOrigins`, spanning-divider clips and the cache clone. Against the whole ~800ms tick the fuse is ~28%.

It cannot simply be deleted: `buildBinBoxWithLip` already emits that body fuse-free but is draft-only on purpose, because its single-prism construction leaves the flush wall↔lip seam of #1314 that `scenarios/regressions.ts` asserts against. The exact-path version is deferring the lip past the boolean stage the way the socket already is, which has to answer for features that cut through walls — a separate change.

## Verification

`pnpm exec vitest run --project unit --project dom` — 18,803 passed. Typecheck and lint clean (one pre-existing `max-lines` warning on the hook, whose length this change reduces).

---
## Summary by cubic
Render the instant procedural baseplate preview first, then refine with the Manifold draft; exact BREP still replaces both. This restores an ~11 ms first frame instead of waiting ~350 ms+ while keeping fidelity.

- **Refactors**
  - Added `planPreviewDrafts` to pick the ladder per edit: procedural direct-mesh → Manifold draft → exact BREP. Skips both for shaped plates and, when a preview bridge is available, skips drafts if the last exact build was under the `draftPolicy` gate; without a bridge we still paint the procedural mesh.
  - Reworked `useBaseplateGeneration`: always paint the procedural mesh first; Manifold draft refines it when available; keep the previous mesh when drafts are skipped. Removed dead preview‑timing refs.
  - Added `planPreviewDrafts` unit tests plus two kernel perf harnesses: `scrubPerfBreakdown` (per‑tick scrub costs that match production) and `shellStageBreakdown` (shows the box↔lip fuse dominates the shell stage).

<sup>Written for commit 2c1108e449efdbc22df4f7ae27b30717889c2d69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3393?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Review follow-up.

**`scrubPerfBreakdown` — `missed` naming.** Fixed, it's `replayHits` now. It did hold the hit-rate summary.

**`shellStageBreakdown` — double-free on the cached lip/socket.** Not a defect; the premise doesn't hold. `createCloningAccessors` makes *both* paths return a clone:

```ts
get(key)        { return cloneFromCache(cache, key); }   // unwrap(clone(shape))
set(key, shape) { cache.set(key, shape); return unwrap(clone(shape)); }
```

So `buildTopShape`'s `if (cached) return cached` is returning the clone `getLipCache` just made, not the cache's shape, and the miss path returns `setLipCache(...)` which is also a clone. Same for `buildBaseSocket` via `getSocketCache`/`setSocketCache`. The caller owns what it gets on both paths, so deleting is required — omitting it leaks. The harness also runs a 6-step sweep twice per test with the lip cache hitting from step 2 (`lip 1ms`); a freed cached handle would fail there rather than report a stable timing.

It did point at a real one next to it, though: **`box` was never deleted.** `fuse` does not consume its inputs — that's why `shellStage` registers `binBody` with a disposal scope — so the box clone leaked once per iteration. Fixed, with the ownership contract written down at the call site.

**Corrected numbers.** Re-running on an idle machine dropped the fuse from 340ms to ~226ms and it held within 228-238ms across four runs; the first run had gone straight after the full 18,803-test suite. The share is stable, the absolute is not, so the harness header now says to run it idle.

More importantly the original commit message overstated the result. 91% was the fuse's share of the four pieces the harness times, and I wrote it as the share of the stage. The four are a subset — the real stage also builds compartment cavity drawings, runs `collectOrigins`, applies spanning-divider clips and takes the metadata-preserving cache clone, which is the ~90ms gap between this harness's ~245ms and the scrub harness's ~350ms `base`. Against a whole ~800ms height tick the fuse is ~28%. Still the largest single operation in the pipeline, but not "the whole stage". Commit message corrected; the PR body's scrub table is unaffected and reproduced within ~3%.

No change to the shipped code in this PR — both fixes are in the profiling harnesses.

Issue #3402, opened 2026-08-10
Non-full length label otions


daemian:
**Problem**
customizable label tab length options

**Proposed Solution**

I like using centered cullenect labels, usually 1u. I would prefer not to have the label tab run across the entire width, but instead have it be just wide enough for the 1u label (or 2u or whatever). Basically just want a customizable length of label instead of always being full length.

Love this amazing tool. By far the best one I have seen for creating every manner of GF bin.

ficdogg:
This is already available, just a little deeper in the settings.

<img width="630" height="534" alt="Image" src="https://github.com/user-attachments/assets/c7478550-2996-4a63-af90-2b07ad191f4e" />

andymai:
The width slider @ficdogg screenshotted is real, but it does not apply to the mode you're describing. If you're using Cullenect click-in plates, the tab is in **socket** mode, and there the shelf is always forced to the full compartment width. The builder ignores the width percentage outright in that mode, and the control is hidden from the panel because of it. `alignment` only moves the socket pocket around inside an always-full-width shelf, which is why a centred 1u plate still leaves a full-width tab.

So this is a genuine gap, not a settings hunt. Reopening the width control for socket mode is on my list, with the plate fit setting the minimum so the pocket always survives, and the plate size following the tab rather than the whole compartment. That should give you exactly what you asked for: pick 1u, get a 1u-wide tab.

Tracking it separately from #3401, which turned out to be a related collision between tabs and lid click rails.

daemian:
amazing work. Thank you so much!

Issue #3401, opened 2026-08-10
Label tab interferes with click rails


ficdogg:
Label tab interferes with click rails

1. Add a full width label tab
2. Add a lid with click rails, 
3. Set rail coverage to 75 or higher

<img width="2198" height="1273" alt="Image" src="https://github.com/user-attachments/assets/a1f16b21-0858-4f97-8137-6737d45c7244" />

If the label tab is active, the click rails need to stop a couple of mm before the tab.

I'm making some (heavier) boxes for screws, so the lid needs to be very secure, a lid with 50% rails is a little too easy to open when loaded with screws. When I switched to 75% and 100% rail coverage, the rails are hitting the front of the tab.

Ideally we also get step increments of 10 instead of the 25 right now.

And if the tab is not full width, an option to add a partial click rail on the back side as well would make it even better.

andymai:
All three asks are merged.

**The rails hitting the tab** turned out to be a genuine geometric interference, not just a tight fit. On the assembled bin and lid the rail and the label shelf occupy the same 1.25mm of Z, so the lid physically cannot seat. Rails now stop clear of the tabs and re-centre on what is left, which means 100% coverage keeps meaning "as much rail as fits" instead of capping out below the tab (#3404).

Your "75 or higher" was exactly right and depends on the bin, not just the setting: a 2x3 bin with a 12mm tab is clear at 75% and collides at 100%, while a 2x2 bin with a 16mm tab already collides at 75%.

**Coverage steps** are now 5% from 50 to 100 (#3406). Finer than the 10 you asked for, for a specific reason: the stop list snaps saved values to the nearest option, and a 10% list has no 75, so every design already saved at three-quarter coverage would have quietly re-rendered at 70. 5% avoids that and costs nothing.

**Partial back rails** are in (#3408). Where the tabs do not cover the whole wall, the back rail now fills the gaps rather than the wall going friction-fit wholesale. No new setting: the per-side Back toggle already existed and was simply being overridden, so it just works now. For your screw boxes that should mean noticeably more grip than 100% coverage alone gave you.

Two things fell out of the investigation that were not in the report. Front-anchored tabs (`edges: 'front'`) were disabling the back rail rather than the front one, so the wrong wall lost its snap. And a shelf lowered clear of the rail band was losing its rail for no reason. Both fixed.

[...]

PR #3404, opened 2026-08-10
fix(lid): stop click rails running into label tabs (#3401)

andymai:
Closes part of #3401.

## The bug

A label tab's shelf sits inside the Z band the lid's click rails sweep, so on any wall **perpendicular** to the tab the two interpenetrate once coverage is high enough to reach it. The lid does not seat.

Measured on the real assembly (rail band `[33.85, 37.60]` above the cavity floor, shelf top `35.10`):

| bin | tab depth | coverage | interference |
| --- | --- | --- | --- |
| 2x3 | 12mm | 50% | clear |
| 2x3 | 12mm | 75% | clear |
| 2x3 | 12mm | 100% | **1.25mm** |
| 2x2 | 16mm | 75% | **1.25mm** |

That last row is the reporter's "75 or higher". It depends on bin depth and tab depth, not on coverage alone.

A second defect fell out of the same investigation: the conflict check hardcoded `back`, so `edges: 'front'` disabled a rail that was never in the way and kept the one that was.

## The fix

Rails clip their usable span short of any tab, then re-centre on what survives, so coverage applies to the clipped span and 100% still means "as much rail as physically fits". Shortening a centred rail instead would shed just as much length at the far end, where nothing was in the way.

Which wall a tab blocks is now derived from the tab geometry (anchor edge, inset, shelf height), so front-anchored tabs disable the front rail, and a shelf tucked under the rim (#1898) keeps its rail because a rail passes above it.

## Why the tests look like this

Bin and lid are separate solids. Both stay watertight, both keep plausible triangle counts, and every bounding-box assertion passes while the lid physically cannot seat (CLAUDE.md gotcha #10). The regression test mates the two and probes inside the assembly. With the clipping disabled it fails 6 of 8 cases and still passes the 2 baselines, so it is not vacuous.

## Refactor note

Tab planning moves to `@/shared/utils/labelTabPlan`. Three things now have to agree on where a tab ends up (the builder that cuts it, the rails that must miss it, the panel warning that explains it), and re-deriving that in each is the drift CLAUDE.md gotcha #9 warns about. `planLabelTabLayout` was already pure and documented as separate from the build, so this is a move rather than a rewrite, verified behaviour-neutral against the existing scenario snapshots before anything was built on top. `labelTabPlanDims.test.ts` pins its interior-dimension mirror to the pipeline's own `deriveDimensions` across 10 configurations.

`labelTabBuilder.ts` drops from 1513 to 1076 lines as a side effect.

## Not in this PR

The other two asks on #3401 (10% coverage steps, partial back rail in the gaps between tabs) and #3402 follow separately.

## Verification

- `lidLabelTabClearance.scenario`: 8 real-kernel assembly cases
- `labelTabPlan.test.ts` (14), `lidClickRailLabelClip.test.ts` (6), `labelTabPlanDims.test.ts` (10)
- 362 existing tests across label tabs, sockets, lid, tray and the designer panels
- typecheck, lint (0 errors), all four i18n checks, module boundaries

---
## Summary by cubic
Stops lid click rails from colliding with label tabs by clipping rail spans around tab footprints, including on overhang bins, and by deriving which wall to skip from tab geometry. Keeps 100% coverage meaning “as much rail as fits” and fixes front-anchored tabs disabling the wrong rail. Addresses #3401.

- **Bug Fixes**
  - Clip each wall’s usable rail span against tab footprints, re-center, and apply coverage to the clipped span.
  - Derive blocked wall from tab geometry (front/back/both); shelves tucked under the rim keep their rail.
  - Footprints now mirror pipeline interior including overhang expansion, so clipping works on symmetric/asymmetric overhang; scenario probes rails using the lid’s own bounds.
  - Lid inputs carry `labelFootprints`; compatibility warning shows the correct sides.

- **Refactors**
  - Moved tab planning to `@/shared/utils/labelTabPlan` shared by the builder, rails, and UI; added an interior-dimension mirror pinned by tests.

<sup>Written for commit 0cf6e90f604207f5e1df2bf7fff0b7a03a8473d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3404?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Good catch on the overhang frame, and it was a real defect rather than a theoretical one. Fixed in 0cf6e90.

`labelTabInteriorDims` skipped the overhang expansion, so on an overhang bin the footprints came out several mm narrower than the real cavity. The failure mode is worse than a small misplacement: the rail's cross-axis check then found itself outside the footprint's X span altogether and skipped that footprint entirely, so clipping never ran and the original collision came straight back.

My dimension-mirror test had no overhang case, which is why it passed. There are now two (symmetric and asymmetric), plus symmetric and asymmetric cases in the assembled-solid test. I verified the symmetric assembly case fails without the change, so it is not a vacuous guard.

One part of the report I checked and did not act on: the interior and outer offsets need no handling. Both are `overhangExpansion().offsetX/Y`, so the footprint frame (interior-centre-relative) and the rail frame (lid-centre-relative) share a centre and the offsets cancel. The asymmetric case passes with or without the expansion fix for exactly that reason, which is why the symmetric one is the load-bearing guard here.

Also hardened the probe: positions now come from the lid's own bounding box instead of a re-derivation of its width, so an arithmetic slip cannot move every probe off the rails and quietly report clean.

PR #3407, opened 2026-08-10
feat(labels): apply tab width in socket mode (#3402)

### LINKED ISSUE #3402: Non-full length label otions

andymai:
Closes #3402.

## The gap

The width slider a commenter pointed the reporter at is real, but it **does not apply in socket mode**, which is exactly the Cullenect click-in case the reporter uses. `fitTabInSpan` read `socket ? availableWidth : (availableWidth * widthPercent) / 100`, so the shelf always spanned the whole compartment. The panel hid the control because of it, and `GhostLabelTabs` hardcoded 100 to match. `alignment` only moved the socket pocket around inside an always-full-width shelf, which is why a centred 1u plate still left a full-width tab.

So "pick 1u, get a 1u-wide tab" was not reachable at any setting.

## The fix

The percentage now applies in both modes, and the socket plan is fed the same value so **plate choice follows the tab, not the compartment**. On a 3u-capable compartment: 100% keeps 3u, 70% picks 2u, 40% picks 1u, and below that no standard plate fits so the tab degrades to a plain shelf (the existing `plateWidthU === null` path, which the panel already warns about).

Sizing the plate off the compartment and clamping the width to it would invert the control the user is holding, and would leave a pocket planned for room the shelf no longer has.

Alignment now positions the whole tab, matching what it already did in text mode. At width 100 all three alignments coincide, so a full-width tab does not move.

## Migration

Existing socket designs are reset to width 100. The control was hidden, so a value stored from an earlier stint in text mode was never visible and never applied. Letting it suddenly take effect would narrow a printed part nobody asked to change. Two migration tests cover this and the text-mode case that must be left alone.

## Verification

- `labelSocketPlan.test.ts` (30, incl. 4 new plate-step cases), `labelTabPlan.test.ts` (16, incl. 2 new)
- `paramMigration` (112, incl. 2 new), LabelTabsSection + GhostLabelTabs + label socket scenarios + export + printEstimates (194)
- typecheck, lint (0 errors), all four i18n checks, module boundaries, design-system check

---
## Summary by cubic
Applies tab width in socket mode so plate choice follows the tab, not the compartment, fixing #3402. Alignment now moves the whole tab, and the width control is visible in socket mode.

- **New Features**
  - Width percentage now applies in socket mode; planners and ghost use the same value so 3u → 2u → 1u as you narrow.
  - Alignment positions the entire tab in both modes; hidden at 100%.
  - Panel shows the Width control in socket mode; summary text includes the width; “Widen tabs” suggestion is available.

- **Bug Fixes**
  - Dropped the migration reset of socket widths; values now persist across reloads.
  - Width readout reports the narrowed span in socket mode.

<sup>Written for commit fdb9997ae87e5b787b9f29d2d2f3dc0aecd4c4f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3407?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/bin-designer/components/panel/LabelTabsSection/LabelTabsSection.tsx
+++ b/src/features/bin-designer/components/panel/LabelTabsSection/LabelTabsSection.tsx
@@ -352,10 +352,9 @@ export function LabelTabsSection() {
 
               {/* Alignment — hidden at width=100% because the control has no
                   visible effect when the tab spans the whole column (#1898).
-                  Always shown in socket mode, where it positions the socket
-                  within the (always full-width) tab instead. */}
-              {(state.isSocketMode ||
-                state.label.width < DESIGNER_CONSTRAINTS.MAX_LABEL_TAB_WIDTH) && (
+                  Since #3402 that holds in socket mode too: the width applies
+                  there, so a full-width socket tab has nothing to align. */}
+              {state.label.width < DESIGNER_CONSTRAINTS.MAX_LABEL_TAB_WIDTH && (
                 <div>
                   <span className="mb-1 flex items-center gap-1 text-xs font-medium text-content-secondary">
                     {t('binDesigner.tabAlignment')}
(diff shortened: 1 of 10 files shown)


andymai:
Both findings were real. Fixed in 984d5f9.

**The migration was the serious one.** `migrateParams` runs on every load (via `DesignerStorage` and `designJson`), not once, so forcing socket designs to width 100 was not a one-time normalisation. It wiped the new control's value every time a design reopened, which made the feature impossible to persist at all.

It cannot be salvaged by making it smarter. There is no schema version on `BinParams` to key a genuine one-time migration on, and the migration cannot distinguish a width the user deliberately set in socket mode from one left stale by an earlier stint in text mode. Adding a version field would shift `communityParamsFingerprint` for every published design, which the bin-designer README explicitly warns against.

So the reset is gone. The trade is the right way round: the stale case needs a specific history (set a width in text mode, switch to socket, never touch it again), is visible immediately in the preview, and is now fixable from a control that is no longer hidden. The round-trip bug broke the feature for everyone who used it.

Worth noting how I nearly missed this: my first attempt at a regression test compared the migrated value against itself and passed vacuously. Writing the assertion honestly is what surfaced that no correct version of the migration exists.

**The readout was a third copy of the same assumption.** `tabWidthMm` still hardcoded 100% for socket mode, after the builder and the ghost overlay, so the W x D x H summary described a shelf the generator does not build. Fixed, with a test asserting the readout halves when the width does.

On the bin-designer README note from the 3406 review: taking that separately so it does not ride along in a PR titled for something else.

PR #3408, opened 2026-08-10
feat(lid): keep back rails in the gaps label tabs leave (#3401)

andymai:
Addresses the third ask on #3401: *"if the tab is not full width, an option to add a partial click rail on the back side as well would make it even better."*

## The problem

A tab narrower than its wall leaves usable run either side of it, and the lid threw all of it away. `computeDisabledRails` put the anchor wall into `disabledRails`, so the whole wall went friction-fit regardless of what the tabs actually covered.

## The fix

Clipping and segmenting turn out to be the same operation. `railSegmentsClearOfLabelTabs` subtracts each blocking footprint from the wall's run:

- A tab on a **perpendicular** wall eats one end and leaves one stretch. That is #3404's behaviour, unchanged.
- A tab on the wall the rail runs **along** eats the middle. Full-width leaves nothing, so that wall stays friction-fit exactly as before. Narrow leaves a stretch either side.

Coverage applies per stretch, so a wall with two gaps gets two rails rather than one rail sized for a run it cannot use.

`computeDisabledRails` therefore stops acting on the label-tab issue's `sides` (`SIDES_ARE_ADVISORY`). Deciding the wall is off up front discards the gaps before anything measures them. Cutouts and intruding handles still disable outright, because they remove the lip material a rail grips along the whole wall, so there is nothing to segment around.

## No new control

`clickRails.back` already existed and was simply being overridden, so honouring it is the whole feature. Worth being explicit about the consequence: **a saved design with the back toggle on and a narrow tab gains rail segments it did not have.** A design with a full-width tab is unchanged, and a design with the toggle off is unchanged. If you would rather this never touched saved designs, it wants a new opt-in field instead and I will add one.

## Verification

- `lidClickRailLabelClip.test.ts` (8) covers the segmentation directly, including the per-compartment gap case
- `lidClickRailPartial.test.ts` (6) asserts on real `railPlacements`, not a re-derivation: full-width tab emits none, centred narrow tab emits two symmetric segments, left-aligned emits one on the right, the user's off toggle wins, and coverage shrinks each segment rather than the wall
- `lidLabelTabClearance.scenario` grows two cases proving the **new** back segments do not themselves collide on the assembled solids (12 total)
- Three tests that asserted the old "label tabs disable the back rail" contract were updated; each is an intended contract change, and the replacements assert the new behaviour rather than deleting the coverage
- 338 tests across lid, compat, panel, tray and clearance
- typecheck, lint (0 errors), all four i18n checks, module boundaries, design-system check

---
## Summary by cubic
Keeps back rails in the gaps label tabs leave by segmenting each wall’s rail run, and fixes grip matching so off‑centre rails still split and polygon lids don’t cross‑cut parallel edges. Designs with `clickRails.back` on and a narrow tab now get back‑rail segments; full‑width tabs and designs with the back rail off are unchanged. (Addresses #3401)

- New Features
  - Segment rail runs around label tabs; emit one rail per clear stretch; apply coverage per stretch.
  - Treat label‑tab sides as advisory in compatibility; keep the rail toggle live for tabs; cutouts and intruding handles still hard‑disable.

- Bug Fixes
  - Match grip reliefs to rails by overlap, not centre equality, so off‑centre or segmented rails still split.
  - Share a `subtractSpan` helper; supports multiple grips and drops rails fully behind a grip.
  - Match a grip to its rail on the wall line (cross‑axis), not just orientation, to avoid cutting parallel edges on polygon lids.

<sup>Written for commit c4400796d8bff1f46b8e3747d8205fc3dccd07c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Real, and worse than scoped to this PR. Fixed in 6cbc47e.

`splitRailsAroundGrip` found a rail's grip by comparing the two centres for exact equality (`< 1e-6`). That held only while every rail was centred on its wall. **The clipping from #3404 already breaks it** — a rail shortened to clear a label tab sits off-centre, fails the match, and is passed through whole, so the relief's requested snap-softening silently does not happen. My regression test reaches the bug through that path alone, with no partial rails involved, so this is already on main rather than something this PR introduces. This PR widens it.

Matching is now by overlap, and the split subtracts the grip's blocked span from the rail's own span instead of assuming the grip sits at the rail's midpoint. That also fixes three cases the old symmetric arithmetic never handled: an off-centre rail, more than one grip on a wall, and a rail lying wholly behind a grip (which now drops out rather than becoming two slivers).

The subtraction is the same operation the label pass needed, so both now go through one `subtractSpan`.

Nothing about the old behaviour was visible in the output: a lid with a relief and an un-split rail is a perfectly valid solid. `lidGripLabelInteraction.test.ts` asserts on real `railPlacements` + `splitRailsAroundGrip` output, with the "no label tabs" baseline included so the test cannot pass by the grip pass doing nothing at all.

andymai:
Correct, and my overlap match made it materially worse. Fixed in c440079.

Orientation plus along-overlap does not identify a wall on a polygon lid. On a 3x3 U-shape with right-facing reliefs, the outer right rail came out in **three** pieces instead of two: its own relief cut it once, and the notch prong's relief reached across and cut it again. The old centre-equality match was narrower and mostly avoided this by accident, so widening the match is what exposed it.

The cross-axis position is what identifies the wall. Both placement paths put a grip exactly `lidCornerR` outboard of its own rail (`gripPlacementsForRectangle` uses `halfD`, the rail uses `halfD - lidCornerR`; the polygon path offsets by `fitClearance` vs `fitClearance + lidCornerR`), while two parallel same-facing edges are at least a grid unit apart. So a `lidCornerR`-sized gate separates them by roughly 10x. Matching on side name has the same aliasing defect, which is why the original code avoided it, and the fix keeps that property.

One correction to the report's framing, in case it matters for how you read the diff: `grip.sides.right` gives **every** right-facing edge its own relief on a polygon, so both lines being interrupted is correct behaviour, not the bug. The tell is how many times a single line is cut. My first attempt at the regression test asserted on "how many lines changed" and failed against correct output for exactly that reason; the assertion now counts fragments per wall line, and I verified it reports 3 with the check removed and 2 with it in place.

PR #3526, opened 2026-08-16
[superseded by #3531] feat(bin-designer): add an underside lightweight floor mode

andymai:
Superseded by #3531 — same branch, same commits, merged as `47cd4db440`.

GitHub closed this PR when #3528's base branch was deleted on merge, and a subsequent force-push to `feat/underside-lightweight-relief` made it unreopenable (`state cannot be changed. The branch was force-pushed or recreated`). Pull requests cannot be deleted, so this stub remains as a pointer.

No work was lost. See #3531 for the review history and #3524 for the originating issue.

PR #3531, opened 2026-08-16
feat(bin-designer): add an underside lightweight floor mode

### LINKED ISSUE #3524: Inverted Lightweight Floor option for Bins
**Problem**
Currently the Lightweight Floor feature removes material from the inside of the bins when enabled, meaning it is unsuitable for small or flat components that may get caught in the holes. It also cannot be combined with finger scoops.

**Proposed Solution**
An "Inverted" Lightweight Floor option that removes the same shape of material from the **bottom** of the bin (middles of each socket).
Printers have become so much better at bridging since Gridfinity was designed, that almost _all_ the middle of each socket can be removed, with no loss of rigidity.

This would allow for comparable material saving while still being able to keep a flat bottom to the interior of bins, and the ability to combine with finger scoops.

**Alternatives Considered**
I have currently been exporting as STEP file into Onshape, imitating a cruder version of the Lightweight Floor, and then re-exporting that to my slicer.

Image

I have currently tested this socket border incrementally down to 4.8mm with success. Will test successively further reductions until issues arise.

Without a doubt this _works_ but it adds an extra manual adjustment step into the production of each bin, as the website currently eliminates the need for all other adjustments I would be making (great job!)

andymai:
_Replaces #3526, which GitHub closed when #3528's branch was deleted on merge. Same branch, same commits._

Closes #3524.

## What

The lightweight floor hollows each foot from the inside, which exposes the grid shape in the bin floor. Small or flat items settle into the recesses, and it rules out every feature that needs a floor to stand on. This adds the inverted variant the issue asks for: the same material taken from underneath, interior left flat.

`base.lightweightMode: 'interior' | 'underside'`, optional and **absent** when interior, so no published design's params fingerprint moves.

## Geometry

One new open direction in `lightweightBaseBuilder`: zero z-shift and no floor opening, so the bin's own floor caps the ring. `'down'` (which solid bins already used) leaves a `wallThickness` membrane under a floor that is already solid — a slab of dead material across the footprint that does not even shorten the bridge above it. `'down'` survives only as interior mode's answer for a solid bin, so designs saved before the mode still build what they were published with.

The offset is a 3 mm border rather than `wallThickness`: this ring stands on the bed and carries a bridged floor. It is safe at any width because **the cut is a scaled copy of the socket profile** — the profile's insets are absolute, so an inner foot built at `cell - 2k` is a uniform `k` wall at every depth and cannot reach the baseplate-mating taper. A straight prism, which is what the issue's Onshape model uses, would have to clear `SOCKET_TAPER_WIDTH` (3.2 mm) before it left any wall at all, since a foot's bottom face is that much narrower than its cell.

Measured against a solid base: **5174 mm³ per cell** — 38% of a 1x1x3, 45% of a 2x2x3, 48% of a 3x3x3.

## What the relief unlocks

`dimensions.lightweight` conflated "the base is shelled" with "the interior floor is gone". Those come apart here, so gates now read a separate `liteFloorOpen`. The relief therefore keeps:

- **finger scoops** — the headline ask in the issue
- **drainage holes** — the slab is still there, and the relief cavity vents to the outside
- **top cutouts** — `buildCutoutCuts` already clamps every pocket to `>= Z=0`, and the relief is entirely below it, so the two bands cannot meet
- **the instant preview draft** (`canBinUseDirectMesh`) — the change is invisible to the preview camera, exactly like magnet holes

Floor inserts stay out of **both** modes: a recess in a `wallThickness` floor with open air beneath does not thin it, it holes it.

A **base-only bin** can now be lightweight, in this mode only. It is feet plus one thin plate, so it gains most; the old block was an accident of being bundled with the spacer rule.

The mode selector sits outside the `FeatureToggle` — the only sub-control in `BaseSection` that does. `FeatureToggle` renders children only while it is on, and the mode is what decides whether it *can* be turned on, so nesting it would leave a bin with a finger scoop no way to reach the feature that exists for it.

## Print estimate

The relief's material saving is booked as a per-cell term, measured rather than derived: generating each variant and differencing against a solid base gives a figure constant to the millimetre across 1x1, 2x2 and 3x3 — 6398 (interior) / 5174 (underside) / 4332 and 2458 on half sockets, mm3 per 42mm cell. A base-only bin measures 5173, i.e. the same relief, confirming the saving is a property of the foot alone and transfers to any body sitting on it.

Measuring for that model turned up an unrelated defect in the baseline it subtracts from; that fix landed separately in #3528.

## Verification

Bounding box, triangle count and watertightness pass on every variant here — the outer profile is deliberately identical, because the foot still has to mate with a baseplate. So:

- **`undersideRelief.kernel.test.ts`** — column probes on the fused export solid: the socket band is hollow at the bore and the floor above it is not; the interior mode is its exact complement at the same column; the ring is still standing where the part meets the bed; and `sectionHalfWidth` matches a standard bin to 6 decimal places at nine depths through both tapers.
- **`binSeating.kernel.test.ts`** — a relieved bin dropped onto a generated baseplate still seats, and is still placement-dependent in the same way. "The profile is identical" and "it still drops in" are different claims.
- **13 new scenario cases** — magnets, screws, half sockets, fractional footprint, the export fuse (a *partial* coplanar face at the socket top, unlike a solid foot's full one), dividers, base-only, and the three feature combinations only this mode can carry.
- Constraint-engine tests for each split rule pair, including that a bin with a scoop can actually reach the mode.

`pnpm run quality` clean, full suite green (22,713 tests).

---
## Summary by cubic
Adds an underside lightweight floor mode that shells feet from below while keeping the interior floor. Previously, interior lightweight opened the floor and disabled floor‑dependent features; underside keeps scoops, drainage holes, top cutouts, and direct‑mesh preview, while inserts remain unavailable.

- API/validation: `base.lightweightMode: 'interior' | 'underside'` (absent = interior); only these values are accepted; existing designs are unchanged.
- Engine/geometry: adds `dimensions.undersideRelief` and `dimensions.liteFloorOpen`; `lightweightBaseBuilder` adds a downward, no‑floor‑opening path with a 3 mm border ring using a scaled socket‑profile cut (outer profile and seating unchanged). Updates `shellStage` (open direction), `floorPatterns` (applies when the floor isn’t open), `binDirectMesh` (allowed for underside), `generationTimeout` (budgets a floor only when built), `scoopRampBuilder` (blocks only when `liteFloorOpen`), and constraint rules (mode‑aware). Base‑only bins can be lightweight via underside.
- UI: the mode selector remains outside the Lightweight toggle and is folded behind a disclosure; a non‑default choice stays visible.
- Estimates/tests/copy: print estimates subtract a per‑cell `lightweightBaseSaving` on the refit 5.3785 mm³/mm² baseline; adds `undersideRelief.kernel.test.ts` and expands scenario/constraint/validation coverage; locale strings guide switching to Underside for features that need a solid floor.

<sup>Written for commit adc93e0568e34586b316bbbf624cb601586688bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3526?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


JustCozzie:
Mad turnaround time on implementing this. I've just done a test print and, while the first layer of the bin floor is expectedly rather rough (on the bottom, where you'll never see it) by the time it reaches the top layer of the bin floor there is only an almost imperceptable difference in light reflection over the bridged spot.

You actually took it further than I was willing to risk (chamfering the cutout edges _out_, rather than _in_, to save more filament) and it still printed with zero downsides.

I will be printing all my Gridfinity items with this base going forward.

55a65f17-b715-4690-92e5-aa06890abd2f
972b9c3d-569a-4ebc-98fb-f9940a19e08c

Comparing these two prints, the new feature was ~5% Cheaper on filament, and I imagine that's an absolute minimum, given it's a 1x1 with tall height (5u), finger scoop, and label bracket. 
Also ~4% quicker to print, which was a surprise (I was expecting an increase due to how much slower printers go while bridging).

Even at those minimum savings that would make a **huge** difference over a whole Gridfinity layout.

PR #3571, opened 2026-08-17
refactor(bin-designer): restructure the designer sidebar hierarchy

andymai:
## What

Reorders the parameter groups and rebuilds the Base section around one exclusive choice instead of four cross-disabling toggles.

**Group order and membership.** After Shape, the groups read down the part: Lid caps the top, Interior is inside, Base is underneath. Multi-color and physical units move out of Base, which described neither, into a new Finishing group.

**Base.** It was 13 flat rows mixing five unrelated concerns: a rim feature, mounting hardware, foot geometry, floor treatment, and four archetype flags. `base.flat`, `base.spacer`, `base.tile` and `base.lid` are already mutually exclusive (every rule that disables one names another as its source), so they become a single Body type picker. What remains is grouped under Stacking, Mounting, Feet and Floor, and a family renders only where the chosen body has it.

Two things follow from the picker. Selecting a type clears the outgoing one first, so every archetype is reachable from every other and the "cannot combine X with Y" reasons those pairs used to produce are unreachable. And applicability is asked of the constraint engine (`available || enabled`) rather than kept as a second table, so a rule added later moves the UI with it.

**Progressive disclosure.**

- The lightweight relief mode nests inside its toggle. It sat outside deliberately: a bin with a finger scoop is blocked from the interior relief but allowed the underside one, so choosing the mode is what unblocks the toggle. The blocked toggle now offers that directly, reusing the Fix affordance the lid's compatibility list already uses. Whether the underside relief would help is asked of the engine against a hypothetical selection, not matched against a list of fixable reasons.
- The foot lattice disappears when both axes are locked, and the explanation moves under the control that caused it. Nothing is concealed: a locked axis already reported the default lattice, so there was never a stored customisation in force behind it.

**Also in scope.** i18n for the magnet and screw labels, which were hardcoded English while every neighbouring label went through `t()`. `SubHeader` and a shared `ModeCard` promoted out of `LidSection` and `InteriorSection` so future sections inherit the hierarchy. Group summaries report actual state (a spacer with drainage holes read "Standard (no attachment)"). A dot marks groups holding an edit.

## Verification

- 5334 tests pass across `bin-designer`, `shared/components` and `shared/constraints`; typecheck, lint (0 errors), knip and all four i18n checks clean.
- `bodyType.test.ts` walks all 25 archetype transitions, which is what establishes that clearing the outgoing type makes every card reachable.
- Checked in the running app across all five body types: Mounting and Feet vanish on a flat base, Mounting vanishes on a spacer while Feet stays, and the tray's mating controls sit inside the tray card.
- The card glyphs were iterated against a render at card size; the first version's 3x3 feet were invisible at 20px.

## Notes

- 24 new keys plus 3 renamed labels, translated across all 14 non-English locales. `binDesigner.base.bodyType.standard` is allowlisted in `i18n-values-allowlist.json` because "Standard" is genuinely the correct word in five of them.
- `surfaceText` and `textDefaults` are excluded from the modified-dot comparison: both are edited from two groups, and a dot that is sometimes wrong cannot be trusted anywhere.
- Help entries for colors and physical units retarget from `binDesigner:base` to `binDesigner:finishing`.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3571?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed the review finding, plus a regression/confusion audit.

**The reported defect was real.** `matchesDefault` read an absent key as different from the populated default, so a persisted design omitting `featureColors` or `floorPattern` lit its group's dot over a field the owner never touched. Reproduced, fixed (absent is unmodified), and covered for the four keys that can legitimately be missing.

**Audit findings, in the same push:**

- The foot lattice explained its absence only when half sockets caused it. A custom shape and a fractional axis lock it too, and both are decided in the Shape group where there is no local control to attach the reason to, so those bins lost the control silently. The reason now renders in the slot the lattice would have occupied whatever the cause.
- The dot had an sr-only label and nothing for a sighted user hovering it, where a bare dot reads as "unsaved". It carries a `title` now.

**Invariants added rather than argued:**

- For every body type, no subsection is hidden while a feature inside it is still on. This is the property the whole hiding decision rests on; it passes for all five.
- A body type switch is undoable, restoring both the archetype and whatever the engine cleared. That is what makes the hiding recoverable rather than destructive.
- `base` is byte-identical after every A→standard→A round trip, so a bin that visits a body type still fingerprints as one that never did.

**Two losses pinned as pre-existing, not regressions.** Switching to a spacer clears the divider slots and coming back does not restore them; a base-only bin returns at the height floor rather than its original height. Both come from the constraint engine and the base-only commit path, which the four toggles routed through identically. The picker makes them easier to reach, not different. Both are now recorded in tests so the next change here can tell an intended loss from a new one.

**One thing I did not change:** `resolveConstraints` returns `autoDisabled` naming exactly what it cleared, and nothing in the app surfaces it. That gap predates this PR but matters more now that the row also disappears.

PR #3579, opened 2026-08-18
feat(layout): check the printed layout against the measured drawer height

andymai:
Closes the height half of discussion #2374.

## The defect

The layout's height budget was unit arithmetic. Layer heights summed against `drawer.height`, both in height units, and printed parts are not unit multiples: a bin stands `LIP_PROTRUSION_MM` (4.3mm) above its body, and a stacked bin settles `STACK_JUNCTION_MM` (4.75mm) into the one below. A column that exactly fills the unit budget therefore prints `4.3 - (n-1) x 0.45` mm proud of the drawer it was sized for.

Worked example. A drawer measured at 55mm floors to 7.85u. `useBinInspector` offered 7.85u as the bin's max height, and `SingleBinInspector` rendered "Prints 59.25mm" on the same line as that max. Over by 4.25mm, no warning on any surface. That is the reported "spill issues with lids closing down enough".

Three checks existed and none covered it: the layer budget is in units, the measured-drawer overflow covered width and depth only, and the one real fit check (`useAssembledHeight`) was designer-only.

## What changed

**One derivation, reachable by every consumer.** `assembledHeight.ts` moves to `shared/printSettings/`. `bin-inspector` and `layers` cannot import `bin-designer`, so the derivation had to reach `shared/` for a linked bin to be measurable at all. Its imports of `bin-designer/types/{base,lid}` stay put, matching `dividerRailPlan`, `lidInteriorRelief` and `labelTabPlan`, which already sit in `shared/` and import the same modules.

**The ceiling model.** `drawerCeilingFit` walks each column bottom-to-top by footprint overlap, summing the baseplate rise, the nested stack, the lip, a linked design's lid and any declared `clearanceHeight`. The ceiling is `drawer.measuredMm.height` only, never `drawer.height * heightUnitMm`, which is the measurement floored to 0.01u and would report overflow on layouts that fit. Same rule `useAssembledHeight` already applies.

**Linked designs measure as themselves.** `registryHeightFields` mirrors `registryEdgeFields`, carrying `assembledRiseMm` and `socketless` on the custom-bin registry so the check stays synchronous. Entries saved before the field, and imported meshes with no params to derive one from, measure as plain bins, which is the previous behaviour.

**Four surfaces, all advisory.** Drawer dimensions summary, bin inspector height hint, layers budget, and a ceiling plane in the isometric preview. Nothing blocks: the ceiling is a number the user typed, and they may deliberately print proud of it.

**The solver answers the ceiling question.** It held a target height and back-solved `heightUnitMm`, which lands exactly on the target but yields a non-standard unit, so the bins stop stacking with stock Gridfinity, and it rewrote a global every bin in the layout reads. It now holds the unit and reports the tallest whole-unit bin for one, two and three stacked bins, with the leftover shown so dead space is a choice.

## Verification

`binStackSeating.kernel.test.ts` gains three cases that generate real bins, mate them, and compare the ceiling model to where the solids come to rest, agreeing within 0.05mm. The naive sum is asserted to be outside that band, which is what makes the agreement mean something. Per CLAUDE.md gotchas #14 and #18, asserting the model against `stackedTotalMm` would only prove two copies of the same arithmetic agree.

Also: 12 cases on the ceiling model, 6 on `solveUnitsUnderCeiling` (including that it never returns a height that overflows, and never one unit short of the largest that fits), 5 on the hook, 3 on the plane's mm-to-world conversion, 3 on the summary's states.

Full suite green (20,354 tests), `pnpm quality` clean, all four i18n gates pass with real translations in 14 locales.

## Not included

A drawer-preset library. Each case is already its own layout with its own measurement, so "several drawer heights" is served by several layouts.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3579?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Both Greptile findings addressed in 6f01707.

**Registry erasure — real bug, now fixed.** `upsertRegistryEntry` replaces the whole entry and thirteen call sites write this registry, most carrying only a rename or a fresh thumbnail. `assembledRiseMm` was erased by the next autosave, so every linked bin would have fallen back to plain-bin height almost immediately. My original grep was scoped to `bin-designer` with `--include='*.ts'` and found 5 of the 13.

Fixed structurally rather than by asking the call sites to remember: the upsert carries the geometry-derived fields forward when an update omits them, so a writer that is not touching geometry cannot drop them, and the six writers that do hold `BinParams` spread `registryHeightFields` so theirs stay fresh. Pinned by a test that fails with the carry-forward removed (`expected undefined to be 64.3`).

**Support model — working as intended, now documented.** Footprint overlap rather than layer z-start is deliberate. Layers are a planning abstraction; a printed bin is loose in a drawer and falls until something stops it, so a layer-3 bin with an empty layer 2 beneath it really does rest on the layer-1 bin it overlaps. That is the answer a lid cares about, and it is the conservative direction: the two only diverge downward. A bin is never shorter than its own layer (`useBinInspector` floors it there) and a bin taller than its layer blocks the cells above it, so nothing is ever stranded in mid-air. Added the rationale to the doc comment and two cases pinning it, including the tallest-of-two-supporters fold.

Full suite green at 20,366, `pnpm quality` clean, kernel test still agrees with the mated solids.

PR #3654, opened 2026-08-20
feat(cutout-editor): fill the bin with a repeat in one click

### LINKED ISSUE #3641: Add “Fill Bin” Option for Repeated Cutouts
## Problem

The Repeat tool currently lets me specify the number of rows/columns and the X/Y pitch manually. When I am trying to fill an entire bin with the same cutout, I have to keep adjusting the repeat count and pitch until everything fits.

For many designs, what I really want is simply: **fill as much of the available bin area as possible with this repeated cutout.**

## Proposed Solution

Add a **Fill Bin** option to the Repeat tool.

The user could select a cutout, choose **Fill Bin**, and specify the desired spacing/clearance between repeated instances.

For example:

* Spacing X: 2 mm
* Spacing Y: 2 mm
* Pattern: Grid / Staggered

The tool could then automatically calculate the number of rows and columns that fit inside the available bin area.

It would also be useful to allow an edge offset/margin so the first and last instances can maintain a specified distance from the bin walls.

Ideally, the calculated rows, columns, and pitch would still remain editable afterward.

## Alternatives Considered

Continue manually adjusting rows, columns, Pitch X, and Pitch Y until the bin is filled. This works, but requires quite a bit of trial and error when the goal is simply to distribute as many copies as possible within the available space.

andymai:
Closes #3641

> Stacked on #3650 — its base retargets to `main` when that merges. The diff here is the single Fill bin commit.

## Problem

Filling a bin meant nudging counts and pitch until things stopped fitting, then backing off one.

## Change

**Fill bin** writes into the existing `cols`/`rows` rather than becoming a mode of its own, so the result stays editable — backing off a row does not undo the fill, which is what the issue asked for.

Spacing is whatever the **pitch** fields already say and the master's position is the **leading margin**, so both are set the way the user already sets them rather than through a second set of controls meaning the same thing. The button disables itself once the bin is full, and is not offered for a radial ring, which has no rows or columns to fill.

## The parts that are easy to get wrong

The "how many fit" arithmetic moves into one `spatialCounts`, shared with `arrayFieldBounds`, so the fill cannot disagree with the bounds that will clamp it — a fill that visibly undid part of itself would be worse than no button.

**Rows are counted first, and columns against the rows the fill will actually produce.** A staggered array only pays its half-pitch X offset once it has a second row, so counting columns against the single row it starts on fits one column too many and hangs it off the bin's edge. At a 21mm pitch in a 100mm bin that is an instance reaching 104.5mm; the test for it fails if the ordering is undone.

The instance cap is spent on columns first, so a fill that cannot have everything degrades to complete rows rather than a ragged final one.

## Verification

Unit tests on `fillBinCounts` (far-edge fill, master-as-margin, the stagger ordering, the instance cap, the nothing-fits case, and agreement with `arrayFieldBounds`) and on the button (patch contents, config preserved, disabled when full, absent for radial).

In the running app on a 3×2 bin: a 2×2 staggered repeat of 10mm circles filled to 7×8 = 56 instances reaching both far edges, and the button disabled itself afterwards.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3654?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/bin-designer/components/panel/CutoutsSection/CutoutArrayControls.test.tsx
+++ b/src/features/bin-designer/components/panel/CutoutsSection/CutoutArrayControls.test.tsx
@@ -133,3 +133,44 @@ describe('CutoutArrayControls — overlap', () => {
     expect(screen.queryByText(WARNING)).not.toBeInTheDocument();
   });
 });
+
+describe('CutoutArrayControls — fill bin', () => {
+  const FILL = 'binDesigner.cutouts.repeat.fillBin';
+
+  it('sets rows and columns to what the bin holds at the current pitch', () => {
+    const onUpdate = vi.fn();
+    // 10mm master at 0,0 in a 300mm bin at a 12mm pitch: 1 + floor(290/12) = 25
+    // per axis. 25 x 25 is over the 400-instance cap, which columns win — so
+    // rows take the 16 that are left and every row is complete.
+    renderControls(makeCutout({ array: { ...arrayCfg, cols: 3, rows: 2 } }), { onUpdate });
+
+    fireEvent.click(screen.getByText(FILL));
+
+    expect(onUpdate).toHaveBeenCalledWith({
+      array: expect.objectContaining({ cols: 25, rows: 16, pitchX: 12, pitchY: 12 }),
+    });
+  });
+
+  it('leaves the rest of the config alone, so the result stays editable', () => {
+    const onUpdate = vi.fn();
+    renderControls(makeCutout({ array: { ...arrayCfg, mode: 'staggered' } }), { onUpdate });
+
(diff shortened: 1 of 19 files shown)


andymai:
Superseded by #3656 — GitHub auto-closed this when its stacked base branch was deleted on merge. Same commit, rebased onto main.

PR #3688, opened 2026-08-21
fix(baseplate): report the plate the drawer shape is cut to, and let the grid move inside it

### LINKED ISSUE #3683: Baseplate: Incorrect dimensions shown when sync with layout
**Describe the bug**
The shown baseplate dimensions are not correct on Sync with Layout when using less Cells. Padding is applied but not shown in the baseplate editor. This also removes the option to custom position your grid based on the padding.
**To Reproduce**
Steps to reproduce:

1. Import [untitled-layout (1).json](https://github.com/user-attachments/files/31313572/untitled-layout.1.json)
2. Click on Baseplate
3. See error

**Expected behavior**
Padding should be shown in the baseplate editor. Also the dimensions under the sync with layout tab should represent the models dimensions not just the bare cells. Padding should be adjustable so you can for example align bottom left. The Layout should however clip the model so your model cant grow past your Drawer dimensions.

**Screenshots**

Image

**Environment:**

- Browser: Chrome 151.0.7922.170
- Device: MacOS 26.6.2

andymai:
Closes #3683.

## What was wrong

The reporter's drawer is 931 × 327mm measured, carrying a 21 × 7 grid (882 × 294mm). Their pen outline is therefore *larger* than the socket lattice it hosts.

`resolveOutlineFrame` already handles that inversion: when the perimeter is oversize it centres the lattice inside the shape and reports an overhang (24.5mm left/right, 16.5mm front/back here). The generator widens its slab by exactly that (`slabW = totalW + ohLeft + ohRight`) and intersects it with the perimeter, so the printed plate really is 931 × 327mm.

Two readouts never learned about that step and added up cells and padding instead:

- the panel's mm summary (`gridWidthMm + paddingLeft + paddingRight`) — the `882 × 294 mm` in the issue screenshot
- the 3D dimension annotations, whose leader lines also stopped short of the plate's real edges

With padding at 0, nothing on screen accounted for the solid margin visible on the plate.

## The fix

Both surfaces now derive the footprint from `@/shared/utils/outlineFrame`, the module both sides already share for the frame translation — the plate's material is the slab ∩ the perimeter, and the slab is widened precisely so it contains the perimeter, so the perimeter's own bounds *are* the footprint. That covers the undersized direction too (a shape that falls short of its grid), which previously over-reported.

A trailing note names whatever put the figure past the bare cells: `incl. padding`, `incl. drawer shape`, or both.

Because the readout seeds the click-to-edit field, `EditableDimensions` no longer commits values the user never typed. Opening it and clicking away used to run the caller's snap-to-grid, which on a plate synced to a shaped drawer dropped both the sync and the shape.

## Verification

- New unit coverage at three levels: `drawerFrameExtent` (oversize / undersized / padded), the panel readout with the reporter's exact layout, and the 3D annotations. Reverting just the readout binding reproduces `882 × 294 mm` in the panel test.
- Driven in the real app against the reporter's layout: the panel reads `931 × 327 mm incl. drawer shape`, and probing the live Three scene shows the annotations at `931mm` / `327mm` (previously `882mm` / `294mm`).
- Full unit + dom suites pass (20,854 tests); lint, typecheck, boundaries, design-system, exhaustiveness and all four i18n checks clean. Translations added for all 14 non-English locales.

## Out of scope

The issue also asks for control over where the grid sits inside the shape ("align bottom left"). That already exists as the grid alignment controls on the Layout tab (`drawer.gridShiftX/Y`), which this fix feeds through the same frame. Surfacing them inside the baseplate editor is a separate enhancement.

https://claude.ai/code/session_01PKjJbCN6AycciP9oQxFZaV

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3688?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Self-review turned up a third surface with the same defect, now fixed in 26a2b55.

**Print guide piece size.** `printGuide.ts` documents its `Size:` line as matching "the actual STL bbox so users know what fits the bed", but computed it as `width * gridUnitMm + padding + tongue`. The split planner already propagates `outlineOverhang` per piece (outer sides only) and the generator widens each piece's slab to match, so on the reporter's plate the two outermost pieces are each 24.5mm wider than the guide stated — in the file that ships inside the export ZIP.

While in those two lines: the depth term used `params.gridUnitMm`, the *width*-axis pitch. On a non-square grid that doubled every stated piece depth (a 2 × 2 plate with a 21mm Y pitch reported `84.0 × 84.0` for a part that is `84.0 × 42.0`). Both are covered by new tests that reproduce against the unfixed code.

**Audited and deliberately left alone:**

- `baseplateDirectMesh` — outline-unaware, but `useBaseplateGeneration` bails out of the procedural draft entirely for shaped plates, so it never renders one.
- Export filename / assembly map — both work in grid units, no mm claim.
- Split planner and bed-fit — already overhang-aware.
- `maxCornerRadiusMm(totalWidthMm, totalDepthMm)` — gated on `!outlineActive`, so the padded extent is the correct basis. This is why `totalWidthMm` keeps its old meaning and the footprint is a separate field rather than a redefinition.
- `GhostPaddingOutline` — a transient amber box shown only while regenerating *and* only when padding > 0. Drawing a rectangle around a shaped plate isn't clearly better than the padded rect it draws today.
- Camera framing (`calculateIdealDistance`) — under-frames by the overhang, ~5% here, absorbed by `FRAME_FILL = 0.65`.
- Tessellation tolerance sizing — under-measures `maxDimension`, which errs toward *finer* tolerance.

The STL's 80-byte header still embeds a grid+padding mm string, unchanged from before; it is not surfaced anywhere and touching it would churn the export path for no visible gain.

andymai:
Folded in the enhancement half of #3683 (d0df74b): **"Padding should be adjustable so you can for example align bottom left."**

Padding can't express this. On a synced shaped plate the outer size comes from the drawer, so the slack between the shape and the cells is the only thing left to distribute — and moving the grid is what distributes it. That control already existed on the Layout tab (`drawer.gridShiftX/Y`), so rather than build a second mechanism I moved it to `shared/components` and surfaced it in the baseplate panel under **Grid position**.

Gated on the outline reaching the plate, not on slack already existing: a shape that exactly fills its extent has none yet, and gating it away there would make the first nudge unreachable.

Driven end to end on the reporter's layout. Pushing the grid to the front-left moves the frame from centred to corner-anchored while the footprint stays put:

| | left | right | front | back |
|---|---|---|---|---|
| default (auto-centred) | 24.5 | 24.5 | 16.5 | 16.5 |
| after `−21` / `−16.5` | 3.5 | 45.5 | **0** | 33 |

Readout stays `931 × 327 mm` throughout, and the 3D annotations track the shifted material (the `931mm` label moves from x=0 to x=21) — so the dimension fix and the grid shift compose correctly.

**One honest limitation.** Front alignment lands exactly on 0; left stops 3.5mm short. The manual shift is clamped to ±half pitch (±21mm) and reaching this drawer's left edge needs 24.5mm. The clamp's rationale — "a shift beyond ±half pitch is equivalent to a different whole-cell registration" — holds when the lattice fills its extent, but not when the drawer has more than a full cell of slack (931/42 = 22.17, so 49mm of slack against a 42mm pitch). Widening it would change frame geometry shared by the layout's placement gating and the plate, so I've left it alone rather than fold a geometry change into this PR. The in-panel hint already reports the residual (`−3.5, +0 mm`), so the offset is visible rather than silent.

Also fixed along the way: `BaseplatePanel.test.tsx` stubbed the entire `@/core/store` barrel to mock one export, so any barrel-importing child would have read empty state while the panel rendered a shaped drawer. It now backs `useLayoutStore` with the same fake state as the sibling mock.

andymai:
Clamp fixed (e3e2ad0) — corner alignment is now actually reachable.

**The reasoning behind the old bound, and where it stops holding.** ±half pitch spans every distinct grid position *while the lattice fills its extent*: past that you've just relabelled which cell is which. That stops being true once the perimeter is oversize — the lattice then has real room to slide inside the material, and this drawer has 49mm of slack against a 42mm pitch, so half a pitch physically could not reach the edge.

Half the slack is exactly the reach from the centred registration to either edge, so the bound is now `max(pitch/2, slack/2)`, capped at the static bound the `drawer.update` payload schema already enforces. **Every non-oversize shape keeps its existing range** (slack is 0, so the max is unchanged), and out-of-range stored values still clamp.

There were three clamps, not one, and they had to agree or the UI would offer a position the command refused to store:

| | before | now |
|---|---|---|
| `resolveOutlineFrame` | `±gridUnitMm / 2` | `±shiftLimit{X,Y}` |
| `drawer.update` command | `±layout.gridUnitMm / 2` | `±drawerFrameShiftLimits(...)`, resolved post-change |
| `GridAlignmentControls` steppers | `±gridUnitMm / 2` | `±drawerFrameShiftLimits(...)` |

All three now read one exported `drawerFrameShiftLimits`. The command resolves it against the post-change width/depth, matching the discipline the size floors above it already use.

Driven on the reporter's layout — the front-left corner is now exact on both axes:

| | left | right | front | back |
|---|---|---|---|---|
| auto-centred | 24.5 | 24.5 | 16.5 | 16.5 |
| previous best (`−21`) | 3.5 | 45.5 | 0 | 33 |
| now (`−24.5`) | **0** | 49 | **0** | 33 |

Footprint stays `931 × 327 mm` throughout. The in-panel alignment hint disappears at full alignment, because the residual frame translation really is `(0, 0)`.

Six new tests, all reproducing against the old bound (`expected -21 to be -24.5`, `expected -45.5 to be close to -49`), including one asserting a shape that fills its grid still clamps at `−21`. Full suite 20,866 passing; typecheck, boundaries and doc-drift clean.

**One behavioural note worth stating plainly:** a layout that already stored an out-of-range shift (hand-edited or imported — the UI never produced one) will now clamp to the wider limit instead of half a pitch, so such a plate could render differently. The old clamp treated those as undefined-behaviour values; the new one is the more correct bound for them too.

andymai:
All three review comments were right; fixed in 5af80aa.

**The guard compared raw strings.** Copilot's catch is real and it pokes a hole in exactly the protection the guard exists to provide: re-typing the same measurement in another form (`441.00` over `441`) read as an edit, so the caller's snap-to-grid ran anyway and — on a plate synced to a shaped drawer — still dropped the sync and the shape. Now compared through `formatMm`, which is the precision the field actually displays: anything it rounds away was never on screen to change. Covered by a new test.

**Test name didn't match its assertion.** Renamed to `treats a value edited back to its starting point as no edit` — the old name claimed a commit that the assertion denies.

**Stale schema comments.** Both copies (`updateDrawer.ts` and `validation/schemas.ts`) still described the re-clamp as the layout's ±pitch/2. Updated to name `drawerFrameShiftLimits` and note it is half a pitch until an oversize perimeter gives the lattice more room.

Full suite 20,867 passing; typecheck, lint and format clean.

Issue #3537, opened 2026-08-16
Generator tests run near their timeout, so self-hosted PR shards fail under any load


andymai:
Several generator tests run at roughly 40% of their per-test timeout, so the self-hosted PR shards fail whenever the machine is doing anything else. The failing test differs run to run, which is what a timing margin looks like rather than a defect.

Observed on #3529 across four CI attempts on run `31936519601`. Each time the shard failed with `Test timed out`, and each time on a different set:

| attempt       | shard | test                                                                                                                       | cap       |
| ------------- | ----- | -------------------------------------------------------------------------------------------------------------------------- | --------- |
| 1             | 2/2   | `exportIntegrityRunner.ts:220` (x2), `labelPlateBuilder` › leaves wider plates identical regardless of the v1 flag         | 60s / 30s |
| 2             | 1/2   | `baseplateGenerator.scenario.overtile` › over-tile half-grid keeps a solid leftover consistent across pockets and magnets  | 30s       |
| 2             | 2/2   | `labelPlateBuilder` › leaves wider plates identical regardless of the v1 flag                                              | 30s       |
| 3 (idle pool) | 2/2   | `binGenerator.export.handles` › oval shape handles on front wall, `binGenerator.export.spacer` › 2x2 spacer + half sockets | 60s       |

Local timings for the two that recur, both the slowest in their files:

```
over-tile half-grid keeps a solid leftover consistent across pockets and magnets   11341ms
leaves wider plates identical regardless of the v1 flag                            12489ms
```

Against a 30s cap that is ~40% of budget. A 2.5x slowdown breaches it, and on a shared self-hosted runner 2.5x is not exotic.

**Why the self-hosted path is tighter than the hosted one.** `Unit Tests` on `ubuntu-latest` splits generators **6** ways, and the comment above the matrix explains why — generators carry ~3700s of WASM test CPU, and hash-based sharding means a heavy file lands where it lands. `Unit Tests SH` splits the same project **2** ways at `--maxWorkers=5`, so each shard carries ~1850s of that CPU at concurrency 5. The hosted path has ~3x the headroom for identical tests. Only the self-hosted path is failing.

Attempt 3 is the useful one: it ran with all four `gflt-ci` runners idle, and shard 1/2 went green while shard 2/2 still failed. So contention is a contributor but not the whole story — shard 2/2 has little margin even on a quiet machine.

[...]

PR #3781, opened 2026-08-23
feat(generation): magnet and screw holes on half-size feet

### LINKED ISSUE #3778: Magnet holes for half bin sizes
Was making some hex bit bins for some half grid sizes I need to fill a drawer. I noticed that the magnet holes disappear when going to a half bin size. Is it possible to retain a single magnet hole for the half bin sizes? Thanks. 

Wonderful tool. I'm very much a gridfinity noob and this has been very user friendly for someone just getting started.

andymai:
Closes #3778.

A half foot got no attachment holes at all. A 0.5×0.5 bin came out with none, and a 1.5×1 bin got them only on its full foot, with no UI hint that the magnet toggle had stopped doing anything.

## Cause

Seven independent cell loops each carried the same guard:

```ts
if (cell.widthUnits < 1 || cell.depthUnits < 1) return;
```

It cited the Gridfinity spec not defining fractional-cell magnet positions, but the job it was actually doing was manufacturability, and those diverge: what matters is how much plastic survives around the hole, which depends on pitch and hole radius rather than the unit count. `magnetPositionsForCell`, the shared placement authority the guard stood in front of, already handles a small cell — two holes on its long axis, one centered, or none when nothing fits.

## Change

One predicate replaces all seven:

```ts
export function cellHostsAttachmentHoles(cell, holeRadius, pitchX, pitchY): boolean {
  if (cell.widthUnits >= 1 && cell.depthUnits >= 1) return true;
  const minSpanMm = 2 * (holeRadius + MAGNET_EDGE_CLEARANCE + INSET_BOT) + CLEARANCE;
  return cell.widthUnits * pitchX >= minSpanMm && cell.depthUnits * pitchY >= minSpanMm;
}
```

It asks what the guard meant to ask: does the tapered face the hole opens on still hold a printable wall? A 21mm foot at spec pitch keeps 4.3mm, more than the 1.8mm a standard 42mm cell keeps at its own bottom. A 12.5mm foot on a custom 25mm grid keeps 0.35mm, and is still skipped.

Full cells answer true unconditionally. That asymmetry is deliberate: they carry holes today at every pitch, so the change stays purely additive and no existing design can lose a hole.

Bin base, lightweight base, lid, baseplate, baseplate screws, floor-pattern keep-outs and the draft mesh all move together, so a half foot and the half pocket under it place holes in the same spot.

## Two things that are not the guard swap

- **Draft mesh double-drill.** `baseplateDirectMesh` packs nominal cells and over-tile frame tiles into one array and used `widthUnits < 1` to tell them apart, since frame tiles are sub-unit. Loosening the guard would have magnetized every frame tile twice — once in the nominal pass, once in the over-tile pass. Now split by index.
- **Stranded magnets.** `lightweightFloorCutter` removed a half cell's entire floor on the grounds that it held no magnet. That branch is now gated on the same predicate, so a half cell that takes a magnet keeps its floor.

## Placement choice

A half foot's centered hole matches what this tool's own baseplate puts in a matching half pocket. Against a third-party full-cell plate it lands ~2.5mm off that plate's lattice magnet — partial engagement, where the foot previously offered none. Centered is also the only rotation-invariant choice for a lone half cell, which has no adjacent full cell to take its lattice reference from.

`MESH_CACHE_VERSION` bumped to `v13`: identical params now generate different bytes for any fractional-dimension part with attachment holes on.

## Testing

- `cellHostsAttachmentHoles` unit tests: admits/rejects across pitch and hole radius, full cells unconditional.
- Real-WASM geometry: 0.5×0.5 bin gets a centered hole (with a flat-base negative control), 1.5×1 magnetizes its half foot at the standard ±13 Y spacing.
- Updated the two tests that asserted the old behavior, each keeping a case that still exercises the skip at a pitch where the half cell genuinely cannot hold a hole.
- Full local suite green: unit (11998), dom (9709), generators (3161). Seven generator scenarios timed out on the first pass under CPU contention and pass idle; all are integer-dimension or magnet-free scenarios this change cannot reach.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3781?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/generation/worker/generators/baseplateDirectMesh.ts
+++ b/src/features/generation/worker/generators/baseplateDirectMesh.ts
@@ -55,7 +55,7 @@ import { addPocketWalls, addOuterWalls } from './directMeshWalls';
 import { addPlateFace, addSolidBottomFace } from './directMeshFaces';
 import { addMagnetHoleAt } from './directMeshMagnets';
 import { addScrewHoleAt } from './directMeshScrews';
-import { magnetPositionsForCell } from './baseplateMagnets';
+import { cellHostsAttachmentHoles, magnetPositionsForCell } from './baseplateMagnets';
 import { planBaseplateScrewHoles } from './baseplateScrews';
 import { addConnectorNub, addConnectorHole } from './directMeshConnectors';
 
@@ -140,6 +140,10 @@ export function generateBaseplateDirect(
   // ring, while a sub-threshold sliver keeps its solid fill.
   const cells: CellInfo[] = [];
   forEachCell(width, depth, (cell) => cells.push(cell), cellOpts);
+  // Everything appended past here is an over-tile frame tile. The magnet pass
+  // below magnetizes frame tiles separately, so it needs the nominal grid alone
+  // — it can no longer tell the two apart by cell size.
+  const nominalCellCount = cells.length;
   // Plain solid padding fills the whole margin ring (no pockets); over-tile
   // restricts the ring to the un-pocketed outer band via per-side pocket depths.
   let drawRing = !overTile;
@@ -252,12 +256,10 @@ export function generateBaseplateDirect(
 
   if (magnetHoles) {
     const magnetRadius = magnetDiameter / 2;
-    // Nominal grid: full cells get the standard 4 corners (fractional nominal
-    // cells are skipped, matching the BREP build). `cells` also holds the frame
-    // tiles, but they're <1u so the guard skips them here — they're magnetized
-    // by the over-tile pass below.
-    for (const cell of cells) {
-      if (cell.widthUnits < 1 || cell.depthUnits < 1) continue;
+    // Nominal grid: full cells get the standard 4 corners, half cells whatever
+    // of that pattern their tapered floor holds — matching the BREP build.
+    for (const cell of cells.slice(0, nominalCellCount)) {
+      if (!cellHostsAttachmentHoles(cell, magnetRadius, gridUnitMm, gridUnitMmY)) continue;
       // Shared placement (wall-distance clamp) — identical to the BREP plate,
       // bin base, and lid so the draft preview and all mating surfaces agree.
       for (const [x, y] of magnetPositionsForCell(
--- a/src/features/generation/worker/generators/baseplateMagnets.test.ts
+++ b/src/features/generation/worker/generators/baseplateMagnets.test.ts
@@ -1,7 +1,11 @@
 import { describe, it, expect } from 'vitest';
-import { magnetPositionsForCell, MAGNET_EDGE_CLEARANCE } from './baseplateMagnets';
+import {
+  cellHostsAttachmentHoles,
+  magnetPositionsForCell,
+  MAGNET_EDGE_CLEARANCE,
+} from './baseplateMagnets';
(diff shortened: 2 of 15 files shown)


andymai:
Pushed 1481d1ce9 addressing all four review points. Thanks — the P1 was real and led somewhere I had not looked.

## Lid pockets vs magnets (Greptile P1) — real, and worse than reported

Greptile flagged that `lidMagnets` and `lidStackGrid` could disagree about a fractional masked cell. Digging in, the stack-grid side was already broken on its own:

```
half cell of a 1.5u lid → centerX 21
old: Math.round((21 + 31.5 - 21) / 42) = 1  → isCellFilled checks mask cols 2 and 3
     mask.cols = 3                          → col 3 out of bounds → false
```

So on a masked fractional lid, the trailing half cell got **no pocket at all**, always — the whole-cell query reached past the mask's own last column. Consistent-by-accident before, since the magnet pass skipped half cells too; my change would have made it visible as a magnet with no pocket under it.

Both passes now share one predicate over the cell's **own** extent (`isLidCellFilled`), which is bit-identical to the old query for full cells and correct for half ones. Pinned in `lidStackGrid.test.ts`; the regression case fails against the old index math and passes against the new.

While there: `buildStackGrid` never received `fractionalEdgeX/Y`, so it laid the half cell on the default side while the bin's feet and the lid's magnets both followed the setting. On a `'start'` bin the foot met slab and the pocket met air. `LidInputs.fractionalEdgeX` documents itself as existing so the lid lines up with the feet — the pocket pass just never read it.

## Derivable comments (Greptile P2)

Fair. Trimmed the predicate's docblock to the two things the code cannot state: that the nominal edge is not the face the hole opens on, and that full cells *bypass* the span check rather than passing it (a compatibility decision, not arithmetic). Same pass over the others I added.

## Import placement (Copilot)

Folded the mid-file `import type` into the top `./generatorTypes` block — that was pre-existing, but it was in the file I was already editing.

## `Math.min(...spread)` (Copilot)

Legitimate. Replaced with a single pass over `mesh.vertices`, dropping both the spread and the intermediate array — and `hasHoleAt` calls this per assertion, so it mattered more after my refactor than before.

## Test status

Full local suite green: unit 11998, dom 9709, generators 3171 (+ the 4 lid-predicate tests).

Two generator scenarios fail on my machine under full-project parallelism — `2×2 underside + drainage holes`, scenario and export. They are **not** from this change: the scenario is a `style: 'standard'` bin, so `withMagnet` and `withScrew` are both false and every path I touched is gated off. Confirmed against a clean `origin/main` worktree, which fails the same two at 30720ms and 31293ms against the 30000ms `defineScenario` budget with an identical triangle count (27340). A knife-edge budget, same shape as the note in `vitest.config.ts` about #3537. Passes in isolation. Left alone rather than papered over here.

PR #3794, opened 2026-08-23
fix(tests): scenario cases inherit the project timeout instead of pinning 30s

### LINKED ISSUE #3792: Flaky: "2×2 underside + drainage holes" exceeds its 30s cap — defineScenario overrides the raised project timeout
`lightweight › 2×2 underside + drainage holes` fails on a full local `--project generators` run, in both `binGenerator.scenario.lightweight.test.ts` and `binGenerator.export.lightweight.test.ts`. It is a timing margin, not a defect: the triangle count is identical (27340) on every run, and the scenario report prints `PASS` on its own assertions — vitest fails it afterward on elapsed time.

Surfaced while validating #3781. Confirmed **not** caused by that PR: reproduced on a clean `origin/main` worktree at `71fc6656b`, and the scenario is a `style: 'standard'` bin, so it touches none of the paths that PR changed.

## Measurements

All on the same machine, same scenario, `timeout: 30000`:

| how it was run | elapsed | verdict |
| --- | --- | --- |
| single file, 27 scenarios ahead of it in the worker | **16785ms** | pass (56% of budget) |
| `-t drainage` on clean `origin/main` — only these 2 tests, machine idle | **30720ms / 31293ms** | fail |
| full `--project generators`, machine idle | **33625ms / 33180ms** | fail |
| full `--project generators`, under CPU contention | **69924ms / 72398ms** | fail |

The first row is the interesting one. Warm — 27 scenarios have already run in that worker — it lands at 16.8s. Cold, with the whole machine to itself and only two tests running, the same scenario takes 30.7s. Warm-up is worth roughly 2x here, so what the budget measures is mostly which tests happened to run first.

## Root cause: the #3537 fix cannot reach scenario cases

#3537 diagnosed exactly this band ("roughly 40% of their per-test timeout") and was closed by raising the generators project to `testTimeout: 120_000` in `vitest.config.ts:164`, with a comment describing 120s as "the tier the deliberately-annotated heavy files already use".

That raise cannot reach any scenario case. `defineScenario` hard-codes a default:

```ts
// scenarioTypes.ts:67
timeout: 30_000,
```

and `runScenarios` passes it to `it()` as an explicit third argument:

```ts
// scenarioRunner.ts:113-119
it(scenario.name, () => { runScenario(scenario, timings); }, scenario.timeout)
```

An explicit per-test timeout overrides the project default, so every scenario case is pinned at 30s no matter what the project config says. **270 of 452** scenario cases sit on that default; the other 182 carry an explicit raise. `lightweight.ts` has two explicit raises (`4×4 lite export (stress)` and `2×2 lite + 2×2 compartments + magnet`, both 60s) — the drainage case is not one of them, despite being the slowest in the file.

## Why CI has not caught it

Hosted `Unit Tests` shards generators 6 ways; self-hosted `Unit Tests SH` shards the same project 2 ways. Both shards passed on #3781. This is currently latent in CI and biting local full-project runs — but it is the same margin band #3537 was filed about, so the self-hosted path is where it would show up first.

## Fix

`defineScenario` leaves `timeout` unset, so `it(name, fn, undefined)` falls through to the project's 120s and the ceiling lives in one place (`vitest.config.ts`). `ScenarioCase.timeout` becomes optional and now means only "this case needs more than the ceiling".

That turns an explicit value into a plain override in **both** directions, which exposed a second problem: 152 cases annotated 60s or 90s would have ended up with *less* budget than an unannotated one — and those are the cases someone had flagged as slow. So every explicit value at or under the ceiling is removed (125 at 60s, 27 at 90s, 14 restating 120s). The 16 kumiko cases at 180s stay; they genuinely need more than the ceiling. No scenario's budget drops.

The 120s ceiling itself is unchanged — it is what #3537 chose, and it is 1.7x the worst figure observed above.

## Not the fix: shard or runner count

Recorded here because it is the obvious wrong turn. `ci.yml` documents that the self-hosted pool was briefly four runners and was deliberately reverted to two, under the invariant `gflt-ci runners x maxWorkers <= 10 performance cores`. A 4x3 trial (run `31960552340`) made the slowest shard *worse*, 250s -> 355s, from disk contention between concurrent installs and poorer hash balance at 4-way. The comment is explicit: "the fix is NOT more shards" and "Registering a third runner silently breaks that product and reopens #3537."

`--shard` splits the file set, so it shortens a job without making any single test faster — it cannot help a per-test cap.

## No new slow-scenario signal

Raising 436 cases to 120s does weaken the signal for a genuine slowdown. Deliberately not addressed: the scenario report already prints per-case ms, and `vitest.config.ts` states this cap is "a liveness bound, not a performance budget". Adding a budget back is the thing being removed.

## Repro

```bash
pnpm run test:run --project generators -t drainage
```

Passes in isolation:

```bash
pnpm run test:run src/features/generation/worker/generators/binGenerator.scenario.lightweight.test.ts
```

andymai:
Closes #3792.

`lightweight › 2×2 underside + drainage holes` fails on a full local `--project generators` run, in both the scenario and export files. It is a timing margin, not a defect — identical triangle count (27340) every run, and the scenario's own assertions pass; vitest fails it afterward on elapsed time.

## Cause

#3537 diagnosed this exact band and was closed by raising the generators project to `testTimeout: 120_000`. That raise could not reach a single scenario case. `defineScenario` set a default:

```ts
timeout: 30_000,
```

and the runner passes it to `it()` as an explicit third argument:

```ts
it(scenario.name, () => { runScenario(scenario, timings) }, scenario.timeout)
```

An explicit per-test timeout overrides the project default, so **270 of 452** cases stayed on 30s no matter what the config said.

## Why 30s was never a stable line

Same scenario, same machine, `timeout: 30000`:

| how it ran | elapsed | verdict |
| --- | --- | --- |
| single file, 27 scenarios ahead of it in the worker | **16785ms** | pass (56% of budget) |
| `-t drainage` on clean `origin/main` — 2 tests, machine idle | **30720ms / 31293ms** | fail |
| full `--project generators`, idle | **33625ms / 33180ms** | fail |
| full `--project generators`, under contention | **69924ms / 72398ms** | fail |

Warm it lands at 16.8s; cold, with the whole machine to itself, the same scenario takes 30.7s. Warm-up is worth roughly 2x, so what the cap measured was mostly which scenarios happened to run first in the worker.

## Change

**Commit 1** — leave the key unset so `it(name, fn, undefined)` falls through to the project ceiling. `ScenarioCase.timeout` becomes optional and now means only "this case needs more than the ceiling". The value lives in one place.

**Commit 2** — that turns an explicit value into a plain override in *both* directions, which exposed a second problem: 152 cases annotated 60s or 90s would have ended up with **less** budget than an unannotated one, and those are the cases someone had flagged as slow. So every explicit value at or under the ceiling is removed — 125 at 60s, 27 at 90s, 14 restating 120s. The 16 kumiko cases at 180s stay; they genuinely need more than the ceiling.

No scenario's budget drops. None of the removed lines carried a comment. The 120s ceiling is unchanged — it is what #3537 chose, and 1.7x the worst figure above.

## Not the fix: shard or runner count

Noted because it is the obvious wrong turn, and #3537 reads as though it were untried. `ci.yml` documents the self-hosted pool briefly running four runners and being deliberately reverted to two, under `gflt-ci runners × maxWorkers ≤ 10 performance cores`. A 4×3 trial (run `31960552340`) made the slowest shard *worse*, 250s → 355s. `--shard` splits the file set, so it shortens a job without making any single test faster — it cannot help a per-test cap.

## Deliberately not added

Raising 436 cases to 120s does weaken the signal for a genuine slowdown. The scenario report already prints per-case ms, and `vitest.config.ts` states this cap is "a liveness bound, not a performance budget" — adding a budget back is the thing being removed.

## Verification

The exact repro that failed on clean `main` now passes:

```
$ pnpm run test:run --project generators -t drainage
Test Files  2 passed | 323 skipped (325)
```

That is the load-bearing check: the scenario needs more than 30s, so it only passes if the inherit actually reaches `it()`.

Commit 2 is mechanical — 166 deletions across 27 files, zero insertions, and every changed line is a `timeout:` deletion (verified by diffing with `-U0` and matching every `[+-]` line against the expected pattern).

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3794?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/generation/worker/generators/__kernel-tests__/scenarioTypes.ts
+++ b/src/features/generation/worker/generators/__kernel-tests__/scenarioTypes.ts
@@ -40,7 +40,15 @@ export interface ScenarioCase {
   params: Partial<BinParams>;
   forExport?: boolean;
   assert: AssertionMode;
-  timeout: number;
+  /**
+   * Per-case override of the project's `testTimeout`. Absent means inherit it.
+   *
+   * The runner passes this straight to `it()`, where an explicit value wins in
+   * BOTH directions — a number below the project ceiling LOWERS this case's
+   * budget rather than raising it. Set it only for a case that needs more than
+   * the ceiling.
+   */
+  timeout?: number;
   /** Runs after standard assertions with the generated mesh. */
   customAssert?: (result: MeshData, params: BinParams) => void;
   /** Generate a second mesh and compare. */
@@ -52,7 +60,7 @@ export interface ScenarioCase {
 /**
  * Create a scenario case with sensible defaults.
  * - `assert` defaults to `'snapshot'`
- * - `timeout` defaults to `30000`
+ * - `timeout` is left unset, so the case inherits the project's `testTimeout`
  */
 export function defineScenario(
   category: string,
@@ -64,7 +72,6 @@ export function defineScenario(
     name,
     category,
     assert: 'snapshot',
-    timeout: 30_000,
     ...config,
   };
 }
--- a/src/features/generation/worker/generators/scenarios/baseStyles.ts
+++ b/src/features/generation/worker/generators/scenarios/baseStyles.ts
@@ -42,7 +42,6 @@ export const baseStyles: ScenarioCase[] = [
         depth: 4,
         base: { ...DEFAULT_BIN_PARAMS.base, style },
       },
-      timeout: 60_000,
       customAssert: (result, params) => {
         assertBoundingBoxMatchesParams(result, params, `4x4-${label}`);
       },
--- a/src/features/generation/worker/generators/scenarios/binStyles.ts
+++ b/src/features/generation/worker/generators/scenarios/binStyles.ts
@@ -39,7 +39,6 @@ export const binStyles: ScenarioCase[] = [
         style,
         base: { ...DEFAULT_BIN_PARAMS.base, ...base },
       },
-      timeout: 60_000,
       customAssert: (result, params) => {
         assertBoundingBoxMatchesParams(result, params, `4x4-${label}`);
       },
--- a/src/features/generation/worker/generators/scenarios/combinedFeatures.ts
+++ b/src/features/generation/worker/generators/scenarios/combinedFeatures.ts
@@ -15,7 +15,6 @@ export const combinedFeatures: ScenarioCase[] = [
       compartments: { cols: 2, rows: 2, cells: [0, 1, 2, 3], thickness: 0.8 },
       scoop: { enabled: true, radius: 'auto' },
     },
-    timeout: 60_000,
   }),
   defineScenario('combined features', '4\u00d74 magnet + label bracket + half-sockets', {
     params: {
@@ -29,7 +28,6 @@ export const combinedFeatures: ScenarioCase[] = [
       },
       label: { ...DEFAULT_BIN_PARAMS.label, enabled: true, support: 'bracket' },
     },
-    timeout: 60_000,
   }),
   defineScenario('combined features', '1.5\u00d72 flat + slotted + 3\u00d72 merged compartments', {
     params: {
@@ -39,7 +37,6 @@ export const combinedFeatures: ScenarioCase[] = [
       base: { ...DEFAULT_BIN_PARAMS.base, style: 'flat' },
       compartments: { cols: 3, rows: 2, cells: [0, 0, 1, 2, 3, 4], thickness: 0.8 },
     },
-    timeout: 60_000,
   }),
   defineScenario('combined features', '2\u00d72 compartments + insert (overlap interaction)', {
     params: {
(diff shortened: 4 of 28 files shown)


andymai:
Closing as a duplicate. #3793 landed the identical fix ~20 minutes before this one, from a parallel session, and closed #3792.

Verified `main` already carries the whole change: the 30s default is gone from `defineScenario`, all 166 explicit timeouts at or under the 120s ceiling are stripped (0 remain), and the 16 genuine 180s raises are preserved.

The only delta was doc-comment wording, and #3793's version is cleaner — it also dropped `'timeout'` from the `Omit`/`Partial<Pick>` signature, which is redundant once the field is optional, and it explicitly confirmed `exportIntegrityRunner` reads the same field. Nothing here is missing from `main`.

PR #3809, opened 2026-08-24
fix(designer): ghost outline follows the cutout's true shape and cut

### LINKED ISSUE #3797: Cutout editor bounding box doesn't update when changing lean or corner radius
In the cutout editor, the box highlighting the edges does not update when changing lean or corner radius. The correct behavior is for the yellow bounding box to represent the cutout's true shape and position.

Image

Image

andymai:
## Summary

The yellow ghost in the 3D preview drew every non-circle cutout as a sharp vertical box, so it stopped matching the real cut as soon as the shape had any character:

- **Corner radius / slot ends / polygon sides** were ignored — a pill-shaped slot got a sharp rectangle.
- **Lean** kept the ghost upright while the actual pocket tilts about its opening.
- **Repeats** drew only the master, though the worker cuts every instance (the knife ghost already expands arrays).
- **Fill top offset** was ignored, so with a lowered fill surface the ghost floated at the rim.

The outline is now built from the tool the worker subtracts. Each shape samples its real footprint (shared `cutoutPolygon` helpers + flattened paths, in the same box-center frame the worker builds tools in), and the lean is applied exactly as `applyCutoutLean` does — rotate about the mouth — which gives:

- the opening ring in the surface plane, stretched by `1/cos(lean)` along the tilt axis (what the cut opening really measures),
- the floor ring shifted `cutDepth·sin(lean)` along the tilt and itself tilted out of horizontal,
- side edges running along the tool's slanted walls.

Plan rotation composes on top at `-rotation`, matching the renderers and the worker. Repeat masters expand through `expandCutoutArray`, and the bin host passes `wallHeight − cutoutConfig.topOffset` as the surface (skipping entirely when no fill remains, as `buildCutoutCuts` does). The lid ghost zeroes lean the same way the lid builder's `through()` does, since lid holes bore straight through.

Insertion clearance and the entry chamfer stay undrawn, matching the editor-wide nominal-size policy documented on `cutoutToPolygon`.

## Testing

- New `ghostCutoutGeometry` unit tests assert the emitted positions directly: rounded corners and slot ends replace the sharp box, the 45° lean's opening stretch / floor shift / floor tilt match the closed-form values, rotation carries the lean direction, non-tilting shapes ignore `leanDeg`, repeats expand, and degenerate cutouts are skipped like the worker skips them.
- Full bin-designer suite (5416 tests) green.

Fixes #3797

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3809?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/bin-designer/components/preview/GhostCutouts/GhostCutouts.tsx
+++ b/src/features/bin-designer/components/preview/GhostCutouts/GhostCutouts.tsx
@@ -113,10 +113,12 @@ export function GhostCutouts() {
     } else {
       return [];
     }
-    if (previewOverrides.size === 0) return result;
     return result.map((c) => {
       const overrides = previewOverrides.get(c.id);
-      return overrides ? { ...c, ...overrides } : c;
+      const merged = overrides ? { ...c, ...overrides } : c;
+      // The worker cuts at most the remaining fill (`effectiveDepth`), so a
+      // deeper stored cutDepth must not draw a floor below the bin's own.
+      return merged.cutDepth > fillSurface ? { ...merged, cutDepth: fillSurface } : merged;
     });
   }, [isSolid, cutouts, isGenerating, hasSelection, selectedIds, previewOverrides, fillSurface]);
 
--- a/src/features/bin-designer/components/preview/GhostCutouts/GhostCutouts.tsx
+++ b/src/features/bin-designer/components/preview/GhostCutouts/GhostCutouts.tsx
@@ -41,6 +41,7 @@ export function GhostCutouts() {
     heightUnitMm,
     wallThickness,
     cutouts,
+    cutoutConfig,
     base,
     lid,
     overhang,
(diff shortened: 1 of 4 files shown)


andymai:

- **Cut depth vs remaining fill**: fixed in 4a786aae0 — the bin host now clamps `cutDepth` to the fill surface, mirroring the worker's `effectiveDepth = min(cutDepth, solidSurfaceZ)`.
- **Malformed paths**: leaving as is. The ghost keeps the previous behavior (draw the flattened outline); pulling the worker's coincident-point/self-intersection validation into the render path buys little for a state the editor already flags.
- **Scoop fillet in the floor ring**: intentionally undrawn, like clearance and the chamfer — the wireframe shows the nominal tool footprint, and modeling the fillet's inset floor would double the worker's scoop resolution logic in the UI layer.

PR #3811, opened 2026-08-24
fix(designer): warn when a cutout's cut depth cannot be fully generated

### LINKED ISSUE #3800: Model generator silently increases/shrinks specified cut depth when using lean
When creating a bin in the cutout editor, there is no warning or other indicator that tells the user their bin will be shallower than expected. This happens when setting a position and lean value that makes rendering the bin at the desired dimensions impossible due to there not being enough horizontal space remaining to make room for the cutout.

In this case, 3 of the 4 bins have the same depth (59.7mm) while 1 bin has a depth of 47.3mm. The cutout dimensions in the web interface are W=125mm, H=11mm, Cut depth=54mm, Lean=45deg. I imagine 59.7mm is necessary to get 54mm of depth at this angle, but I'm no mathematician. I think users should see a prominent warning when there's not enough room to generate all of the bins at the defined dimensions.

Image
Image
Image

### LINKED ISSUE #3802: Cut depth and width can be set to an impossible value in cutout editor
To reproduce:

- In the cutout editor, set the height to some value, say 9u.
- Set the cut depth to its maximum via the slider, say 58mm.
- Reduce the height below 9u by clicking the minus sign.
- Observe unbounded cut depth.

Image

andymai:
## Summary

Two silent "requested vs generated" divergences in the cutout editor, one warning system for both:

- **#3802**: reducing the bin height under a stored cut depth left the inspector showing a depth the worker cannot reach (it clamps to `effectiveDepth = min(cutDepth, fill surface)`), with no indication anywhere.
- **#3800**: a leaned pocket needs more room than its cut depth suggests. The tool tilts about its opening, so the floor slides sideways and down; when it runs into a wall or through the bin's own floor, the interior clip silently shortens the pocket.

## What changed

- New pure helper `cutoutDepthShortfall`: mirrors the worker's tool math (lean about the mouth, plan rotation, interior clip). Each footprint corner's floor position is linear in the depth, so every interior face (floor + four walls) yields a linear bound; the achievable depth is their minimum. A repeat master answers for its worst instance, since the worker cuts them all. A mouth already off the board is left to the off-board warning, and mesh imprints (cut in the mesh domain) are never judged. The footprint box stands in for exact outlines, which is slightly conservative for round shapes: a warning threshold, not geometry.
- **Per-cutout warning** under the depth field in the inspector, with both numbers: "The bin has room for {achievable}mm of this {requested}mm cut…". Live during drags via preview overrides.
- **Aggregate banner** in the bin-size section beside the off-board alert, so shrinking the bin surfaces the problem at the control that caused it: "{count} cutout(s) will be cut shallower than requested".
- The depth ceiling (`maxCutDepth`) now measures from the fill surface (`wallHeight − topOffset`) rather than the wall top, matching what the worker actually cuts; the lid's through-cut plate is unchanged.
- Both strings added to all 14 non-English locales.

The bin-width half of #3802's title is already covered by the existing off-board warning; depth had no equivalent until now.

## Testing

- 10 unit tests on the helper: the shrunken-bin case, lean floor clip and wall clips (both lean signs), rotation carrying the clip direction, worst-instance repeats, off-board and mesh deferrals, and float-noise tolerance — each asserted against the closed-form clip values.
- Banner render test in `BinSizeSection.test.tsx`.
- Full bin-designer suite (5429 tests), typecheck, and all four i18n checks green.

Fixes #3800
Fixes #3802

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3811?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Addressed: the deferral now judges the nominal (un-leaned) footprint corner, the same one the off-board warning measures, so the two warnings cannot both fire for one corner (0442ac506).

PR #3814, opened 2026-08-24
fix(designer): rotate validation silhouettes clockwise like the renderer

### LINKED ISSUE #3805: Cutout placement validation rotates asymmetric shapes backwards (mirrored silhouette)
Found while working on #3797/#3799.

`Cutout.rotation` is clockwise-positive: `CutoutShapeMesh`/`PathShapeMesh` render at `-rotation`, `booleanGeometry.rotatePair` negates it (with a comment explaining exactly this convention), and the worker extrudes tools at `rotate(shape, -rotation)`.

The containment path rotates the other way:

- `src/features/bin-designer/components/panel/CutoutsSection/cutoutOutline.ts` — `rotateRing` calls `rotatePoint(p.x, p.y, cx, cy, cutout.rotation)` with the un-negated angle (`rotatePoint` is CCW-positive).
- `src/features/bin-designer/components/panel/CutoutsSection/maskFit.ts` — `rotatedPathBounds` does the same.

For mirror-symmetric shapes (rectangles, circles, even-sided polygons) the error is invisible, and the existing test rotates a square path, so it never catches direction. For asymmetric silhouettes — bezier paths, mesh-imprint outlines — `cutoutFitsInMask`, `getCutoutBounds`, and everything built on them (off-board detection, `OffBoardFrames3D`, placement validation on masked boards) evaluate a mirror image of the real cut: an L-shaped path rotated 90° can validate at a position where the actual cut overhangs an unfilled cell, or get flagged where it actually fits.

Repro sketch: masked (custom-shape) bin, draw an asymmetric path cutout near a mask notch, rotate it 90° so one arm points into the notch; compare the off-board flag against the generated mesh.

andymai:
## Summary

`Cutout.rotation` is clockwise-positive: `CutoutShapeMesh`/`PathShapeMesh` render at `-rotation`, `booleanGeometry.rotatePair` negates it explicitly, and the worker extrudes tools at `rotate(shape, -rotation)`. The containment path rotated the other way: `cutoutOutline.rotateRing` and `maskFit`'s rotated-path bounds fed the un-negated angle to the CCW `rotatePoint`.

For mirror-symmetric shapes the error is invisible, and the existing tests rotate symmetric shapes (a square path, a plain bar), so direction was never pinned. For asymmetric silhouettes — bezier paths and mesh-imprint outlines — `cutoutFitsInMask`, `getCutoutBounds`, and everything built on them (off-board detection, masked-board placement validation) judged a mirror image of the real cut: an L-shaped path rotated 90° could validate where the actual cut overhangs a mask notch, or get flagged where it fits.

Both sites now rotate through the negated angle, with comments naming the convention. `groupRotateHandler` was audited too and is correct as is (its CCW positional delta pairs with a negative stored-rotation delta).

## Testing

- New direction-pinning tests use asymmetric shapes: a right-triangle path rotated 90° must land its square corner top-left (clockwise), and a rotated asymmetric path's `getCutoutBounds` must match the closed-form clockwise bounds — the CCW mirror produces distinctly different values in both.
- Existing symmetric-shape rotation tests pass unchanged, as they must.
- Full bin-designer suite (5431 tests) and typecheck green.

Fixes #3805

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/bin-designer/components/panel/CutoutsSection/cutoutOutline.test.ts
+++ b/src/features/bin-designer/components/panel/CutoutsSection/cutoutOutline.test.ts
@@ -111,6 +111,22 @@ describe('getCutoutOutline', () => {
     expect(ringArea(rings?.[0] ?? [])).toBeCloseTo(900);
   });
 
+  it('rotates clockwise, matching the renderer and the worker', () => {
+    // A right triangle with its square corner at bottom-left (0,0). Stored
+    // rotation is clockwise-positive (PathShapeMesh renders at -rotation), so
+    // 90° carries that corner to the TOP-LEFT (0,10); the CCW mirror would put
+    // it bottom-right instead, and a mask check made there validates a shape
+    // the cut does not match.
+    const path = [corner(0, 0), corner(10, 0), corner(0, 10)];
+    const ring = getCutoutOutline(makeCutout({ shape: 'path', path, rotation: 90 }))?.[0] ?? [];
+    const has = (x: number, y: number) =>
+      ring.some((p) => Math.abs(p.x - x) < 1e-6 && Math.abs(p.y - y) < 1e-6);
+    expect(has(0, 10)).toBe(true);
+    expect(has(0, 0)).toBe(true);
+    expect(has(10, 10)).toBe(true);
+    expect(has(10, 0)).toBe(false);
+  });
+
   it('rotates a path about its own vertex bounds, not the width/depth box', () => {
     // x/y/width/depth deliberately disagree with the path — the renderer pivots
     // on the vertex bounds, so the outline must too.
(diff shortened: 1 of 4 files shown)


andymai:
Good catch, and verified: `meshImprint.ts:312` rotates the tool by `+cutout.rotation` (CCW in Manifold), while `MeshFootprintMesh` renders at `-rotation` like every other renderer. That is a distinct bug in the worker rather than in this validation change: the editor draws the mesh footprint clockwise-rotated, so validating against the clockwise silhouette (this PR) matches what the user sees and places; the export then cutting it counter-clockwise is a preview-vs-export divergence on the generation side. No existing worker test pins that direction (all mesh-imprint tests use rotation 0). Fixing it separately as fix(generation) with a direction test, immediately after this merges, so the worker joins the same convention instead of this PR aligning validation to a mirrored export.

PR #3818, opened 2026-08-24
fix(generation): cut rotated mesh imprints clockwise like the editor

andymai:
## Summary

Follow-up to #3814, from its review discussion. `Cutout.rotation` is clockwise-positive: `MeshFootprintMesh` renders the mesh footprint at `-rotation` and every BREP cutout tool extrudes at `rotate(shape, -rotation)`. The mesh-imprint path was the one placement rotating the raw angle through Manifold's CCW Z rotation, so a rotated asymmetric imprint (the entire point of shadow-board tools) exported mirrored from the footprint the editor showed and validated.

One-line negate in `placeLocal`, with the convention named in the comment. Radial repeat instances (`rotateToCenter`) flow through the same site, so they pick up the fix too.

## Testing

- New real-WASM direction test: an L-shaped tool (top-right quadrant of its footprint empty) rotated 90° must land the empty quadrant bottom-right — carved where the filled quadrants rotate to, solid where the gap does. Verified the test fails against the previous rotation (mirrored pocket) before the fix.
- No existing test pinned the direction (all mesh-imprint tests used rotation 0); the full imprint suite, normals suite, and the mesh split-step scenario pass.

Follows up on #3805.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3818?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/features/generation/worker/generators/meshImprint.test.ts
+++ b/src/features/generation/worker/generators/meshImprint.test.ts
@@ -264,8 +264,13 @@ describe('mesh imprint generation (occt + manifold)', () => {
       });
 
     // The bottom-left quadrant lands top-left under a clockwise turn: carved.
+    // `lessThan` cannot pass vacuously — an empty probe reads Infinity.
     expect(probe(-5, 5)).toBeLessThan(solidTop - 4);
     // The empty quadrant lands bottom-right: the surface there stays solid.
+    // Infinity (no centroids sampled) is itself a pass: an untouched top face
+    // tessellates as a few large triangles, while a wrong-direction pocket
+    // fills the window with sub-threshold centroids — the reverted-fix run
+    // fails exactly here.
     expect(probe(5, -5)).toBeGreaterThan(solidTop - 0.3);
   }, 120_000);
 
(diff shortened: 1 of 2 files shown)


andymai:
On the Infinity concern: for the carved-quadrant probe, lessThan cannot pass vacuously (an empty probe reads Infinity and fails). For the solid-quadrant probe, Infinity is itself the pass condition rather than a false positive: an untouched top face tessellates as a few large triangles whose centroids may miss the window, while a wrong-direction pocket necessarily fills the window with sub-threshold centroids. That is empirically pinned: running this test against the reverted fix (+rotation) fails on exactly that assertion. Added a comment in the test stating this. The generators-shard failure on the previous run was the unrelated 'direct is at least 10x faster than BREP' timing assertion on the shared runner; rerunning with this push.

PR #3857, opened 2026-08-24
fix(bin-designer): keep merged compartment shapes through move, clone and stash

### LINKED ISSUE #3825: Bug in bento workspace when using the new merge feature
**Describe the bug**
When performing any move action (draging the shape to move it, cloning a shape with alt-drag, moving a shape to the stash) on a group boxes merged using the new merge function in the bento workspace, the shape turns into a rectangle the size of the outer bounds of the shape.

**To Reproduce**
Steps to reproduce:

1. Go to the bento workspace
2. Create 2 boxes and merge them together by shift-clicking
3. Try to move the newly created shape
4. The shape turns into a rectangle

**Expected behavior**
The shape should stay the same shape when moving or cloning it

**Screenshots**

Image

Image

Image

Image

**Environment:**

- Browser: Chrome 151.0.7922.173
- Device: Desktop, AMD 5600x, Win10

andymai:
Merged L/S/T/U compartments in the Bento workspace flattened into their bounding rectangle the moment they were touched.

## Cause

`moveCompartment`, `duplicateCompartment`, `stashCompartment` and `resizeGridPreservingCompartments` each read a compartment through `getCompartmentRect()` and then repainted `rectIndices(target)` — the full rectangle. Correct for every compartment except a merged one, whose cell set is smaller than its bounding box. The canvas already rendered those shapes properly; only the mutations lost them.

The issue reports drag, alt-drag clone and stash. The header grid resize is a fourth door onto the same bug.

## Fix

Every op that rebuilds `cells` now carries the compartment's footprint as offsets inside its bounding box (`compartmentMask`) rather than re-deriving a rect.

- Collision testing follows via `canPlaceMask`, so an L drops into an L-shaped gap its bounding box overlaps — a rect test was refusing legal placements.
- The drag ghost and the stash shelf tile draw the real outline instead of promising a rectangle the drop will not fill.
- `resizeCompartment` refuses a non-rectangular compartment, matching the canvas, which already hides its handles: dragging one edge of an L has no defined result.
- Context-menu duplicate seats copies with `findFreeSpot`, so it can use a gap the bounding box would have been turned away from.

## Persisted change

`StashedCompartment` gains an optional `cells` footprint. Absent means the full box, so every existing design and every older client is unaffected and rectangles serialize byte-identically. Updated in lockstep: the JSON schema, `api/lib/designerCompartmentValidation.ts`, the `binDesignKeys` manifest and `docs/schemas`.

`migrateBentoCompartmentFields` carries the field through on load. It was silently dropping it — a whitelist that rebuilds the entry from `{ w, h, label }` — which no unit test could catch because the model was already correct. A corrupt footprint degrades to its bounding box rather than dropping the entry.

## Verification

Reproduced first: six failing tests across all five affected ops, then fixed. Full suite green (25,104 tests).

Checked in the running app with a real mouse drag: an L seeded at (0,0)(1,0)(0,1) lands at (2,0)(3,0)(2,1) with its shape intact, the drag ghost renders as an L outline, and a stashed L survives a save/reload cycle and comes back out of the shelf as an L with its label.

Closes #3825

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/3857?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.claude/skills/bin-designer/SKILL.md
+++ b/.claude/skills/bin-designer/SKILL.md
@@ -22,6 +22,8 @@ Read `src/features/bin-designer/README.md` first — it covers the store/worker/
 4. **Every client constraint has a hand-maintained server mirror.** `constants/gridfinity.ts` `DESIGNER_CONSTRAINTS` ↔ `api/lib/designerValidationConstants.ts` `CONSTRAINTS`; type unions in `types/index.ts` ↔ `VALID_BIN_STYLES` / `VALID_LABEL_TAB_SUPPORTS` in `api/lib/designerValidation.ts`; top-level param keys ↔ `ALLOWED_PARAM_KEYS` there. Divergence means 400s on sync or params silently stripped from shares. `sanitizeTags` (server) must stay behaviorally identical to `normalizeTags` (`utils/tags.ts`).
 5. **Wall-feature units are mixed by design.** `WallCutout` width/depth are percentages (0–100 of wall span / wall height); `widthMm: null` means "use the % field". Handle width is % of span but handle height is mm. See the JSDoc on each field in `types/index.ts` — mixing these passes typecheck and produces silently wrong geometry.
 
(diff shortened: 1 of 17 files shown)


andymai:
Superseded by #3849, which landed the same fix while this was in review. Closing as a duplicate rather than resolving the conflict.

The one piece not covered there: #3849 deliberately keeps placement bounding-box based, so a merged L is still refused a gap its empty notch overlaps, and the drag ghost still draws a rectangle. Following up with just that.

PR #4162, opened 2026-09-09
feat(lid): add a finger catch pull to the sliding lid

### LINKED ISSUE #4157: Finger catch to sliding lid
**Problem**
The existing slide lid options don't quite meet my needs primarily with small boxes. The existing finger notch is too small to get a finger in and the tab overhangs the box's footprint.

**Proposed Solution**
Is it possible to add a raised edge to the lid that mirrors the opposite side?

Image

**Alternatives Considered**
A depressed dimple in the lid? I think this might require a thicker lid though.

andymai:
## Summary

A fourth **Pull** option for sliding lids, **Finger catch** (#4157). The finger notch is too small for a finger on a small box and the tab hangs past the footprint, so this one is the section of stacking rim the entry notch removed, carried on the lid.

- A bar the full width of the plate, rising from the plate's top to the lip's top plane. Shut, the rim reads continuous across the entry wall; open, the whole width is something to push on.
- Confined to the entry wall's thickness, one clearance in from the notch on every side, so it slides under the lip's inward overhang and never meets a stacked bin's feet. It prints flat with the plate, no support.
- On a flush or lipless bin the same bar stands the lip height above the rim.
- A compatibility warning when the wall is thin enough to leave the bar under two perimeters, naming the wall control as the fix.
- The lid size readout now reports a sliding plate (and its catch) instead of the cap-shell formula it used for every sliding lid.
- `lid.slide.pull` gains `catch` in the type, the server allowlist, the schema enum and all 14 locales, plus a what's-new entry.

## Verification

- Plan tests: span, rise, depth and the depth-inside-the-wall invariant, recessed and flush.
- Kernel seating test: the closed lid's top at the bar lands on the lip's top plane, and the worst shared volume over the plate's whole travel stays at the contact floor.
- Compatibility, server validation, control and readout tests.
- Local: typecheck, lint, the four i18n checks, doc-drift, the sliding-lid seating and lid generator suites, and the lid section suites.
- Headless render of the control, its hint, the half-open lid with the bar and the closed lid with the rim completed.

Closes #4157.

https://claude.ai/code/session_01UcvYmj4zKK2kDHoMHB6WF2

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/4162?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


brendo:
Is it intended for the catch to be higher than the remaining sides when stacking is disabled? 

Screenshot 2026-09-10 at 10 05 19 am

PR #4221, opened 2026-09-11
fix: finger catch lip height is now flush when stacking lip is disabled

brendo:
## Why

The finger catch introduced in #4162 has a height which exceeds the other sides. 

Screenshot 2026-09-10 at 10 05 19 am

## Changes

This PR fixes that so the height is flush with the remaining sides. Stacking lip catch height is unaffected.

Screenshot 2026-09-11 at 11 12 12 am

## Risk

I have manually printed this part and it worked for my needs:

<details>
<summary>Screenshots</summary>
PXL_20260911_022516198
PXL_20260911_022657257
</details>

---

<sub>Title format, lint, types, i18n, tests, coverage and bundle size are gated by
`pr-title.yml`, the pre-commit hooks and CI. No need to restate them here.</sub>

### DIFF
--- a/src/shared/utils/slideLidPlan.test.ts
+++ b/src/shared/utils/slideLidPlan.test.ts
@@ -494,12 +494,9 @@ describe('finger catch', () => {
     expect(g.plate.trailingX - g.plate.pullDepthMm).toBeGreaterThan(TRAVEL_INNER / 2);
   });
 
-  it('stands the lip height above the rim on a flush lid', () => {
+  it('ends flush with the wall top when the bin has no stacking lip', () => {
     const g = geometryOf({ slide: { ...catchConfig, placement: 'flush' }, hasLip: false });
-    expect(g.plate.pullReachMm).toBeCloseTo(
-      g.plateTopBelowWallTopMm + GRIDFINITY_SPEC.LIP_HEIGHT - GRIDFINITY_SPEC.LIP_OVERLAP,
-      9
-    );
+    expect(g.plate.pullReachMm).toBeCloseTo(g.plateTopBelowWallTopMm, 9);
   });
 
   it('has no depth on the other pulls', () => {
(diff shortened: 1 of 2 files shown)


andymai (approved):
Verified and approved — thanks, this is a clean diagnosis.

The reason it's right is that the catch exists to replace what the entry notch removed, and the notch already branches on the same flag:

```
notchTopAboveWallTop = input.hasLip ? LIP_HEIGHT + 1 : 1
```

So on a lipless bin the notch only ever cut to 1mm above the wall top, while `pullReach` unconditionally added `LIP_HEIGHT - LIP_OVERLAP`. The bar was restoring a lip band that was never there — 4.3mm of it. Making `pullReach` consult `hasLip` puts the two back in agreement, which is the invariant that should have been holding all along.

Checked the four combinations, and they're all consistent now — `slidePlateTopBelowWallTopMm` already folds in both `placement` and `hasLip`, and it documents that the two placements coincide on a lipless bin, so the `recessed` + lipless case is covered by equivalence with the `flush` one you tested.

Also traced the downstream consumers, since a compensating offset elsewhere would have turned this into a regression: `catchSolid` extrudes `thicknessMm / 2 + pullReachMm`, the footprint clip window uses `rise = pullReachMm`, and `useLidSection` adds it for the height readout. All three scale with the value rather than assuming the old constant, so nothing double-counts.

The lipped path is genuinely untouched: `spans the whole plate and fills the rim on a recessed, lipped bin` still asserts the full `LIP_HEIGHT - LIP_OVERLAP` rise, and the mesh-level `a finger catch fills the rim it took away` in `slideLidSeating.scenario` still measures the bar's top landing on the lip's top plane. Both pass.

Renaming the old test rather than deleting it was the right call — it was asserting the defect, and the new name says what the geometry should actually do.

Ran `slideLidPlan`, `slideLidSeating`, `slideRailBuilder`, and the slide-tray scenario and export suites locally (80 tests), plus the `LidSection` panel tests for the other consumer (116). All green, no snapshot movement.

Appreciate you printing it before sending — that's better evidence than anything CI produces here.

PR #4291, opened 2026-09-15
fix(generation): rebuild the baseplate pocket and bin foot to the Gridfinity profile

cinnayyannic:
## Summary

The baseplate pocket and the bin base were each built as the full-cell contour
displaced 0.25mm **along its own 45° taper**, rather than offset perpendicular to
it. Both are rebuilt to the published profiles: pocket `2.15 / 1.8 / 0.7` at
**4.65mm** deep on a 42mm cell with r4 corners, base `2.15 / 1.8 / 0.8` at
**4.75mm** on a 41.5mm cell with r3.75.

## Why

Displacing a profile along a 45° face and offsetting it across that face are not
the same thing. The old pair measured 0.25mm apart horizontally but only
**0.177mm** (= 0.25/√2) perpendicular, which is the number that matters for fit,
and it left every pocket face *parallel* to the face it receives — so a seated
foot had nothing to land on and sank the full clearance before contacting
anything.

The measurable consequence is interoperability. Our base's vertical band sat at
half-width 18.85 where the spec puts it at 18.60, so against a third-party
baseplate (whose vertical band is also 18.85) there was **zero** clearance over a
1.7mm-tall band. Our plate was correspondingly 0.25mm loose for a spec bin.

Separately, both draft paths drew the pocket as a *single* taper — `addPocketWalls`
emitted only a top and bottom ring, and the preview cutter was a two-section loft.
That renders as a cone: a different part, not a coarser one. The base's draft mesh
had the same defect.

With both profiles corrected, the clearance is 0.25mm perpendicular to every
mating face, and because the base's cell is already `TOLERANCE` narrower, the two
corner arcs become concentric at (17, 17) — so the corners carry the same 0.25mm
as the flats. Measured on the generated meshes, with both parts on a shared datum:

| z | pocket ½-width | base ½-width | gap |
|---|---|---|---|
| 0.00 | 18.15 | 17.80 | 0.35 (45° → **0.25 ⊥**) |
| 0.70 | 18.85 | 18.50 | 0.35 |
| 0.80 | 18.85 | 18.60 | **0.25** |
| 2.50 | 18.85 | 18.60 | **0.25** |
| 2.60 | 18.95 | 18.60 | 0.35 |
| 4.65 | 21.00 | 20.65 | 0.35 |

## Changes

- `FOOT_PROFILE` / `POCKET_PROFILE` in `generatorConstants` are now the single
  source both BREP builders and both direct-mesh builders walk, so the four
  copies of the profile can't drift apart again.
- Baseplate printed height follows the pocket: **4.65mm** (+ floor), via
  `baseplateTotalHeight`. `SOCKET_CORNER_RADIUS` → `BASEPLATE_CORNER_RADIUS`,
  since the socket doesn't use it — the base sweeps `BOX_CORNER_RADIUS`.
- Preview and export now build the same pocket. The 2-section "simplified"
  cutter is gone, and `getPocketTemplate` loses its `forExport` parameter.
- `INSET_BOT` is split into `POCKET_INSET_BOT` (2.85) and `FOOT_INSET_BOT`
  (2.95). They are 0.1mm apart and interchangeable-looking, and
  `floorPatternWindow` / `lightweightFloorCutter` use it to keep drainage holes
  inside the **foot's** flat underside — the pocket value there would put them
  0.1mm into the tapered flank a baseplate mates against.
- `STACK_JUNCTION_MM` follows the profile to `SOCKET_HEIGHT` (see Breaking).
- New `seatClearance.verification.test.ts` measures the gap *perpendicular* to
  each mating face. It fails on the old profile at 0.1768 and passes at 0.25.

## All core dimensions now manually measured against the Gridfinity spec

Baseplate profile -> matches
Bin base -> matches
Bin dimensions -> match
Stacking lip -> matches

sepc_compliance_1
sepc_compliance_4
sepc_compliance_3

## Risk

**Look hardest at `STACK_JUNCTION_MM`.** It changes what every stack-height
readout prints. I did not take it on faith: `binStackSeating.kernel.test.ts`
measures 4.75 by dropping real solids together across every height, footprint and
height unit, and that agrees with an independent derivation from the published lip
and base profiles (the base's flare and the lip's funnel are parallel 45° faces;
all three contact conditions bind at `h₀ = −0.35`, giving `4.4 − (−0.35)`). The
result is robust to the one lip number worth questioning: at `2.15` over a `4.65`
lip instead of `1.9` over `4.4`, the funnel line becomes `r = 16.1 + z` and the
junction is still 4.75.

**79 triangle-count snapshots move**, because the base and the pocket each lost a
loft section.

Two existing tests were passing vacuously and now assert something real:
`detachableFeet.kernel` probed planes from a 5mm profile this repo has never had,
so its `extremeAt` returned `-Infinity` on both sides and compared equal — it now
reads its planes off `FOOT_PROFILE`. And an outline invariance check compared
`vertices.length`, which counts vertex *welding*: the drifted outline is flush
with the plate perimeter while the re-based one is a unit inside it, so the two
weld a different number of seam vertices onto an identical top face. I confirmed
the point sets are congruent under the shift (3584 distinct vertices each) and
switched the assertion to triangle count.

`lidStackGrid` is deliberately **not** migrated. It still carries the pre-spec
insets, so it grips 0.25mm/side looser than a baseplate — loose, never tight.
Migrating it also has to move `JUNCTION_RELIEF_FLOOR_Z`, which is positioned by a
vertical rim the corrected profile doesn't have, and #4272 has since reworked that
file independently.

**Verification is geometric, not physical — I have not printed these parts.**
Full suite green apart from three script checks (`check-docker-parity`,
`check-csp-hashes`, `schema/docs`) that fail identically on an unmodified
`upstream/main` checkout here; they look like a local CRLF artefact and should
pass in CI.

## Breaking changes

- **Baseplates print 0.1mm shorter** (4.75 → 4.65mm plus floor). This is the spec
  height, and it's the change that lets a bin's base bottom out on the pocket
  floor instead of wedging on its tapers. Exported geometry differs from
  previously generated plates.
- **Stack height readouts drop 0.25mm per junction**, since a stacked bin settles
  one whole base profile below the lip top rather than one less the clearance
  step. Affects the assembled-height panel, drawer-ceiling fit and stack pitch.
- Bins generated before this change are 0.25mm/side fatter below the top face
  than bins generated after it. Both seat in this plate; only the new ones seat
  in a third-party one.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/4291?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thank you for this. The perpendicular-offset point was right, and the measurements checked out on the built meshes here: the foot's band was 18.85 mm half-width against the spec's 18.60, and the pocket's top step was also removing the top 0.25 mm of every ridge on export, so the fix does more than the description claims. The full generators suite and the main-only heavy suite both pass on the branch.

I pushed one commit to your branch for the Prettier failure in three margin scenario tests and a stale "5mm slab" note. Merging once CI is green.

PR #4294, opened 2026-09-15
fix(lid): rebuild the stack grid as a trimmed baseplate pocket

cinnayyannic:
## Summary

Finally fixed the stack grid on lid option.

Before:
grafik

After:
grafik

grafik

Rebuilds the lid's stack grid as the baseplate pocket cut at **full size** into a
slab one half-clearance shorter, letting the slab's own top face do the trimming.
Replaces the hand-written profile table, the land inset and the
"same value as `baseplateGenerator`" cross-reference.

## Why

A grid on a lid is not a Gridfinity feature, so there is no profile to quote.
What it has to do is receive a bin's foot exactly as a baseplate does, and the
only thing in the way is size: a lid is the **bin's** footprint —
`gridUnitMm - 2 * LID_FIT_CLEARANCE`, with `lidCornerR = LID_CORNER_RADIUS - fitClearance`
— which is the nominal cell offset inward by exactly `CLEARANCE / 2`, flats and
corners alike. A pocket grid on the nominal lattice therefore overhangs the lid
by precisely that, and trimming the overhang off a 45° face lowers its high
point by the same amount.

That framing exposes a bug in #4272. It pulled the rim in by `CLEARANCE / 2` but
left the taper's far end in place:

```
[SOCKET_HEIGHT - CLEARANCE/2,  0.25]   → z 4.50
[SOCKET_HEIGHT - BIG_TAPER,    1.90]   → z 2.60
```

1.65mm of inset over 1.90mm of drop is **≈41°**, not 45 — so the funnel no longer
mates face-to-face with the foot's 45° flare, it meets it on an edge. And its rim
sat at half-width 20.75 against the base profile's own widest 20.75, i.e. **no
clearance at all** at the rim. #4272's note that it "narrows the rim opening to
~the real Gridfinity socket spec, snugging the stack a touch" lands as line-to-line
against the spec base.

Trimming instead of re-deriving cannot produce that class of error: a cut does not
change a face angle. Measured on the generated mesh, both tapers are back to slope
`1.0000`, and a seated foot holds `CLEARANCE / 2` perpendicular to every mating
face, landing on the grid floor with the foot 0.35mm proud.

## Changes

- `lidStackGrid` maps `POCKET_PROFILE` onto Z and extrudes the slab to
  `LID_STACK_GRID_HEIGHT_MM`. The cutter deliberately **overhangs** the slab in
  both directions — half a clearance past the lid's edge and half above its top.
- `LID_STACK_GRID_HEIGHT_MM` is derived in `shared/printSettings` so
  `assembledHeight` can reach it without crossing the module boundary.
- `STACK_INSET_BOT` is now the baseplate's own floor inset: the trim takes its
  slice off the **top**, so the floor an upper bin lands on is untouched.
  `lidCutoutPlan`'s mirror follows.
- `seatClearance.verification` covers the lid alongside the baseplate, asserting
  on `POCKET_SECTIONS` — the table the builder actually lofts — rather than on a
  number copied twice.

## Risk

**The overhang is load-bearing, and the obvious simplification breaks it.**
Pre-trimming the profile so the rim sits at inset 0.25 lands it exactly on the lid
outline, arcs included. That is a tangency rather than a cut, and it sheds sliver
triangles that no watertight or triangle-count check reports — I hit exactly this
on the first attempt and `lidGenerator.scenario`'s `sliverCount` caught it (3).
There is now a test asserting the rim stays above the slab.

**Two visible changes.** Lids print **0.35mm shorter** (grid 4.75 → 4.4).
`assembledHeight.scenario` pins `LID_STACK_GRID_HEIGHT_MM` against a real
generated lid mesh, so the billed number and the solid cannot drift apart.
**Lip-only now tops flush with the full grid** instead of 0.25mm below it — the
lip is that same grid's perimeter, so ending where the lands do is the defensible
result; the test that previously distinguished them by height now leans on the
triangle-count assertion that was always the real discriminator.

**Verification is geometric — nothing has been printed.** Full suite green apart
from three script checks (`check-docker-parity`, `check-csp-hashes`,
`schema/docs`) that fail identically on an unmodified `main` checkout here and
look like a local CRLF artefact.

I checked the new assertions can actually fail: reverting `POCKET_SECTIONS` to
#4272's shape fails four of the five lid cases, including the tangency guard, at
`0.868421` — exactly the 41° ratio.

## Breaking changes

Lids with a stackable top are 0.35mm shorter than previously generated ones, and
stack-height readouts follow. The grid still receives a foot at full depth; only
the material above the seat changes.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/4294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Thanks for closing the loop on this one. The angle claim checked out on the built lid mesh: the old funnel ran 1.65 mm of inset over 1.9 mm of drop, the new one is 1.9 over 1.9, and the pocket meets the lid outline at the slab top. The full generators suite and the main-only heavy suite pass on the branch.

I pushed one commit to your branch for the Prettier failure in two files and a stale comment in assembledHeight. Merging once CI is green.

Issue #4041, opened 2026-08-31
issues with the new spacemouse support


tg73:
This is effectively a first actual spacemouse hardware test report. This report includes successes and failures. I'm writing from a CAD (Fusion) user's perspective in terms of expected behaviour.

For reference:
<img width="136" height="282" alt="Image" src="https://github.com/user-attachments/assets/2f3c9952-7e8a-4346-ba51-761094474a62" />

- Movement`1` adjusts the correct DoF but the action is inverted: moving the puck left should move the model left, but it moves the model right.
- Movements `2` and `3` are cross-mapped in the app. `2` should be vertical panning, `3` should be zoom in and out, but the actual behaviour is the other way around.
- Movement `4` works correctly.
- Movement `5` does nothing, and I think this is correct behaviour. Rotating the horizon is not desirable, and does not seem to be possible with regular mouse control.
- Movement `6` works correctly.

**To Reproduce**
Steps to reproduce:

1. Enable SpaceMouse support in the labs settings
2. Try manipulating any 3d viewport with the spacemouse

**Environment:**

- Browser: Chrome 152.0.7977.65 (Official Build) (64-bit)
- Device: win11 desktop, SpaceMouse Pro

tg73:
Additional issue:

The app responds to spacemouse input even when it does not have focus. Eg, if I move a model in PrusaSlicer, the app also moves at the same time.

andymai:
Fixed in #4048 (merged): corrected the inverted left/right translation and the pan/zoom (lift vs push) axis swap. The axis-to-motion assignment is fixed; the exact directions of the swapped axes were set from your report and are overridable per-axis via the SpaceMouse settings invert toggles. Closing; please reopen if, once this deploys, lifting the puck pans down or pushing forward zooms out. I couldn't confirm the signs without the hardware.

tg73:
@andymai please reopen, I don't have permission

Ok, so I was partly mistaken in my first report. Motion was weird but I misinterpreted what the weirdness was. I'm now directly comparing how Fusion and PrusaSlicer behave compared to the app. I've taken advantage of the issue that the app responds to spacemouse input even when it does not have focus to record PrusaSlicer and the app responding to the same input. The zip files contain the corresponding mp4 videos.

In PS/Fusion:
- Moving the puck back/forward (action `2`) zooms into the object (pull puck towards you) and out from the object (push puck away).
[...]

PR #4328, opened 2026-09-18
fix(spacemouse): let the driver keep the model in view

andymai:
**Draft — needs SpaceMouse hardware verification before merge.** Addresses the open item on #4041 (the view limits). I have no puck, so the reasoning is verified against the SDK and the code, not the feel.

## Why the current approach fights the driver

The 3Dconnexion SDK's own `web_threejs` sample **never clamps the view**. It hands the driver `getModelExtents` + `getViewFrustum` and lets the driver keep the model framed — that's why Onshape-in-Chrome (which the reporter confirmed works) holds the model in view.

This app instead re-applies a pan leash (`clampPan`) *after* every driver-written pose in `setViewMatrix`. Keeping the model in view is a geometric relationship (model box vs frustum, unit-independent) the driver already handles, so the leash is redundant — and because it cancels motion the driver has already committed, it produces exactly the "stuck in treacle, then whips across" the reporter described and you diagnosed in-thread as "cancelling motion."

## Changes

1. **Drop the post-hoc pan clamp on the driver path** (`setViewMatrix`), leaving framing to the driver. The WebHID path (no driver) keeps its own leash unchanged. Elevation is still clamped — an explicit canvas polar limit is a fact the driver can't know, and that clamp is a no-op on a canvas that doesn't set one, so it isn't part of the treacle.

2. **Report a real `getUnitsToMeters`.** It was hardcoded to `1`, so the driver sized its motion as if a 168 mm bin were 168 **m** — the SDK sample returns the true scale (10.0). The canvases draw in different units (mm in bin/baseplate, grid cells in the layout isometric), so instead of a per-canvas constant this sizes the model's bounding sphere to a fixed hand scale (`TARGET_MODEL_METERS`): the driver sees every model at the same physical size whatever it's modelled in.

## Tradeoffs / what to check on hardware

- This **reverses a deliberate, tested decision** (the app used to leash the driver's pan; the unit test that asserted that is rewritten to assert the pose is now applied faithfully). If your driver/setup does *not* keep the model framed on its own, the model can leave view with no app backstop — that's the key thing to confirm.
- `getUnitsToMeters` recalibrates the driver's motion scale, so **speed/feel may change** and you may want to re-tune the 3Dconnexion speed slider once.
- Suggested checks (overlay on): pan the bin hard toward every edge — does the driver hold it in view like Onshape? Then movement 2 in until it fills the screen and back out; movement 4 over the top; a hard pan held to the edge, released, continued with the mouse.

If the driver does frame it, this both fixes the view-limit complaint and removes the treacle. If it doesn't, we've learned the driver needs more (or the leash stays, looser) — either way it's a cleaner base than the fighting clamp.

Refs #4041.

<a href="https://cubic.dev/pr/andymai/gridfinity-layout-tool/pull/4328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


andymai:
Closing without merging (keeping the branch for the analysis).

The reasoning that motivated this draft was the "stall-then-lurch" — the app's post-hoc pan leash cancelling the driver's committed motion. But re-reading the last report on #4041, that's already resolved: in v4.492.2 the reporter says motion "feels natural." #4167 fixed the treacle.

So the two changes here no longer pencil out:

- **Removing the pan leash** now fixes a symptom that's gone, while giving up a backstop that provably works — a hard-panned model stays intersecting the frustum across every aspect ratio and zoom in a unit probe. With no SpaceMouse on either side, trusting the driver to frame the model instead is unverifiable, and if it doesn't, the model leaves view with nothing to catch it (the exact residual complaint).
- **Correcting `getUnitsToMeters`** (hardcoded 1) is legitimately more correct per the SDK sample, but depending on how the driver's motion model consumes it, it could shift speed by up to ~1000x. Unverifiable, and the reporter just called the current speed natural.

The remaining "view limits quite broken" item is real but the reporter deprioritized it ("less of an issue"), and I couldn't reproduce it in code — the leash holds. Best not to gamble an unverifiable change onto a working state.

The SDK findings if this is ever revisited with hardware: the sample relies entirely on the driver to keep the model framed (no app clamp), driving that off `getModelExtents` + `getViewFrustum`; and it returns a real `getUnitsToMeters` (10.0) where we return 1. Branch `spacemouse/driver-owns-keep-in-view` has a working implementation of both if a puck is available to validate.

